PR #7, opened 2018-07-05
Fix/counter irate/srikanth

srikanthops:
Reviewed all the metrics used for graphs 
Fixed all of them that are counters to use irate


srikanthops:
https://github.com/lightbend/console-home/issues/164

PR #9, opened 2018-07-13
es-backend/issues/145: remove grafana login

marcoderama:
User will no longer need to enter username/password for access to grafana console.

Fix for https://github.com/lightbend/es-backend/issues/145


Makefile
```diff
@@ -4,7 +4,7 @@ all: docs/es/all.yaml docs/index.yaml
 RELEASE := $(shell awk '$$1 == "version:"{v=$$2} $$1 == "name:"{n=$$2} END {print n "-" v}' enterprise-suite/Chart.yaml )
 
 docs/es/all.yaml: docs/$(RELEASE).tgz
-	helm install $< --name=es --namespace=lightbend --dry-run --debug | sed -e '1,/^---$$/ d' > $@
+	helm install $< --namespace=lightbend --dry-run --debug | sed -e '1,/^---$$/ d' > $@
```
marcoderama: Even though the "helm install" has dry-run specified, this line would fail if the 'es' release happened to be installed on a running minikube.  The name value shows up only in the section removed by the following sed expression, so we can use the default name and not worry about the state of the minikube.

marcoderama:
We can't merge this without changing the beta behavior (to pick up latest es-monitor-api for eg.).  I suggest we hold off until we know how we're going to manage that...

PR #10, opened 2018-07-13
release 0.0.14.  Go back to using latest es-monitor-api and es-console

marcoderama:


marcoderama:
We need to use :latest for es-monitor-api and es-console while we're still fixing so many things.  cf. Step 1 of https://github.com/lightbend/es-backend/issues/168#issuecomment-404951399

PR #11, opened 2018-07-18
Extract versions into values.yaml

jsravn:
This is the first step to improving the release process as discussed at
https://github.com/lightbend/es-backend/issues/168.

Also:
* Use `helm template` to generate `all.yml` so we don't need
minikube/tiller up.
* Extract es-grafana configmaps into imported json files, same as was
done for the other components.


Makefile
```diff
@@ -14,6 +14,9 @@ docs/$(RELEASE).tgz: enterprise-suite/Chart.yaml enterprise-suite/templates/*.ya
 	helm lint enterprise-suite
 	helm package enterprise-suite -d docs
 
+lint:
```
dbrinegar: add to .PHONY please

enterprise-suite/es-grafana/ernie.json
```diff
@@ -0,0 +1 @@
+[]
```
dbrinegar: delete?

PR #13, opened 2018-07-18
lightbend/es-backend/issues/207

srikanthops:
add akka-http, jvm exporter charts
fix workloadId is all exporters


dbrinegar:
see lightbend/es-backend#207

PR #14, opened 2018-07-19
Add /service/grafana/ reverse proxy path in nginx configs.  Use in es…

marcoderama:
…-grafana configs.

Add clusterNameIP to values.yaml.  Update default es-grafana version


marcoderama:
Really need to update the README to explain how to set this up for dev.  (e.g. Change IP if required and change various versions to latest.)

marcoderama:
For the console to work with these changes, several changes to the FE code were required.  If gets merged as is, and given @jsravn changes in 
https://github.com/lightbend/helm-charts/pull/11, then in order to run the latest beta as a dev, one has to do the following:

- `cd lightbend/helm-charts/`
- edit `values.yaml` thusly:
   - change `clusterNameIP` to match the output from `minikube ip`
   - change `esConsoleVersion` to `latest`
   - change `esMonitorVersion` to `latest`
   - change `esGrafanaVersion` to `0.0.5`
- `make clean && make`
- `make install-local`   (if all went well up to this point)

(We should change es-grafana-release build to generate a `latest` version.)  It would probably be a Good Thing to add a Makefile target to automatically edit `values.yaml` for this purpose.

jsravn:
@marcoderama I'll be modifying it soon to auto generate charts with "latest" versions. Yes we should be publishing latest everywhere I think, then we can just use latest in all the versions.

jsravn:
https://github.com/lightbend/helm-charts/pull/15 has the changes to publish `latest` charts.

enterprise-suite/es-grafana/kubernetes-exporter.json
```diff
@@ -0,0 +1 @@
+[]
```
jsravn: I think you can remove this?

enterprise-suite/templates/es-grafana.yaml
```diff
@@ -132,6 +132,10 @@ spec:
             value: "true"
           - name: GF_AUTH_ANONYMOUS_ORG_ROLE
             value: "Editor"
+          - name: GF_SERVER_ROOT_URL
+            value: "http://{{ .Values.clusterNameIP }}:30080/service/grafana/"
```
jsravn: Don't we want to remove dependency on the minikube IP?

PR #15, opened 2018-07-19
Generate "enterprise-suite-latest" charts

jsravn:
These use "latest" for the image versions. The overrides can be changed
in scripts/munge-values-to-latest.yaml.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,8 @@ Refer to the documentation for installation instructions.
 
 ## Publishing Charts
 
-Make any changes to the chart and commit it. Then run:
+Make any changes to the chart, commit it and open a PR. Once the change is merged to master,
+you can release it directly:
 
     scripts/make-release.sh <version> <chart-name>
     git push --tags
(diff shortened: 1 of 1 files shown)


jsravn:
@dbrinegar I thought this might be a simpler approach. We keep the single repo, and in it we have `enterprise-suite` and `enterprise-suite-latest`. `enterprise-suite-latest` mirrors the `enterprise-suite` chart except we replace the image versions with `latest`. Then we can do `helm install lightbend-helm-charts/enterprise-suite-latest`. That means as a user I only need to add and update a single repo.

jsravn:
Okay I've tacked on all the rest of the work in this PR. The release process should be streamlined now. See the README.md for details, but tl;dr is:

```
scripts/make-release.sh 0.0.15 enterprise-suite
# check it looks good
git push
```

And that's it. Similarly it all works now for `reactive-sandbox`, and any other charts we want to put in this repo.

dbrinegar:
cool, I like the sound of it, let me take it for a spin

PR #18, opened 2018-07-20
Remove hardcoded grafana IP

jsravn:
Fix https://github.com/lightbend/es-backend/issues/155

### DIFF
--- a/enterprise-suite/templates/es-grafana.yaml
+++ b/enterprise-suite/templates/es-grafana.yaml
@@ -96,7 +96,7 @@ spec:
     spec:
       containers:
       - image: lightbend-docker-registry.bintray.io/enterprise-suite/es-grafana:{{ .Values.esGrafanaVersion }}
-        name: grafana-sever
+        name: grafana-server
         env:
           # The following env variables set up anonymous access to grafana with editor access.
           - name: GF_AUTH_ANONYMOUS_ENABLED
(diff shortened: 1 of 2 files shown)


enterprise-suite/templates/es-grafana.yaml
```diff
@@ -96,15 +96,15 @@ spec:
...
+        name: grafana-server
         env:
           # The following env variables set up anonymous access to grafana with editor access.
           - name: GF_AUTH_ANONYMOUS_ENABLED
             value: "true"
           - name: GF_AUTH_ANONYMOUS_ORG_ROLE
             value: "Editor"
           - name: GF_SERVER_ROOT_URL
-            value: "{{ .Values.esMonitorBaseURL }}/service/grafana/"
+            value: "/service/grafana/"
```
dbrinegar: omg, I thought we tried this and it didn't work?...

jsravn: It should work, I hope. This is how the default k8s grafana addon is also setup: https://github.com/kubernetes/contrib/blob/master/ansible/roles/kubernetes-addons/templates/cluster-monitoring/influxdb-grafana-controller.yaml.j2#L65

jsravn: (I tested it and it worked, btw, if that wasn't clear :))

PR #24, opened 2018-07-21
Cut a relase 0.0.19  (Fix akka-http 5xx errors)

katsutoxin:


README.md
```diff
@@ -2,6 +2,13 @@
 
 Contains public Helm charts for various Lightbend projects. This project is hosted on [GitHub Pages](https://lightbend.github.io/helm-charts/index.yaml).
 
+
+## How to cut a new release
+```
+change version in enterprise-suite/Chart.yaml
```
marcoderama: Turns out you don't need to do this anymore.  The script at scripts/make-release.sh will take care of this automatically.  But no point doing another release for that.

Makefile
```diff
@@ -36,6 +36,8 @@ docs/$(RELEASE_LATEST).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
 
 clean:
 	rm -rf build
+	rm -f docs/es/all*
```
dbrinegar: this is a bit more than "clean", this is "remove the release" ?  I think there's some basic misunderstanding here

PR #28, opened 2018-07-22
remove imagePullPolicy

dbrinegar:
part of lightbend/es-backend#150


dbrinegar:
right, this should be a no-op, just explicitly / consistently taking the default behavior and removing the last of all the policy statements

PR #33, opened 2018-07-26
refactored graphs

alanngai:
trimmed akka graphs, there were too many of them, which slowed everything down.  Also went through and reviewed each graph for reasonable dimensions to aggregate on as well as to select quantiles where applicable


enterprise-suite/es-grafana/akka-http-exporter.json
```diff
@@ -1,16 +1,8 @@
...
   {"graphName":"Response Rate", "promQL":["sum by (request_path)(irate(akka_http_request_path_endpoint_responses{workload=\"$workloadId\"}[1m]))"]},
-  {"graphName":"Response Time", "promQL":["max by (request_path)(akka_http_request_path_endpoint_response_time_ns{workload=\"$workloadId\"})"]}
+  {"graphName":"Response Time", "promQL":["max by (request_path)(akka_http_request_path_endpoint_response_time_ns{workload=\"$workloadId\",quantile=~\"0.99\"})"]},
+  {"graphName":"Connections", "promQL":["sum by (http_server)(irate(akka_http_http_server_connections{workload=\"$workloadId\"}[1m]))"]},
+  {"graphName":"Response Time 2XX", "promQL":["max by (http_server)(irate(akka_http_http_server_response_time_2xx_ns{workload=\"$workloadId\", quantile=~\"0.99\"}[1m]))"]},
+  {"graphName":"Response Time 4XX", "promQL":["max by (http_server)(irate(akka_http_http_server_response_time_4xx_ns{workload=\"$workloadId\", quantile=~\"0.99\"}[1m]))"]},
+  {"graphName":"Response Time 5XX", "promQL":["max by (http_server)(irate(akka_http_http_server_response_time_5xx_ns{workload=\"$workloadId\", quantile=~\"0.99\"}[1m]))"]}
```
jsravn: I see these queries already existed, but I'm finding them a little weird. Because they all do `irates`. Shouldn't they be labeled like `Average Connections/s`, `Average Increase in Response Time 2xx/s`?

Also the use of irate itself will be misleading I think. It will artificially dampen the true increase. For example if there are two scrapes in the 1m period, the first one is 10s, second is 20s, that is a whopping 10s increase in response times. But irate will divide that by number of seconds in the period 1m, so divide by 60 -> it will show 0.6s increase in response times. Which is really misleading, and doesn't make much sense to me given we are already taking the 99%. It'd make a lot more sense here to use `increase` or `offset` (accurate but slower).

dbrinegar: Agree this is weird / probably wrong.

Should say `Connection Rate` in the first one, which is the one that makes sense.

But "rate of 99th" is... I guess I don't know what that is.  The last 99th, divided by 60?  Just remove the irate?

alanngai: the queries were copied over from the Cinnamon dashboard definitions.  In fact most of the queries in the Cinnamon dashboards have this form.   Perhaps someone from @henrikengstrom or @pvlugter can comment on this?

As for the percentile question, that's a more fundamental question and would need to be changed on the cinnamon side.  Open to discussing that after beta

pvlugter: The quantiles reported by the prometheus client can be configured. For the dashboards based on Coda Hale Metrics we use max by default, for the usual reasons. Don't think the prometheus summary has a built-in max, but something like `quantile = 1.0, error = 0.0` could probably be used.

pvlugter: The `irate` seems to be a copy-paste bug that made its way into some of the dashboard when templates were set up (just the http ones). Fix in https://github.com/lightbend/cinnamon/pull/1299. Prometheus dashboards haven't seen much usage yet.

pvlugter: https://www.robustperception.io/irate-graphs-are-better-graphs

PR #34, opened 2018-07-26
Update liveness/ready probes for prometheus

jsravn:
Use the correct endpoints (status can timeout, /-/ready and /-/healthy are
the correct endpoints). Also add liveness/ready for es-monitor-api to
pick up issues on the k8s layer. Bumped alertmanager version for the
/-/ready endpoint.


dbrinegar:
good change, bumping the alertmanager slowed me down tho.  I want to test that a bit

jsravn:
@dbrinegar what's the concern?

PR #35, opened 2018-07-26
remove the demo rules

dbrinegar:


alanngai:
will this affect the default akka monitors that we have (inbox growth and processing time)?  I don't want to be just left with k8s monitors as default

dbrinegar:
no there are left-over "slow_akka" rules from 4 months ago.  they aren't specified quite correctly, and will fubar things with the new es-monitor-api

PR #38, opened 2018-07-29
make it a good hack

srikanthops:
Wait till grafana comes up and then enable cinnamon plugin.. 
This is only temporary fix.. I will investigate make this enabled by default and burn it into es-grafana image


dbrinegar:
This is ok for short term, but let's make an issue to figure out how to do this properly

PR #39, opened 2018-07-29
enhance clean and add install-local-latest command

srikanthops:
I feel we need this changes in the Makefile..


Makefile
```diff
@@ -35,6 +35,7 @@ docs/$(RELEASE_LATEST).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
 	helm package build/$(CHART_LATEST) -d docs
 
 clean:
+	rm -rf docs/$(RELEASE).tgz docs/$(RELEASE_LATEST).tgz docs/es/all.yaml docs/es/all-latest.yaml
```
dbrinegar: this doesn't seem quite right to me.  There are definitely times you want to do this step--like when you have just built something but want to force rebuild it--but you also don't want to do "make clean" and remove the last release.

I think we should consider the use case for when you want to get a new build, and why that isn't working when source are updated.  There's probably some dependency missing in the Makefile specification for how those artifacts should be updated.

Makefile
```diff
@@ -63,5 +64,8 @@ delete-es:
 install-local: install-helm delete-es
 	helm install docs/$(RELEASE).tgz --name=es --namespace=lightbend --debug
 
+install-local-latest: install-helm delete-es
+	helm install docs/$(RELEASE_LATEST).tgz --name=es --namespace=lightbend --debug
+
```
dbrinegar: 👍

Makefile
srikanthops: added clean-all command

Makefile
```diff
@@ -37,6 +37,9 @@ docs/$(RELEASE_LATEST).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
 clean:
 	rm -rf build
 
+clean-all: clean
+	rm -rf docs/$(RELEASE).tgz docs/$(RELEASE_LATEST).tgz docs/es/all.yaml docs/es/all-latest.yaml
```
dbrinegar: this is not what we want to do for the same reason as before

dbrinegar:
you've included a bunch of removals from docs/ which is a great demonstration of the bug that you're committing in the Makefile as well.

We should never remove things that have been published with "make clean" things.

dbrinegar:
@srikanthops please ping me when you're ready to talk about this one

srikanthops:
David I am closing this PR looks like you added install-local in the another PR

PR #41, opened 2018-07-30
remove es-demo

dbrinegar:


srikanthops:
why not disable it.. instead of removing..

dbrinegar:
ya good question how to bundle the demo, but this demo is not the demo we want

PR #42, opened 2018-07-30
Kube pod fix

dbrinegar:
fixes lightbend/es-backend#238

### DIFF
--- a/enterprise-suite/es-monitor-api/static-rules.yml
+++ b/enterprise-suite/es-monitor-api/static-rules.yml
@@ -26,7 +26,7 @@
   expr: kube_pod_status_phase{phase="Failed"} * on (pod) group_left(workload) kube_pod_info
 
 - record: kube_pod_not_running
-  expr: kube_pod_status_phase{phase=~"Pending|Unknown"} * on (pod) group_left(workload) kube_pod_info
+  expr: max without (phase) (kube_pod_status_phase{phase=~"Pending|Unknown"} * on (pod) group_left(workload) kube_pod_info)
 
 - record: kube_workload_generation_lag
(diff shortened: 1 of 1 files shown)


enterprise-suite/es-monitor-api/static-rules.yml
```diff
@@ -26,7 +26,7 @@
   expr: kube_pod_status_phase{phase="Failed"} * on (pod) group_left(workload) kube_pod_info
 
 - record: kube_pod_not_running
-  expr: kube_pod_status_phase{phase=~"Pending|Unknown"} * on (pod) group_left(workload) kube_pod_info
+  expr: max without (phase) (kube_pod_status_phase{phase=~"Pending|Unknown"} * on (pod) group_left(workload) kube_pod_info)
```
jsravn: Should we consider `Failed` phase as well?

dbrinegar: I _think_ that Failed is when Kubernetes is not trying to run the pod anymore, and will trigger other alerts.  This one is perhaps not named clearly, but is trying to identify things that Kubernetes is trying to run but aren't running yet.  So Kubernetes is spinning away retrying or waiting for something, the alert lets you put an SLA on that so you can be alerted if things don't get going in X minutes.

dbrinegar: But yeah there's a lot of field testing and validation needed on these, much of this is conceptual or borrowed, not based on in-house rules we know work well.

dbrinegar: "kube_pod_hung_in_limbo" ?

PR #45, opened 2018-07-31
Relabel rules

dbrinegar:
fixes lightbend/es-backend#239

(this has to be merged in conjunction with the es-monitor-api version that supports this change)


enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -1,12 +1,10 @@
-{{- $promDomainUnderscored := .Values.prometheusDomain | replace "." "_" }}
-
```
jsravn: I think we need a comment here explaining that this file will be parsed by es-monitor-api. I wonder also if we need to provide a reference to the list of functions so readers understand where e.g. `dot2underscore` comes from. Maybe if we keep those functions in their own file we can just link to it here in a comment?

enterprise-suite/templates/prometheus.yaml
```diff
@@ -61,18 +61,36 @@ spec:
...
+            - /bin/sh
+            - -c
+          args:
+            - mkdir -p /etc/config/rules;
+              cp /etc/bare/prometheus.yml /etc/config/
+          volumeMounts:
+            - mountPath: /etc/config
+              name: config-volume
+            - mountPath: /etc/bare
+              name: bare-prometheus
```
jsravn: I don't see the volume `bare-prometheus` anywhere...

enterprise-suite/templates/prometheus.yaml
```diff
@@ -93,20 +111,6 @@ spec:
...
-            - -volume-dir=/etc/config
-            - -volume-dir=/etc/rules
-            - -webhook-url=http://localhost:9090/-/reload
-          volumeMounts:
-            - name: config-volume
-              mountPath: /etc/config
-              readOnly: true
-            - name: rules-volume
-              mountPath: /etc/rules
-              readOnly: true
```
jsravn: We still need this for the prometheus template, right? Otherwise if someone does a helm update where only the template changes, they won't get the changes. Can the configmap-reload kill/restart the pod? That would be simple enough for us I think.

enterprise-suite/templates/prometheus.yaml
```diff
@@ -198,9 +197,21 @@ metadata:
...
+      scrape_interval:     1m
+      evaluation_interval: 1m
+    scrape_configs:
+      - job_name: wait
+        static_configs:
+          - targets: ['localhost:9090']
+        metric_relabel_configs:
+        - source_labels: [__name__]
+          regex: .+
+          action: drop
```
jsravn: Any reason to not put this in `enterprise-suite/es-monitor-api`? Then it can just be included like the other files in there. Sort of nicer I think than inline yaml.

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -251,3 +250,6 @@ scrape_configs:
         action: replace
         regex: (.*)-[0-9]+
         target_label: workload
+
+    metric_relabel_configs:
```
jsravn: Is this right? According to the docs metric_relabel_configs doesn't apply to autogenerated series like `up` - I thought we use those metrics, same with scrape time, etc.

enterprise-suite/templates/prometheus.yaml
dbrinegar: This one is the prometheus reload, which is no longer needed.

There's another task for es-monitor-api to reload configs, but I think we could do that built in without needing a side car like this?  Anyway, it isn't a common path, don't think we need reload for things to work properly in the beta.

enterprise-suite/templates/prometheus.yaml
dbrinegar: Yeah this is for the init container to set things in place before es-monitor-api and prometheus are started.  The idea is for prometheus to be able to start with a bare config, and essentially wait for the real config instead of exiting.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: Agree on the note, trying to sort out how to add a note that doesn't also end up in the final prometheus config file haha

PR #46, opened 2018-08-02
latest

dbrinegar:


enterprise-suite/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 description: Lightbend Enterprise Suite
 name: enterprise-suite
 icon: https://lightbend.github.io/helm-charts/lightbend.svg
-version: 0.8.1
+version: 0.8.2
```
marcoderama: This means a new stable will have version 0.8.3 (with no intervening 0.8.2).  I guess that's okay and I don't see a way around it.  If you reverted this, and then pushed out yet another latest-only, it would overwrite the the 0.8.2 that this one will create...

PR #50, opened 2018-08-08
Namespace aggregation

dbrinegar:
Rollup prometheus config fixes for various bugs related to namespace relabeling.  This makes namespace labeling consistent, so we have a tuple of namespace, workload, monitor_type hanging off of incoming data in the right way.

fixes lightbend/es-backend#268
fixes lightbend/es-backend#267
fixes lightbend/es-backend#253

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,10 @@ docs/$(RELEASE_LATEST).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
 	scripts/munge-to-latest.sh build/$(CHART_LATEST)
 	helm package build/$(CHART_LATEST) -d docs
 
+# duplicate method here to generate the index, to avoid pulling in RELEASE as a dependency
+latest: init lint docs/es/all-latest.yaml docs/$(RELEASE_LATEST).tgz
+	helm repo index docs --url https://lightbend.github.io/helm-charts
+
 clean:
 	rm -rf build
 
@@ -63,5 +67,8 @@ delete-es:
 install-local: install-helm delete-es
 	helm install docs/$(RELEASE).tgz --name=es --namespace=lightbend --debug
 
+install-local-latest: docs/$(RELEASE_LATEST).tgz install-helm delete-es
+	helm install docs/$(RELEASE_LATEST).tgz --name=es --namespace=lightbend --debug
+
 # always run these steps if in dependencies:
-.PHONY: all build install-local install-helm delete-es lint init clean lint-json lint-promql
+.PHONY: all build install-local install-local-latest install-helm delete-es lint init clean lint-json lint-promql latest
--- a/Makefile
+++ b/Makefile
@@ -23,11 +23,11 @@ docs/index.yaml: docs/$(RELEASE).tgz docs/$(RELEASE_LATEST).tgz
 	$(call banner)
 	helm repo index docs --url https://lightbend.github.io/helm-charts
 
-docs/$(RELEASE).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
+docs/$(RELEASE).tgz: $(CHART)/* $(CHART)/*/*
 	$(call banner)
 	helm package $(CHART) -d docs
 
-docs/$(RELEASE_LATEST).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
+docs/$(RELEASE_LATEST).tgz: $(CHART)/* $(CHART)/*/*
 	$(call banner)
 	rm -rf build/$(CHART_LATEST)
 	cp -r $(CHART) build/$(CHART_LATEST)
--- a/enterprise-suite/es-monitor-api/prometheus.yml
+++ b/enterprise-suite/es-monitor-api/prometheus.yml
@@ -21,26 +21,6 @@ alerting:
       action: keep
 
 scrape_configs:
-  - job_name: prometheus
-    static_configs:
-      - targets:
-        - localhost:9090
-    metric_relabel_configs:
-      - target_label: workload
-        replacement: prometheus-server
-      - source_labels: [__meta_kubernetes_namespace]
-        action: replace
-        target_label: namespace
-      - source_labels: [__meta_kubernetes_service_name]
-        action: replace
-        target_label: kubernetes_name
-      - source_labels: [__meta_kubernetes_pod_node_name]
-        action: replace
-        target_label: node_name
-      - source_labels: [__meta_kubernetes_pod_host_ip]
-        action: replace
-        target_label: node_ip
-{{ .MonitorTypeRules | indent 6 }}
 
   # This uses separate scrape configs for cluster components (i.e. API server, node)
   # and services to allow each to use different authentication configs.
--- a/enterprise-suite/es-monitor-api/prometheus.yml
+++ b/enterprise-suite/es-monitor-api/prometheus.yml
@@ -54,6 +54,8 @@ scrape_configs:
     kubernetes_sd_configs:
       - role: node
 
+    honor_labels: true
+
     relabel_configs:
       - action: labelmap
         regex: __meta_kubernetes_node_label_(.+)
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -23,17 +23,21 @@ docs/index.yaml: docs/$(RELEASE).tgz docs/$(RELEASE_LATEST).tgz
 	$(call banner)
 	helm repo index docs --url https://lightbend.github.io/helm-charts
 
-docs/$(RELEASE).tgz: $(CHART)/Chart.yaml $(CHART)/templates/*.yaml
+docs/$(RELEASE).tgz: $(CHART)/* $(CHART)/*/*
```
marcoderama: Why these changes?

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -134,7 +130,8 @@ scrape_configs:
       - action: labelmap
         regex: __meta_kubernetes_service_label_(.+)
       - source_labels: [__meta_kubernetes_namespace]
-        action: replace
+        target_label: namespace
+      - source_labels: [__meta_kubernetes_namespace]
```
marcoderama: It's okay that particular source_labels line appears twice?

Makefile
dbrinegar: these map all the dependencies properly, so when I twiddle the `es-monitor-api/prometheus.yml` template everything rebuilds properly.

My test loop was basically:

1. change the configs
2. `make install-local-latest` which then makes latest and installs it in minikube
3. run smoketest
4. poke around at the UI

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: yep, I want to normalize everything on `namespace` but we have one hanger on in the UI that needs the `kubernetes_namespace` tag for now at least.  Dupe tags are ok since they don't change the number of metrics.

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -114,6 +108,8 @@ scrape_configs:
     kubernetes_sd_configs:
       - role: endpoints
 
+    honor_labels: true
```
jsravn: honor_labels makes https://github.com/lightbend/es-backend/issues/257 more important - to avoid conflicts.

jsravn:
I'm curious why we have all of the various SD configs for services/endpoints/etc. When, as far as I know, we don't test, document, or support that config. Doesn't it make sense to delete everything but pod discovery until we plan to support the other configs?

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -148,23 +145,29 @@ scrape_configs:
...
-      - source_labels: [created_by_kind, pod]
-        regex: '.none.;(.*)'
+      - source_labels: [created_by_kind, pod, workload]
+        regex: '.none.;(.+);'
+        target_label: workload
+      - source_labels: [deployment, workload]
+        regex: '(.+);'
+        target_label: workload
+      - source_labels: [daemonset, workload]
+        regex: '(.+);'
```
jsravn: Don't we need these rules for the pods? How come they are only added to `Service` service discovery? Which I don't think we're using anywhere, are we?

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -207,7 +210,7 @@ scrape_configs:
       - action: labelmap
         regex: __meta_kubernetes_service_label_(.+)
       - source_labels: [__meta_kubernetes_namespace]
-        target_label: kubernetes_namespace
+        target_label: namespace
```
jsravn: Worth keeping both so we don't break the frontend?

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -251,13 +255,13 @@ scrape_configs:
       - source_labels: [__meta_kubernetes_pod_host_ip]
         action: replace
         target_label: node_ip
-      - source_labels: [__meta_kubernetes_pod_label_pod_template_hash, __meta_kubernetes_pod_name]
+      - source_labels: [__meta_kubernetes_pod_label_pod_template_hash, __meta_kubernetes_pod_name, workload]
```
jsravn: Some explanation around this would be useful. It's not clear to me why we need to match on workload source label - since it won't be added yet.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: this section is for kube-state metrics

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: the one they're using is retained, hoping we can pick a label and use it consistently tho

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: This is a guard to require "workload" not be set, as a consistent theme on these kinds of rules we want to preserve the original if there is one already.  It could come from the exporter (like Cinnamon is configured to add a workload label) or from an annotation (annotation: {workload: blahblah}) before it hits a relabeling rule.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: Conflicts on these labels-- workload and monitor_type -- are expected and when we see them we always want to keep the original.

Am seeing some of this as well in the `namespace` label which is why I think some things were labeling `kubernetes_namespace` as a way of sidestepping that conflict.  I'd rather just use one label for namespace and honor_labels.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: hmm replied on the other view dunno why it isn't flowing here as a response as well.  github bug?!

anyway, these `created_by_kind` etc are from kube state metrics, which we scrape this way.  We could scope the job to only find kube state metrics, just haven't thought about it.

This change is about getting the relabeling we are doing to be consistent, maybe we can add other issues if we want to refactor the config more?

enterprise-suite/es-monitor-api/prometheus.yml
jsravn: Not sure what you mean that they are expected. What if a customer already uses `monitor_type` for their own purposes? We can't ask them to change their app right? That's why I thought https://github.com/lightbend/es-backend/issues/257 is necessary.

enterprise-suite/es-monitor-api/prometheus.yml
jsravn: So users can override the value?

jsravn:
Few more questions. The only change I'd like to see is making the kube-state-metrics relabeling more explicit, so it's obvious its for kube-state-metrics pods only rather than any generic service. We could merge it as-is and do that separately though.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: Well that would definitely be a bummer, and indeed the point of the other issue.  No this is the design of intentionally setting monitor_type upstream, to adopt a set of monitors for a workload.  For example, you can cook up a set of monitors with
```
monitor_type="my_special_akka_monitors"
```
and then use those to monitor your deployment by using that as an annotation.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: yes!  this and monitor_type are intended to work if set by user.  to be determined if/when we leverage that, of course.

jsravn: I really wish we had these reqs captured somewhere. It will make the support user story really difficult if these things are not captured anywhere, I think.

enterprise-suite/es-monitor-api/prometheus.yml
jsravn: https://github.com/lightbend/es-backend/issues/279 to document this

PR #53, opened 2018-08-13
Add travis build and smoke tests

jsravn:
Also move ES specific things into `enterprise-suite/Makefile` to allow for sharing the repo with others.

At the top level:
* I renamed `lint` to `test`, so it will apply more generically to non-linting tests.
* Added `minikube-test` target, which is for the slower smoke tests. This runs separately to `test`.

For https://github.com/lightbend/es-backend/issues/241


srikanthops:
Please update Readme, especially on how to use smokeTests

jsravn:
@srikanthops Okay did that.

PR #57, opened 2018-08-16
enable ability to scrape external service endpoints

alanngai:


enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -145,6 +145,14 @@ scrape_configs:
...
     # workload labels:
     metric_relabel_configs:
+      - source_labels: [__meta_kubernetes_pod_label_pod_template_hash, __meta_kubernetes_pod_name]
+        action: replace
+        regex: '[^;]+;(.*)-[^-]+-[^-]+'
+        target_label: workload
+      - source_labels: [__meta_kubernetes_pod_label_statefulset_kubernetes_io_pod_name]
+        action: replace
+        regex: (.*)-[0-9]+
+        target_label: workload
```
jsravn: Does this work with kube-state-metrics? This entire section actually is meant for kube-state-metrics scraping in the local helm install. We discussed this before and decided we need to make a separate job_name for service scraping, and make this tied to kube-state-metrics. https://github.com/lightbend/es-backend/issues/304 is the issue.

alanngai: this is frankly voodoo to me.  @dbrinegar helped me make these changes so that I could enable metric scraping from localhost.  I didn't need kube-state-metrics for that.  If there is a better way to do it, I'm all for that, but I don't know enough about prom config to do that.

dbrinegar:
This PR is incomplete, few other things to shuffle around per issues in backlog.  If I get through the sprint planned stuff I can take this chunk next.  We want this cleanly split out so we aren't accidentally mucking up one data set with the other.

dbrinegar:
@alanngai asking for a release to show off dev mode.  Note that we could add a kube_pod_info row for dev mode and it would show up in the UI properly.  We don't have any issues tracking dev mode.

PR #59, opened 2018-08-17
Add service resource to support ingress

srikanthops:
to set up ingress we need service in the backend.. so keep the expose-es-console service in non minikube deployments

connect to https://github.com/lightbend/es-backend/issues/270


jsravn:
@srikanthops we shouldn't use this service because it's a nodeport. We should have an additional service that we always load, which is not a nodeport. Then it can be used for ingress and other internal access.

srikanthops:
@jsravn I think you are recommending clusterIP style service.. ingress will fail for this type of service... While I cannot guarantee ingress will always work with nodePort style service.. It worked in GKE

https://github.com/kubernetes/kubernetes/issues/26508

jsravn:
@srikanthops we still shouldn't use the expose-es-console service, because the nodeport is fixed. For GCE ingress, we would just use NodePort type (and let the port be assigned randomly). For other ingress controllers, like the nginx ingress controller, NodePort isn't necessary.

I think in this case we can create `es-console` service with `ClusterIP` as default, and we can have an option to set its service type?

enterprise-suite/values.yaml
```diff
@@ -23,3 +23,5 @@ imagePullPolicy: IfNotPresent
 prometheusDomain: prometheus.io
 # minikube debug resources
 minikube: true
+# expose-es-console
+exposeConsoleService: false
```
dbrinegar: gotta stop using vim man, or set vim to handle line endings properly.  see the little red note here, we see those in your files all the time

image

enterprise-suite/templates/es-console.yaml
```diff
@@ -39,8 +40,29 @@ spec:
   selector:
     run: es-console
   type: NodePort
-{{ end }}
 
+{{ else }}
```
dbrinegar: why is this conditional?  what does the complexity cost of another option buy us in this case?  Is there some concern that there is a typical case where we wouldn't want to define this service, or we'd be unpleasantly surprised if this were defined?

I'd just define the service, for all cases, no condition.

enterprise-suite/templates/es-console.yaml
```diff
@@ -24,7 +24,8 @@ spec:
           configMap:
             name: es-console
 
-{{ if .Values.minikube }}
+{{ if or .Values.minikube .Values.exposeConsoleService }}
```
dbrinegar: I don't think you're reading the feedback on this one properly.  This fixed node port expose service must only be defined on minikube.  It must not be defined for anything else.

enterprise-suite/values.yaml
srikanthops: this is n't vim fault, apparently there is a setting in intellij I didn't set 
No more from now on
Mac : Preferences > Editor > General > Ensure line feed at file end on save

srikanthops:
Closing use this new PR
https://github.com/lightbend/helm-charts/pull/72

PR #61, opened 2018-08-21
Move the kubernetes-exporter graphs to cluster-exporter

jsravn:
As these are cluster level metrics. Kubernetes-exporter should have
container metrics, to be done in
https://github.com/lightbend/es-backend/issues/310.

This will require a change in the console to set `service-type=cluster`
for the top level grafana link.

For https://github.com/lightbend/es-backend/issues/289
For https://github.com/lightbend/es-backend/issues/306

### DIFF
--- a/enterprise-suite/es-grafana/cluster-exporter.json
+++ b/enterprise-suite/es-grafana/cluster-exporter.json
@@ -1,0 +1,7 @@
+[
+   {"graphName": "CPU Percent", "promQL":["node_cpu_percent"]},
+   {"graphName": "Memory Available Percent", "promQL":["node_memory_usable_percent"]},
+   {"graphName": "File System Free Percent", "promQL":["node_filesystem_free_percent"]},
+   {"graphName": "Bytes Received Per Second", "promQL":["sum by (instance) (irate(node_network_receive_bytes[1m]))"]},
+   {"graphName": "Bytes Transmitted Per Second", "promQL":["sum by (instance) (irate(node_network_transmit_bytes[1m]))"]}
+]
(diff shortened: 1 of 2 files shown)


katsutoxin:
Add UI designer @craigupson  in the loop.

Hi @jsravn ,  I have some question for this PR.

I noticed that we  rename kubernetes-exporter.json to cluster-exporter.json  in this PR , It will break how grafana show "kubernetes" service type graph.  I am curious how to handle it.  Let me explain the detail here.

In the current release, if we go to es-console workload page http://192.168.99.100:30080/workloads/es-console,  we see "Service Types: kubernetes"   in the left detail panel.
screen shot 2018-08-21 at 10 17 06 am

When we click "grafana icon" on the top,  it shows "Kubernetes Metrics" in grafana, and load kubernetes metrics from "kubernetes-exporter.json"  like the following screenshot
screen shot 2018-08-21 at 10 13 02 am

However, If we rename "kubernetes-exporter.json" to "cluster-exporter.json", It will make "Kubernetes Metrics" shows in grafana page with zero graph, I guess that we don't want that. How do we want to show the metrics for "kubernetes" service type for the above example?  Do we want to rename service type "kubernetes" to "cluster" as well?  Service type "cluster" sounds a little bit odd to me.

jsravn:
@katsutoxin that's intentional, idea is to replace the graphs with container specific ones like CPU usage, memory usage, that kind of thing. Service type stays as 'Kubernetes', but we'll put more meaningful graphs as part of https://github.com/lightbend/es-backend/issues/310.

If you'd like we can keep the graphs in both service types for now.

katsutoxin:
@jsravn I prefer to keep the kubernetes exporter as-is until we replace them with container specific ones. Thanks !

jsravn:
@katsutoxin Okay made that change.

PR #63, opened 2018-08-24
scrape refactoring

dbrinegar:
This change is some fairly straightforward refactoring--same ins and outs, no data schema change--on the prometheus scrape configs.  It is probably easiest to review the new prom template standalone rather than the diffs.  We now have a separate config for node-exporter, another for kube-state-metrics, and a third for Kubernetes Services with scrape annotations.  This includes whitespace changes in the other sections, and a smoke test for scraping pods under a service with namespace, workload, and right cardinality.

* separate kube-state-metrics, node-exporter
* scrape services with same relabeling as pods

fixes lightbend/es-backend#304
fixes lightbend/es-backend#280


enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -92,41 +72,65 @@ scrape_configs:
...
+  - job_name: 'pushgateway-exporter'
+    honor_labels: true
+
+    kubernetes_sd_configs:
+      - role: service
+
+    relabel_configs:
+      - source_labels: [__meta_kubernetes_service_annotation_{{ dot2underscore .PrometheusDomain }}_probe]
+        action: keep
+        regex: pushgateway
```
jsravn: Don't we need the `.MonitorTypeRules` here as well? To add workload, etc. to pods that are using pushgateway.

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -92,41 +72,65 @@ scrape_configs:
...
-  # * `{{ .PrometheusDomain }}/scrape`: Only scrape services that have a value of `true`
-  # * `{{ .PrometheusDomain }}/scheme`: If the metrics endpoint is secured then you will need
-  # to set this to `https` & most likely set the `tls_config` of the scrape config.
-  # * `{{ .PrometheusDomain }}/path`: If the metrics path is not `/metrics` override this.
-  # * `{{ .PrometheusDomain }}/port`: If the metrics are exposed on a different port to the
-  # service then set this appropriately.
-  - job_name: 'kubernetes-service-endpoints'
+  # * `{{ .PrometheusDomain }}/probe`: Only probe services that have a value of `true`
+
+  - job_name: 'prometheus-blackbox'
```
jsravn: Do we have a test for this probe stuff?

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -171,48 +175,109 @@ scrape_configs:
...
+        action: replace
+        target_label: kubernetes_pod_name
+      - source_labels: [__meta_kubernetes_pod_label_pod_template_hash, __meta_kubernetes_pod_name, workload]
+        action: replace
+        regex: '[^;]+;(.*)-[^-]+-[^-]+;'
+        target_label: workload
+      - source_labels: [__meta_kubernetes_pod_label_statefulset_kubernetes_io_pod_name, workload]
+        action: replace
+        regex: '(.*)-[0-9]+;'
+        target_label: workload
```
jsravn: There is a lot of duplication with the pod SD section. Would it make sense to stick this in a common file and include it in both places?

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: there is no change to pushgateway or blackbox exporter.  The metadata would describe the exporter, not the workload pushing data to the exporter.

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: no, this is just default prometheus config.  we don't use it

enterprise-suite/es-monitor-api/prometheus.yml
dbrinegar: prometheus config doesn't have that feature, this is just a file and is not evaluated as a helm chart, so.... don't think there's a easy shot for that one

dbrinegar:
per slack discussion, deleted the unsupported blackbox exporter and pushgateway

dbrinegar:
ya cleaned up a lot of copy/paste comments from prometheus defaults.  This isn't the file one should use to learn about prometheus, we want this focused on the things we support

PR #64, opened 2018-08-24
Changes to helm-charts Makefiles and scripts to support other projects

marcoderama:
This is as WIP.  Changes are for https://github.com/lightbend/es-backend/issues/315

From the helm-charts directory, any of make {all test clean lint-helm build} will run the command over all the projects (as defined in the top of the Makefile)

From a project directory (e.g. enterprise-suite), the same targets operate on that project only. They also have other targets available. (e.g. install-helm) Each project can also define their own additional targets if they want. (e.g. lint-promql) See the sample-project for how it looks in the "simple" case.

If a project defines BUILD_LATEST, then the build or install-local-latest targets build 'latest' things. Undefined by default.

The Travis build works.  I've tested most but not all of the targets. Like I said, mostly working but needs polishing.  In particular, some of the scripts still need tweaking.


enterprise-suite/scripts/run-minikube-tests.sh
```diff
@@ -8,18 +8,20 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 
 # setup
 echo "Installing ES from helm charts, [USE_LATEST_IMAGES=$USE_LATEST_IMAGES]"
-cd $script_dir/../../
-
-make build CHART=enterprise-suite
+cd $script_dir/../
+echo "pwd:$( pwd )"
```
marcoderama: Debugging.  To be removed.

enterprise-suite/scripts/run-minikube-tests.sh
```diff
@@ -8,18 +8,20 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
...
 fi
-make $target CHART=enterprise-suite
+make $target
+#echo make $target CHART=enterprise-suite
 
 # run tests
 echo "Running tests"
 cd $script_dir/../tests
-./smoketest
+echo $( pwd )/./smoketest
```
marcoderama: Debugging.  To be put back.

Makefile
```diff
@@ -1,75 +1,14 @@
-all: init test build
-build: init docs/index.yaml docs/es/all.yaml docs/es/all-latest.yaml
+# Collection of charts to process.
+# These are subdirectories of helm-charts.  Add to the list as requried.
+#
+CHARTS := enterprise-suite sample-project
```
marcoderama: Will remove `sample-project` before merge.

.makefile
```diff
@@ -0,0 +1,127 @@
...
+	helm package -d $(HELM_CHARTS_DIR)/docs $(HELM_CHARTS_DIR)/build/$(CHART_LATEST)
+
+install-local-latest: $(HELM_CHARTS_DIR)/docs/$(RELEASE_LATEST).tgz install-helm delete-local
+	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE_LATEST).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait
+
+.PHONY: install-local-latest
+
+endif
+##
+## end conditional things for 'latest' builds
```
jsravn: I think we should just put this in enterprise-suite/Makefile. Having the ifdef stuff complicates it as well, and I'd rather avoid having conditionals in makefiles if we can help it. (It would be better I think to have a separate include to turn on this ability rather than rely on the ifdef stuff).

.makefile
```diff
@@ -0,0 +1,127 @@
...
+
+# Not sure what this is/was about...
+# # duplicate method here to generate the index, to avoid pulling in RELEASE as a dependency
+# latest: init test docs/$(RELEASE_NAME)/all-latest.yaml docs/$(RELEASE_LATEST).tgz
+# 	helm repo index docs --url https://lightbend.github.io/helm-charts
+
+# What is this?
+release:
+	$(call banner)
+	echo "stub: release"
```
jsravn: I think I added this as part of the travis automation because I wanted it to do a release on merge to master. But I think I ripped that out.

sample-project/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+name: sample-project
+description: Sample Lightbend project for helming
+icon: https://lightbend.github.io/helm-charts/lightbend.svg
+version: 0.0.1
```
jsravn: Don't forget about reactive-sandbox.

enterprise-suite/Makefile
```diff
@@ -1,27 +1,23 @@
+CHART = enterprise-suite
+RELEASE_NAME = es
+
+BUILD_LATEST = true
+
+include ../.makefile
+
+# Overriding test target to include other linters
 test: lint
```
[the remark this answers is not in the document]
marcoderama: I was going to work on getting rid of those warnings somehow...  The :: syntax just says "This is an additional rule for this target", not "override this target".  (Although that _might_ work...)  I'll figure something out for this.

.makefile
```diff
@@ -0,0 +1,127 @@
+# This file should be included from chart-specific Makefiles
```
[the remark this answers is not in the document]
marcoderama: No reason.  Just a habit from a past life.  Don't really want people mucking with this one much but I was thinking of renaming it anyway.  Will do.

sample-project/Chart.yaml
marcoderama: Done

jsravn:
@marcoderama do you mind not amending the prior commit during review? Just because it makes it hard for me to see what changed (can't view the commits that address review comments).

Makefile
```diff
@@ -1,75 +1,27 @@
...
+CHARTS = enterprise-suite enterprise-suite-latest reactive-sandbox #sample-project
 
-define banner
-	$(info === $@)
-endef
 
-docs/es/all.yaml: docs/$(RELEASE).tgz
-	$(call banner)
-	helm --namespace=lightbend template $< > $@
+TOPTARGETS := all build clean test lint-helm minikube-test
```
jsravn: I'm not a big fan of sticking everything in `common.mk`, I think it makes it difficult to follow the logic (at least it did for me). At least I'd like if `all` and `build` remain in the top level Makefile since they don't need to be redefined in every sub-Makefile, as far as I can tell. I think `common.mk` should be used only when needed to reduce duplication across the sub-Makefiles.

I also think `lint-helm` is something test specific so shouldn't bleed into the top level here at all (it should be a dependency of `test` in sub-Makefiles).

Also I'm not sure about the `TOPTARGETS` and `MAKECMDGOALS` usage - I get the purpose but for someone who's not a make guru I think it's rather confusing. I'm really allergic to make voodoo, so maybe I'm biased, but I believe simpler the better when it comes to make. Maybe that's a bad reason, but I think if we only have a handful of targets it's better to explicitly spell them out. I think this needs to be understandable/maintainable even by people who are not super familiar with make.

common.mk
```diff
@@ -0,0 +1,97 @@
...
+define banner
+	$(info === $@)
+endef
+
+# Note: We may want to require explicit lists.  As implemented, we may end
+# up with random cruft in the packages...
+COMPONENTS := $(wildcard */.)
+SUBCOMPONENTS := $(wildcard */*/.)
+
+HELM_CHARTS_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
```
jsravn: At glance this looks bad and order dependent... I can't confirm this works reliably in all cases by simply looking at it. Is it really necessary? If we assume this file is only ever included in subdirectory Makefiles, we know the root directory is always going to be `..`.

enterprise-suite-latest/Makefile
```diff
@@ -0,0 +1,40 @@
...
+#
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+BASE_CHART = enterprise-suite
+HELM_CHARTS_DIR := $(abspath $(dir $(lastword $(MAKEFILE_LIST))))
+SCRIPTS_DIR := $(HELM_CHARTS_DIR)/scripts
+
+# Defines before include to tweak behavior as required
+VERSION = $(shell $(SCRIPTS_DIR)/export-chart-version.sh $(BASE_CHART))
+ALL_YAML = all-latest.yaml
```
jsravn: I'm not sure about this - it's trying to pass arguments in via global variables defined in another file. Feels pretty hacky and easy to break or misunderstand. Can we do this another way - `common.mk` should be as simple as possible without much conditional logic. Even if it means copy pasting some of common.mk in here, I think that's better.

Makefile
jsravn: Okay I understand the point now - it's because we're trying to template for every single chart in the CHARTS array. Are we making this too complicated? Maybe we should just build a single CHART in the top level Makefile, the way it was before?

Then you can move many of the targets out of `common.mk` into the top level `Makefile` and it removes a lot of the complexity. This fits with travis as well, since there we want to build the charts in parallel so we don't really have the need for a makefile target which builds all charts. We could always write a script which invokes make with each of the charts if we needed to.

common.mk
```diff
@@ -0,0 +1,97 @@
...
+
+
+all: test build  ## Test then build chart
+build: init $(HELM_CHARTS_DIR)/docs/index.yaml $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)/$(ALL_YAML)  ## Build chart
+
+
+$(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)/$(ALL_YAML): $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz
+	$(call banner)
+	mkdir -p $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)
+	helm --namespace=$(NAMESPACE) template $< > $@
```
jsravn: Would it be a good time to remove all.yaml files? They make little sense for generic usage, because we can't parameterize them.

enterprise-suite-latest/Makefile
```diff
@@ -0,0 +1,40 @@
...
+SUBCOMPONENTS := $(wildcard ../$(BASE_CHART)/*/*/.)
+
+
+# Special rule to build 'latest' chart
+$(HELM_CHARTS_DIR)/docs/$(filter %-latest,$(CHART))-$(VERSION).tgz: $(COMPONENTS) $(SUBCOMPONENTS)
+	$(call banner)
+	rm -rf $(HELM_CHARTS_DIR)/build/$(CHART)
+	mkdir -p $(HELM_CHARTS_DIR)/build
+	cp -r ../$(BASE_CHART) $(HELM_CHARTS_DIR)/build/$(CHART)
+	$(SCRIPTS_DIR)/munge-to-latest.sh $(HELM_CHARTS_DIR)/build/$(CHART)
```
jsravn: Please move munge-to-latest.sh and munge-values-to-latest.yaml into this folder.

dbrinegar:
I've read through the issue and the discussion, and can certainly understand moving enterprise-suite things into its folder (no-brainer).  But what I'm not grokking is why anyone would use the top level thing.  I might need a quick walk through.

Is this just about managing a central index?

Are there other central needs?

Are the other teams on board with using `make`  or might they plug in their own thing?  I suppose they could have a `Makefile` that points to their own thing.  Is this the emergent part of this change, to see how much common Make stuff is wanted?

enterprise-suite/Makefile
```diff
@@ -1,27 +1,20 @@
...
-lint-helm:
-	helm lint .
-
 lint-json:
-	find . -name \*.json | xargs -tn 1 jq . >/dev/null
+	find $(CHART_DIR) -name \*.json | xargs -tn 1 jq . >/dev/null
 
 lint-promql:
-	scripts/validate-promql.sh
+	$(CHART_DIR)/scripts/validate-promql.sh
```
jsravn: Not clear to me why CHART_DIR is needed here? make should already be running from the chart directory.

common.mk
```diff
@@ -0,0 +1,97 @@
...
+
+clean::  ## Clean up.  (Removes all.yaml file)
+	rm -f $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)/$(ALL_YAML)
+
+minikube-test:
+
+help:  ## Print help for targets
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(lastword $(MAKEFILE_LIST)) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: all build test init lint-helm install-helm delete-local install-local clean minikube-test help
```
jsravn: I would probably make all these special targets enterprise-suite specific - we shouldn't require other charts to include this file  (it brings in a lot of stuff that only we care about or want, like all the install targets).

Makefile
jsravn: lint-helm should be in the ES Makefile. Just generic targets only in this top level Makefile.

marcoderama: The whole point of this exercise is to allow users to add helm-charts to some repository.  How is `lint-helm` ES-specific?  That would apply to all clients no?  (I did make `lint-json` and `lint-promql` ES-specific though.)

common.mk
marcoderama: Fine with me.  Makes some things easier.  (I could (try to) leave them in the enterprise-suite charts for now...)

common.mk
marcoderama: Sorry @jsravn, I don't understand this comment at all.  If we wanted to do those things with our chart, why wouldn't another user?  (Especially things like `build` and `test`.)  Without any of those targets, what's left and how would it be useful?  I'll followup to @dbrinegar's comment with more general verbiage about how I see this entire feature being used and how that influenced the design.  I sense we're not in agreement still.

enterprise-suite-latest/Makefile
marcoderama: It is hacky (and that's acknowledged in the comments).  What we're trying to do with `enterprise-suite` and `enterprise-suite-latest` is a hack and the requisite ugliness has to show up _somewhere_.  The current design limits the ugliness to `enterprise-suite-latest/Makefile` to the extent that it can.  (There is some very small leakage into `common.mk` but not in a way that a client would care/notice.)  (See for e.g. `reactive-sandbox/Makefile` or `sample-project/Makefile`.)

An alternative design would be to symlink everything from `enterprise-suite` into `enterprise-suite-latest` and then `enterprise-suite-latest` could be treated like "any other helm-charts client."  That felt _very_ ugly to me though (with confusing 'duplication', and the way Git handles symlinks).

If we went the copy-paste route, then I suspect we'd pretty much duplicate `common.mk` and then we'd risk fixing things in one place and not the other.  I'd argue strongly against going that route.

marcoderama:
Wrt @dbrinegar's questions and my discussion with @jsravn...

My basic premise is that we want other groups to be able to do the kinds of things with helm charts that we have been doing up to now.  That includes at a minimum running their charts through lint and templating, and getting them published.  We have also made extensive use of the `install-local` target for testing prior to committing.  To that end, the `common.mk` file provides the targets outlined in 
https://github.com/lightbend/es-backend/issues/315#issuecomment-416005161.  If that's all they want, then their Makefiles are as simple as the ones I show in `reactive-sandbox/Makefile` and `sample-project/Makefile`.

- Do we agree on this?

The top-level `Makefile` is intended for use by Travis at least.  (I also envisaged a possible "helm-charts" maintainer.)  As it stands, it allows one to run `make` (possibly with any of `all build clean test lint-helm` targets) and have them run over all the charts in `helm-charts`.  There's also a `build-index` target to rebuild `docs/index.yaml` if need be.  The `minikube-test` target exists for the sake of Travis.

- Do we agree on this?  (I think so.)

marcoderama:
The issue of creating "latest charts" for `enterprise-suite` is really independent of the feature-set discussed above.  It's a hack we want just for us.  One constraint I understood from discussions with @jsravn was that Travis be able to treat `enterprise-suite-latest` the same as `enterprise-suite`.  That is, one works with it the same as any other helm-chart, _with the exception_ of it using the 'latest' things.

The proposed changes do just that.  From the `enterprise-suite-latest` directory, one can run `make` with any of the common targets and things work as for any other chart, with the exception that they use the 'latest' things.  The cost of this feature is an atypical (aka hack) `Makefile` for `enterprise-suite-latest`.

enterprise-suite/Makefile
marcoderama: This is a concession to making things work for enterprise-suite-latest.

marcoderama:
@jsravn said:
> 1. Top level makefile should build only a single chart, that way we don't need to add complexity for looping through a list of charts. Looping logic should be done in .travis.yml or in a script.

I didn't make any changes to the top-level Makefile only cuz I wasn't sure what to do.  Do you want separate targets for each chart project, or require that some var (e.g. `CHART`) be specified on the command line?  (In which case, is there a default?)  (And it can already build a single chart...  Just sayin'.)

.travis.yml
```diff
@@ -16,13 +16,11 @@ jobs:
   - name: Minikube
     script:
     - scripts/setup-minikube-for-linux.sh
-    - make minikube-test
-    env: USE_LATEST_IMAGES=false
+    - make minikube-test CHARTS=enterprise-suite
```
marcoderama: In my previous incarnation of this file, I had one call to `make minikube-test` that let the `Makefile` run that on all the projects.  I've changed this to invoke `make minikube-test` on `enterprise-suite` and `enterprise-suite-latest` individually.  This is more in line with what you had and what I think you're looking for.  Works for me.  I agree that it clarifies the intent.

Makefile
marcoderama: Thought about this a bit more...  I think the `lint-helm` target makes sense in `common.mk` so that all clients can use it.  I realize now that your comment is on the occurrence in the top-level `Makefile` though.  I think it's reasonable there, so one could easily run it across all the projects at once, but I could be convinced to remove it from the top-level.

jsravn: We don't have any say in how other teams test their helm charts. We shouldn't be making anything common unless someone else wants it first. Premature optimization and all that.

enterprise-suite-latest/Makefile
jsravn: As long as this file is only required for ES charts, that is okay. My goal is to avoid having common dependencies across team boundaries, which makes that code very difficult to change. Also, imo, duplication in make is usually preferable to make madness (automake, config tools, macros, etc.).

jsravn:
I had an idea - should we just remove the "latest" chart? That should simplify things a lot. Personally I think we can live without it, it's not needed imo. For dev we can install directly from the master branch checkout, or use `--set esMonitorVersion=latest` type of thing.

marcoderama:
The most recent changes are close to done but still a WIP.  I thought it'd be good to give you a chance to look at them before polishing though.
- The top-level Makefile doesn't do much more than build `index.yaml` and run a test over things.
- Got rid of `common.mk` and moved most of that functionality back into the land of enterprise-suite (and enterprise-suite-latest).

This means that we expect users to build their tarballs and put them in the `docs` directory.  All they  really need to do is implement a `test` target.  (I may tweak `sample-project/Makefile` to demonstrate how to do that.)

You don't have to examine things too closely.  I'm more interested in your thoughts on the changes at a high-level.  (Things work though.)

enterprise-suite/Makefile
```diff
@@ -1,27 +1,91 @@
...
 
-install-local-latest:
-	helm install ../docs/$(RELEASE_LATEST).tgz --name=es --namespace=lightbend --debug --wait
+# This target required by top-level helm-charts Makefile.  Tests release on Travis
+test:
+	$(CHART_DIR)/scripts/run-minikube-tests.sh
 
-.PHONY: test lint lint-helm lint-json lint-promql minikube-test \
-	delete-local install-local install-local-latest
+# We no longer need a target to build the index.yaml file since that's up to the
```
marcoderama: I think including this would be useful.  Thoughts?

enterprise-suite/Makefile
```diff
@@ -1,27 +1,91 @@
...
+#
+# $(HELM_CHARTS_DIR)/docs/index.yaml: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz
+# 	$(call banner)
+# 	( cd .. ; $(MAKE) -C $(HELM_CHARTS_DIR) docs/index.yaml ; )
+
+init:
+	@$(SCRIPTS_DIR)/lib.sh
+	@helm init -c > /dev/null
+
+$(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)/$(ALL_YAML): $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz
```
marcoderama: Leaving the building of `all.yaml` in for now since the front-end guys apparently rely on it at this time.  We can easily yank it out later if need be.

README.md
```diff
@@ -4,57 +4,56 @@
...
-
-# or use "latest" container images
-kubectl --namespace=lightbend apply -f https://lightbend.github.io/helm-charts/es/all-latest.yaml
+helm repo update
+helm upgrade lightbend-helm-charts/<chart-name> --name=<release-name> --namespace=lightbend --debug
 ```
 
 ## Cutting a Release / Publishing Charts
 
+_This needs updating..._
```
marcoderama: This section needs work, but perhaps as part of a separate PR.

.travis.yml
```diff
@@ -10,19 +10,14 @@ install: scripts/setup-tools-for-ubuntu.sh
...
-    name: Validate
-    script:
-    - make test
-  - name: Minikube
+    name: Minikube
     script:
     - scripts/setup-minikube-for-linux.sh
-    - make minikube-test
-    env: USE_LATEST_IMAGES=false
+    - make test CHARTS=enterprise-suite
```
marcoderama: I'll leave doing these in parallel to another PR.

marcoderama:
So now...
- The top-level makefile invokes `package` and `test` targets recursively.  The default target builds `index.yaml` from the `.tqz` files.
- A `common.mk` file _may_ be used by a project to implement reasonable defaults for the `package` and `test` targets.
   - `package` creates the tarball and pushes to the `docs` directory.
   - `test` runs tests on the installed chart
- All the 'latest' messiness is limited to `enterprise-suite(-latest)`

enterprise-suite/Makefile
```diff
@@ -1,27 +1,95 @@
...
+	$(call banner)
+	mkdir -p $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)
+	helm --namespace=$(NAMESPACE) template $< > $@
+
+# Note: This works for enterprise-suite and allows us to define a latest-specific version in enterprise-suite-latest
+$(HELM_CHARTS_DIR)/docs/$(filter-out %-latest,$(CHART))-$(VERSION).tgz: $(COMPONENTS) $(SUBCOMPONENTS)
+	$(call banner)
+	helm package -d $(HELM_CHARTS_DIR)/docs $(CHART_DIR)
+
+lint: lint-helm lint-json lint-promql  ## Run all lint tests on chart files
```
jsravn: It doesn't look like we call `lint` anymore as part of the travis build? It should be invoked as part of tests.

Makefile
```diff
@@ -1,75 +1,40 @@
...
+# 
+# The 'common.mk' file can be used to implement these targets if desired.
+
+# Collection of charts to process.
+# These are subdirectories of helm-charts.  Add to the list as required.
+#
+# To package/test a subset, just define a value on the command line.
+#    e.g. "make package CHARTS=sample-project"
+CHARTS = enterprise-suite enterprise-suite-latest reactive-sandbox #sample-project
+
```
jsravn: I think it would be nice to have a top level `all` target here which packages everything. E.g. `all: package docs/index.yaml`. So make by default builds all the charts.

enterprise-suite/Makefile
marcoderama: After discussion with @jsravn, have decided that all client Makefiles need to provide a `lint` and a `test` target.  The intention is that the `lint` target just checks that things look good pre-packaging.  The `test` target can be used to run tests post-packaging.

PR #72, opened 2018-08-28
 Add console-server

srikanthops:
provide helm setting esServiceType to switch between ClusterIP and NodePort

this change is not 100% needed, 
however for GKE clusters, you cannot create ingress for ClusterIP service.. 
to make it easy for customers I added 
console-server service 
// Please comment on service name, as it will be referred in docs

Also provided helm setting esServiceType that will change service into clusterIP 

I am using esServiceType for other services too, grafana-server, prometheus-server,  es-monitor-api
I did that because, to make life easy in smoke-test I just have to create ingress for these services 

connect to https://github.com/lightbend/es-backend/issues/270


enterprise-suite/templates/es-grafana.yaml
```diff
@@ -60,7 +60,7 @@ spec:
   selector:
     app: grafana
     component: "server"
-  type: "ClusterIP"
+  type: {{ .Values.esServiceType }}
```
dbrinegar: this will break the proxy, which works on the cluster network

enterprise-suite/templates/prometheus.yaml
```diff
@@ -164,6 +164,7 @@ spec:
   selector:
     app: prometheus
     component: server
+  type: {{ .Values.esServiceType }}
```
dbrinegar: bug, as above

enterprise-suite/templates/prometheus.yaml
```diff
@@ -177,6 +178,7 @@ spec:
...
```
dbrinegar: bug, as above

enterprise-suite/values.yaml
```diff
@@ -23,3 +23,6 @@ imagePullPolicy: IfNotPresent
 prometheusDomain: prometheus.io
 # minikube debug resources
 minikube: true
+# console server type
+esServiceType: ClusterIP
```
dbrinegar: I've said this many times, but bear with me as I repeat:

    Please treat NodePort as a security hole.

We should _not_ install it.  If customer needs to install it for some vendor specific reason in a circumscribed manner and on their (not our) recommendation, hey they can do that.  But we should treat it as an electric fence not to be touched.

https://oteemo.com/2017/12/12/think-nodeport-kubernetes/

jsravn:
@dbrinegar @srikanthops is this ready to merge? Can we merge it and close the associated issue?

dbrinegar:
I approved this one 7 days ago

jsravn:
Please merge it @srikanthops

dbrinegar:
@srikanthops is there some reason you aren't merging this PR?  I'm going to cut a release shortly...

PR #73, opened 2018-08-28
WIP Slack msg

dbrinegar:
Fix slack message to use `elapsed_time` instead of total `duration` of all builds in matrix.

Since can only specify one template for all types of builds, using the lowest common denominator branch build template as a starting point.  This means it does not mention PR numbers, but the commit link will still take you to the PR on those builds.

https://docs.travis-ci.com/user/notifications/#customizing-slack-notifications


jsravn:
@dbrinegar yup it seems like we have to encrypt the whole config. Sorta stupid. Oh well. See https://docs.travis-ci.com/user/notifications/#configuring-slack-notifications - need to do it using the `travis encrypt` command.

dbrinegar:
@jsravn could you show me how to do the travis encrypt command?  If I did it I would likely change something you've already set previously, I think.  I need to know what went into generating the existing `secure` value.

Thanks!

jsravn:
@dbrinegar yes, of course. Forgot that you need the slack token - will do it offline.

jsravn:
I've done this in https://github.com/lightbend/es-monitor-api/pull/68.

marcoderama:
Hmm.  I approved but just realized this is marked as WIP.  Still a WIP?

dbrinegar:
ah sorry, left this in the bin, dunno how to do produce the magic string.  Thanks for implementing @jsravn

PR #74, opened 2018-08-29
Play health

dbrinegar:
latency monitors
* modeled after akka processing time: sma, 50th, 100ms minval
* akka http server response time
* akka http client response time
* play http client response time

error monitors
* akka http 5xx errors

fixes lightbend/es-backend#77


enterprise-suite/es-monitor-api/bootstrap-monitors.json
```diff
@@ -571,6 +571,69 @@
...
+        "monitorType": ""
+      }
+    },
+    {
+      "model": "sma",
+      "parameters": {
+        "name": "akka_http_client_response_time",
+        "metric": "akka_http_http_client_http_client_service_response_time_ns{quantile=\"0.5\"}",
+        "period": "15m",
+        "numsigma": "2",
```
jsravn: How are these values chosen?

dbrinegar: following akka_processing_time

is a very conservative:  numsigma 2 means 5% error, 15 minutes is our slow window (5, 10, 15), 100ms is our ignore line, ignoring anomalies under 100ms

jsravn: Is it documented anywhere how the values were attained?

Just thinking about how we would explain these things to customers. And also how we add new monitors with sensible values. Why 5% standard error? Is median a good indicator of poor health on response times (which are usually long tailed)?

enterprise-suite/es-monitor-api/bootstrap-monitors.json
```diff
@@ -571,6 +571,69 @@
...
+        "monitorType": ""
+      }
+    },
+    {
+      "model": "threshold",
+      "parameters": {
+        "name": "akka_http_server_5xx",
+        "metric": "akka_http_http_server_responses_5xx",
+        "comparator": ">",
+        "threshold": "0",
```
jsravn: Whoops this should have been rate based.

PR #76, opened 2018-08-30
Add gke test support/srikanth

srikanthops:
Add smoke tests for GKE 
The main problem with GKE is creating ingress and waiting for them.. 
some times it take hell lot of time.. 

Once the ingress is created and worked. the tests pass nicely

### DIFF
--- a/enterprise-suite/tests/resources/gke-ingress-test.template.yaml
+++ b/enterprise-suite/tests/resources/gke-ingress-test.template.yaml
@@ -1,0 +1,9 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{INGRESS_NAME}}
+  namespace: lightbend
+spec:
+  backend:
+    serviceName: {{SERVICE_NAME}}
+    servicePort: {{SERVICE_PORT}}
--- a/enterprise-suite/tests/resources/gke-ingress.yaml
+++ b/enterprise-suite/tests/resources/gke-ingress.yaml
@@ -1,0 +1,38 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: expose-es-console-ing
+spec:
+  backend:
+    serviceName: expose-es-console
+    servicePort: 80
+
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: expose-prometheus-ing
+spec:
+  backend:
+    serviceName: expose-prometheus
+    servicePort: 9090
+
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: expose-grafana-ing
+spec:
+  backend:
+    serviceName: expose-grafana
+    servicePort: 3000
+
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: expose-alertmanager-ing
+spec:
+  backend:
+    serviceName: expose-alertmanager
+    servicePort: 9093
--- a/enterprise-suite/tests/smoke_ingress
+++ b/enterprise-suite/tests/smoke_ingress
@@ -4,9 +4,7 @@
 
 #set -x
 
-if [[ ! $context  ]]; then
-  source smokecommon
-fi
+source smokecommon
 
 if [[ $context == "GKE" ]]; then exit 0; fi
 
--- a/enterprise-suite/tests/smoke_ingress
+++ b/enterprise-suite/tests/smoke_ingress
@@ -4,7 +4,11 @@
 
 #set -x
 
-source smokecommon
+if [[ ! $context  ]]; then
+  source smokecommon
+fi
+
+if [[ $context == "GKE" ]]; then exit 0; fi
 
 TMPPIPE=$(mktemp -t pipe.XXX)
 
--- a/enterprise-suite/tests/smoke_port-forward
+++ b/enterprise-suite/tests/smoke_port-forward
@@ -4,9 +4,8 @@
 
 #set -x
 
-if [[ ! $context  ]]; then
-  source smokecommon
-fi
+source smokecommon
+if [[ $context == "GKE" ]]; then init_gke_ingress ; fi
 
 KUBECTL_PID=
 export TMPPIPE=$(mktemp -t pipe.XXX)
--- a/enterprise-suite/tests/smoke_port-forward
+++ b/enterprise-suite/tests/smoke_port-forward
@@ -4,7 +4,9 @@
 
 #set -x
 
-source smokecommon
+if [[ ! $context  ]]; then
+  source smokecommon
+fi
 
 KUBECTL_PID=
 export TMPPIPE=$(mktemp -t pipe.XXX)
--- a/enterprise-suite/tests/smokecommon
+++ b/enterprise-suite/tests/smokecommon
@@ -16,11 +16,23 @@ error() {
   exit 1
 }
 
+context=minikube
+if [[ $(kubectl config current-context) == gke_* ]] ; then
+  context="GKE"
+fi
+
+
 # busy_wait cmd
 # quickly retry command til it works, up to some limit
 # idea: succeed fast, fail slow
 busy_wait() {
-  limit=90
+  if [[ $1 == limit=* ]]; then
+   limit="${1#*=}"
+   shift
+  else
+    limit=90
+  fi
+
   until "$@"; do
     if [[ limit -le 0 ]]; then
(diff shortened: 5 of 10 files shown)


enterprise-suite/tests/smokecommon
```diff
@@ -112,3 +128,48 @@ test_alertmanager_responding() {
...
+  fi
+}
+
+# Get ingress IP
+# Return http://[[public-ip]]
+get_ing_ip() {
+  kubectl get ing "$1" -n lightbend -o json | jq -e '.status.loadBalancer.ingress[0].ip' | tr -d '"'| awk '$1 != "null" {print "http://"$1}'
+}
+
+init_gke_ingress() {
```
marcoderama: Does this need a different name?  I see it's used in lots more tests than just the ingress ones.

enterprise-suite/tests/smokecommon
```diff
@@ -112,3 +128,48 @@ test_alertmanager_responding() {
...
+  if [[ $context == "GKE" ]]; then
+    kubectl apply -f resources/gke-ingress.yaml
+    ES_CONSOLE_IP=$(busy_wait limit=180 get_ing_ip expose-es-console-ing)
+    busy_wait limit=180 test_es_console_responding $ES_CONSOLE_IP
+  fi
+}
+
+if [[ $context == "GKE2" ]]; then
```
marcoderama: When is context `GKE2` (as opposed to `GKE`)?

enterprise-suite/tests/smokecommon
```diff
@@ -112,3 +128,48 @@ test_alertmanager_responding() {
   results=$( curl_query $1 "" | fgrep -q '<title>Alertmanager</title>' )
   T $? alertmanager accessible via $1
 }
+
+# Create services-name-ing ingress if not present
```
marcoderama: Can these GKE-specific functions have 'gke' added to their names somewhere?

enterprise-suite/tests/smoke_ingress
```diff
@@ -6,6 +6,8 @@
 
 source smokecommon
 
+if [[ $context == "GKE" ]]; then exit 0; fi
```
marcoderama: I'm confused.  The rest of these changes seem to be related to GKE ingress things but both ingress tests are disabled if we're using GKE?

enterprise-suite/tests/resources/gke-ingress.yaml
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  backend:
+    serviceName: expose-es-console
+    servicePort: 80
+
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: expose-prometheus-ing
```
marcoderama: Do these expose-X things need a `NodePort` specified?  I'm confused.  Weren't these added specifically because we needed to use a NodePort?  If so, where is the NodePort referenced?  Maybe I'm missing something.

enterprise-suite/tests/smokecommon
```diff
@@ -112,3 +128,48 @@ test_alertmanager_responding() {
...
+createIngress() {
+  if ! kubectl get ing -n lightbend "$1-ing"  > /dev/null 2>&1 ; then
+    sed -e "s/{{INGRESS_NAME}}/$1-ing/g;s/{{SERVICE_NAME}}/$1/g;s/{{SERVICE_PORT}}/$2/g" \
+            < resources/gke-ingress-test.template.yaml | kubectl apply -n lightbend -f -
+    echo "create $1-ing for service $1 at port $2"
+    busy_wait limit=180 get_ing_ip "$1-ing"
```
dbrinegar: not seeing the need for `limit=*` here, just bump busy_wait to 180

enterprise-suite/tests/smokecommon
```diff
@@ -16,11 +16,22 @@ error() {
...
+if [[ $(kubectl config current-context) == gke_* ]] ; then
+  context=GKE
+fi
+
 # busy_wait cmd
 # quickly retry command til it works, up to some limit
 # idea: succeed fast, fail slow
 busy_wait() {
-  limit=90
+  if [[ $1 == limit=* ]]; then
```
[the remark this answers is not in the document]
dbrinegar: if you want to optionally parameterize a bash function, prefix the options with `--` so they are less likely to accidentally eat something that starts with "limit"


but in this case we don't need it, just change 90 to 180 if the need is to wait that long.  busy_wait is the upper limit, only slows down the broken / failing test

enterprise-suite/tests/smokecommon
```diff
@@ -16,11 +16,22 @@ error() {
   exit 1
 }
 
+context=minikube
```
dbrinegar: This is global, so you're repeating this `kubectl config current-context` command for every job.  At best that's inefficient, at worst that's inconsistent results. This should be established once, and kept.

enterprise-suite/tests/resources/gke-ingress.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+---
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: expose-alertmanager-ing
+spec:
+  backend:
+    serviceName: expose-alertmanager
+    servicePort: 9093
```
dbrinegar: omg new lines

enterprise-suite/tests/resources/gke-ingress-test.template.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{INGRESS_NAME}}
```
dbrinegar: why are we setting up so many ingresses?  Just setup one for nginx and use it

enterprise-suite/tests/smoke_ingress
dbrinegar: yeah I tend to agree that this seems odd.  Why wouldn't we run smoke_ingress etc in gke?

enterprise-suite/tests/smoke_port-forward
```diff
@@ -5,6 +5,7 @@
 #set -x
 
 source smokecommon
+if [[ $context == "GKE" ]]; then init_gke_ingress ; fi
```
dbrinegar: Whatever init gke needs should be done in a separate script, call it `gke_init`, and should be done once before anything else is done, at which point all the tests kick off.

What you've done is ask for init_gke_ingress to be started simultaneously by every test.  no bueno

enterprise-suite/tests/smokecommon
```diff
@@ -62,10 +73,15 @@ test_summary() {
...
 
+
 # Get URL for NodePort assignment
 # nodeport <name>
 nodeport() {
-  minikube service --namespace=lightbend --url $1
+  if [[ $context == "minikube" ]]; then
+    minikube service --namespace=lightbend --url $1
+  elif [[ $context == "GKE" ]]; then
+    get_ing_ip $1-ing
```
dbrinegar: else ....  case please, we should error out if in an unsupported context

dbrinegar: since this is a global and about kubernetes, let's call this `$KubetnetesContext` instead of `$context`

enterprise-suite/tests/smokecommon
```diff
@@ -112,3 +128,48 @@ test_alertmanager_responding() {
...
+  if ! kubectl get ing -n lightbend "$1-ing"  > /dev/null 2>&1 ; then
+    sed -e "s/{{INGRESS_NAME}}/$1-ing/g;s/{{SERVICE_NAME}}/$1/g;s/{{SERVICE_PORT}}/$2/g" \
+            < resources/gke-ingress-test.template.yaml | kubectl apply -n lightbend -f -
+    echo "create $1-ing for service $1 at port $2"
+    busy_wait limit=180 get_ing_ip "$1-ing"
+  fi
+}
+
+# Get ingress IP
+# Return http://[[public-ip]]
```
dbrinegar: no `https` path here?  oops

enterprise-suite/tests/smokecommon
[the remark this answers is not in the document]
srikanthops: I will change it to 180

srikanthops:
Thanks for all your comments.. There is more then anticipated here.. 

- [ ] automatically set up Google Cloud
- [ ] ingress is taking lot of time to create, so use port-forward 
- [ ] include code review comments

PR #78, opened 2018-08-31
default resource requests

dbrinegar:
part of lightbend/es-backend#224

defaultCPURequests     1/10th of a cpu
defaultMemoryRequests  256 meg

used as the default for every container.  Excluding init containers,
every container has a fine grained helm option for CPU and Memory.

alertManager
configmapReload
esConsole
esGrafana
kubeStateMetrics
nodeExporter
esMonitor
prometheus

in case someone wants to override the default in a specific instance.

### DIFF
--- a/enterprise-suite/templates/resourcelimits.yaml
+++ b/enterprise-suite/templates/resourcelimits.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: {{.Release.Namespace}}-namespace-defaults
+spec:
+  limits:
+  - defaultRequest:
+      cpu: {{ .Values.defaultCPURequest }}
+      memory: {{ .Values.defaultMemoryRequest }}
+    type: Container
--- a/enterprise-suite/templates/resourcelimits.yaml
+++ b/enterprise-suite/templates/resourcelimits.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: LimitRange
 metadata:
-  name: {{.Release.Namespace}}-namespace-defaults
+  name: {{ .Release.Namespace }}-namespace-defaults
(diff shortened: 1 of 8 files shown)


dbrinegar:
super curious if this is resource constraining for travis minikube :-|

dbrinegar:
Travis likes them!  yay

enterprise-suite/templates/resourcelimits.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: {{.Release.Namespace}}-namespace-defaults
```
marcoderama: You've put spaces after/before the braces everywhere else...

jsravn:
@dbrinegar 256Mi seems very large. That's going to fill up small clusters quickly, isn't it, like minikube? Remember this will reserve that space on the node, so other things can't deploy. I can't imagine any of our components, except prometheus, uses more than 20-50Mi of memory.

enterprise-suite/templates/resourcelimits.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: LimitRange
+metadata:
+  name: {{ .Release.Namespace }}-namespace-defaults
+spec:
+  limits:
+  - defaultRequest:
+      cpu: {{ .Values.defaultCPURequest }}
+      memory: {{ .Values.defaultMemoryRequest }}
+    type: Container
```
jsravn: Isn't this redundant with specifying the default request on each pod?

dbrinegar: Probably!  Just leaving as a kind of bookmark in case someone wants a thing there.

PR #80, opened 2018-09-05
shrink default mem, except prometheus

dbrinegar:
Setting to 50M per recommendation from @jsravn

On a single node minikube install doing nothing, max memory usage is

kube-state:  19M
node-exporter: 8M
prometheus: 204M
alert-manager: 8M
es-console: 2M
grafana: 33M

and container_spec_cpu_shares is near zero.  Prometheus and Alertmanager are 4.


dbrinegar:
I think the idea is to be pretty close to the minimum.  This won't stop it from taking more, but does stop it if less than the request is available.

PR #81, opened 2018-09-06
release 0.9.0 latest

dbrinegar:


jsravn:
If we were using semver, this should be a 1.x.x.

jsravn:
Or version `enterprise-suite-latest` separately from `enterprise-suite` @marcoderama .

PR #82, opened 2018-09-07
Use commercial registry for es-monitor-api

jsravn:
* Provides an install script `install-es.sh` which looks up
credentials from `~/.lightbend/commercial_credentials` by default, and
provides an easy install option (to be documented at es-docs).

* The credentials can also be passed directly to helm:

  `--set imageCredentials.username=username,imageCredentials.password=pass`

* all.yaml excludes the secret by default, which must be loaded
manually:
  
  ```bash
  kubectl -n lightbend create secret docker-registry \
    commercial-credentials \
    --docker-server=lightbend-docker-commercial-registry.bintray.io \
    --docker-username=mybintrayuser \
    --docker-password=mybintraytoken
  ```


dbrinegar:
Phew, so this looks very nice, but I have some observations that I'm hoping might prod things to be even nicer.

1. Passing login on command line is insecure, as there is now a password in plain text either in process or env text that can be read by other users.  Passwords should ideally be secured in a file, allowing file access control to dictate who can see it, and programs that need those should read those files for the values.  Note that this lesson is learned over and over, and as an example the Docker CLI has security warnings about the legacy deprecated option of passwords on the command line.

2. Requiring command line options is inconvenient, meaning there is no static command that can just "do it" and let customization evolve in steps.  Instead, these things must be provided every time one wants to use it.  Devs will be making tons of hacks around this, I'm sure, as it is too inconvenient to type the password over and over.  We should look at what devs must do with Cinnamon, I think, to do some one-time lay in of options, then you just use the normal commands and off you go.  Set and forget.

It seems to me that some path to reading the bintray credentials that one must already have to use Cinnamon, from `~/.lightbend/commercial.credentials` or wherever the user specifies, to populate these values would be more ideal.

Is it possible for helm to be given a credentials file, then go read and parse it into the values we need?

enterprise-suite/Makefile
```diff
@@ -84,7 +84,15 @@ delete-local:  ## Delete chart from cluster with helm
...
 install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart
-	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait
+	@if test -z "$(BINTRAY_USERNAME)"; then \
+		echo BINTRAY_USERNAME required \
+	fi
+	@if test -z "$(BINTRAY_PASSWORD)"; then \
+		echo BINTRAY_PASSWORD required \
+	fi
+	echo "Installing helm charts"
+	@helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait \
```
marcoderama: Does this have the leading '@' to prevent leakage when `imageCredentials.password=$(BINTRAY_PASSWORD)` is printed?  Otherwise, I think it's nice to see this...

enterprise-suite/Makefile
```diff
@@ -84,7 +84,15 @@ delete-local:  ## Delete chart from cluster with helm
 	-helm delete --purge $(RELEASE_NAME)
 
 install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart
-	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait
+	@if test -z "$(BINTRAY_USERNAME)"; then \
```
marcoderama: Since these are dev-related targets, do we want to try pulling them from some file if not explicitly specified?

enterprise-suite/values.yaml
```diff
@@ -28,6 +28,11 @@ imagePullPolicy: IfNotPresent
 prometheusDomain: prometheus.io
 # minikube debug resources
 minikube: true
+# commercial image credentials - username and password are required
+imageCredentials:
+  registry: lightbend-docker-commercial-registry.bintray.io
+  username: setme
+  password: setme
```
marcoderama: Is this encrypted or something?  This seems dangerous otherwise but maybe I'm missing something.

jsravn:
@dbrinegar I agree with all the things you said, unfortunately helm doesn't support what you describe (and I did check first) - the best it offers is you can pass a file via `-f` with values in it. I think we would need to provide some sort of wrapper script to simplify the install process, including loading commercial credentials. But I wanted to wait because I wasn't sure the best approach to do that.

enterprise-suite/values.yaml
jsravn: What should be encrypted?

enterprise-suite/Makefile
jsravn: Yes that's why.

jsravn:
I'll put together an ES install script as part of this PR, that will read from a credentials file.

jsravn:
I'm going to hold off onto this until we come to some agreement on the path forward. Let me know your thoughts on above @dbrinegar. @alanngai maybe has an idea too on how to best support this.

jsravn:
@dbrinegar Please have another look, I've addressed the changes we discussed yesterday.
Also @marcoderama @srikanthops .

enterprise-suite/Makefile
```diff
@@ -91,10 +91,7 @@ delete-local:  ## Delete chart from cluster with helm
 	-helm delete --purge $(RELEASE_NAME)
 
 install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart
-	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait --set minikube=$(MINIKUBE)
-
-#
-##
+	ES_LOCAL_CHART=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz ES_MINIKUBE=true ./install-es.sh
```
marcoderama: I'd rather see `ES_MINIKUBE=$(MINIKUBE)` in case someone overrides the default.

enterprise-suite/install-es.sh
```diff
@@ -0,0 +1,103 @@
...
+}
+
+# User overridable variables.
+LIGHTBEND_COMMERCIAL_USERNAME=${LIGHTBEND_COMMERCIAL_USERNAME:-}
+LIGHTBEND_COMMERCIAL_PASSWORD=${LIGHTBEND_COMMERCIAL_PASSWORD:-}
+LIGHTBEND_COMMERCIAL_CREDENTIALS=${LIGHTBEND_COMMERCIAL_CREDENTIALS:-$HOME/.lightbend/commercial_credentials}
+ES_REPO=${ES_REPO:-https://lightbend.github.io/helm-charts}
+ES_CHART=${ES_CHART:-enterprise-suite}
+ES_NAMESPACE=${ES_NAMESPACE:-lightbend}
+ES_MINIKUBE=${ES_MINIKUBE:-false}
```
marcoderama: I should change enterprise-suite/Makefile:75 to use this style...

enterprise-suite/install-es.sh
```diff
@@ -0,0 +1,103 @@
...
+        repo_username=$LIGHTBEND_COMMERCIAL_USERNAME
+        repo_password=$LIGHTBEND_COMMERCIAL_PASSWORD
+    elif [ -e $LIGHTBEND_COMMERCIAL_CREDENTIALS ]; then
+        while IFS='=' read -r key value; do
+            local trimmed_key=$(echo -e "$key" | tr -d '[:space:]')
+            local trimmed_value=$(echo -e "$value" | tr -d '[:space:]')
+            if [ "$trimmed_key" == "user" ]; then
+                repo_username=$trimmed_value
+            fi
+            if [ "$trimmed_key" == "password" ]; then
```
marcoderama: elif?

enterprise-suite/install-es.sh
```diff
@@ -0,0 +1,103 @@
...
+    docvar ES_NAMESPACE "Namespace to install ES-Console into"
+    docvar ES_MINIKUBE "Set to true to enable minikube specific options"
+    docvar ES_LOCAL_CHART "Set to location of local chart tarball"
+    docvar ES_UPGRADE "Set to true to perform a helm upgrade instead of an install"
+    docvar ES_VERSION "Set to non-empty to install a specific version"
+    exit 1
+}
+
+function export_credentials() {
+    # Exported variables.
```
marcoderama: Where are these exported?  Doesn't matter but makes me think about the name/comment.

enterprise-suite/values.yaml
```diff
@@ -38,3 +43,9 @@ defaultCPURequest: 100m
 defaultMemoryRequest: 50Mi
 # Prometheus resource request for memory
 prometheusMemoryRequest: 250Mi
+
+##################################################
+# Internal variables - should not be set by users.
+#
+# Enable credentials secret generation, disabled for all.yaml generation.
+internalEnableCredentialsSecret: true
```
marcoderama: Don't let @dbrinegar see this missing newline!

enterprise-suite/install-es.sh
jsravn: I don't think this works in `make`, it's bash specific.

jsravn:
* Provides an install script `install-es.sh` which looks up
credentials from `~/.lightbend/commercial_credentials` by default, and
provides an easy install option (to be documented at es-docs).

* The credentials can also be passed directly to helm:

  `--set imageCredentials.username=username,imageCredentials.password=pass`

* all.yaml excludes the secret by default, which must be loaded
manually:
  
  ```bash
  kubectl -n lightbend create secret docker-registry \
    commercial-credentials \
    --docker-server=lightbend-docker-commercial-registry.bintray.io \
    --docker-username=mybintrayuser \
    --docker-password=mybintraytoken
  ```

PR #85, opened 2018-09-10
Remove LimitRanges, make tiller namespace configurable

jsravn:
A couple small tweaks that came out of playing with openshift.

LimitRanges are restricted on openshift by default, and we don't really
need them. So I removed them.

Also made the tiller namespace configurable - because by default in
openshift, users can't view kube-system so they can't use tiller at all.
This will be the same in other locked down installations. So we should
probably move towards testing tiller outside of kube-system. I left it
as kube-system, because helm defaults to kube-system and it's cumbersome
to use in another namespace.

For https://github.com/lightbend/es-backend/issues/320


jsravn:
There aren't any functional changes here, it's just small fixes.

PR #89, opened 2018-09-12
Fix es-console port to be 8080 so it works on openshift

jsravn:
part of https://github.com/lightbend/es-backend/issues/331


jsravn:
Going to include this in https://github.com/lightbend/helm-charts/pull/90 since I need it there to make the build pass.

PR #90, opened 2018-09-12
Allow setting podUID to run as non-root

jsravn:
By default it is unset, as there is no sensible default for this.

We set it in our smoketests however, so we can verify all our pods can
run as non-root to support Openshift.

part of https://github.com/lightbend/es-backend/issues/331

### DIFF
--- a/enterprise-suite/Makefile
+++ b/enterprise-suite/Makefile
@@ -90,8 +90,9 @@ install-helm:  ## Install required helm components into cluster
 delete-local:  ## Delete chart from cluster with helm
 	-helm delete --purge $(RELEASE_NAME)
 
-install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart
-	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait --set minikube=$(MINIKUBE)
+install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart, used by smoketests.
+	# We set a podUID here for test purposes to ensure everything works as non-root.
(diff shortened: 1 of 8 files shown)


jsravn:
It's failing across all the PRs for some reason. I've disabled the parallel smoketests (obfuscates the output) and added more debug output to try and figure it out. I also sent a request to travis to enable the debug mode for helm-charts.

marcoderama:
Nothing obvious to me as to why things would start failing.  All looks reasonable to me.

enterprise-suite/Makefile
```diff
@@ -90,8 +90,9 @@ install-helm:  ## Install required helm components into cluster
 delete-local:  ## Delete chart from cluster with helm
 	-helm delete --purge $(RELEASE_NAME)
 
-install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart
-	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait --set minikube=$(MINIKUBE)
+install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local chart, used by smoketests.
+	# We set a podUID here for test purposes to ensure everything works as non-root.
+	helm install $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz --name=$(RELEASE_NAME) --namespace=$(NAMESPACE) --debug --wait --set minikube=$(MINIKUBE),podUID=10001
```
dbrinegar: don't we need to make the 10001 user?  I'd expect the runAsUser to fail if the uid doesn't point to a user...

just taking a wild guess

dbrinegar: seems to runAsUser with any uid, at least here

jsravn: You can use any uid without a mapping in /etc/passwd.

jsravn:
@dbrinegar can you approve it please?

PR #96, opened 2018-09-14
Publish to GCS.  docker-based solution for travis or command line

marcoderama:


scripts/deploy-to-gcs.sh
```diff
@@ -0,0 +1,83 @@
...
+# Use caution with the -d flag.  If you happen to run this against the production repo but from a local
+# helm-charts dir with several tarballs missing for some reason, those tarballs will be removed from the
+# production repo.  Deletion is disabled by default.
+
+# This implementation inspired by https://medium.com/google-cloud/google-cloud-sdk-dockerfile-861a0399bbbb
+
+usage () {
+    echo "${0##*/} [-d] [-n] | [-h]"
+    echo "  -d   Delete objects in repo that are not in docs dir.  Default is false."
+    echo "  -n   Dry run.  Show what would be published.  Default is false"
```
srikanthops: I would recommend confirmation by default and a force flag [-y] to override

scripts/deploy-to-gcs.sh
```diff
@@ -0,0 +1,83 @@
...
+    echo "  -h   Print this help."
+    exit 1
+}
+
+# GCS project
+GCS_PROJECT=es-repo
+# GCS bucket within project
+: ${GCS_BUCKET:=marcoderama-test}
```
srikanthops: Is this the final name...
Can we choose bucket names like production and stating 
es-producution
es-staging

Another thing that confuses me the right to use the word "es",  Almost all commercial offerings from LB are under enterprise-suite.

scripts/deploy-to-gcs.sh
```diff
@@ -0,0 +1,83 @@
...
+
+docker pull google/cloud-sdk:${CLOUD_SDK_VERSION}
+
+# Use the decrypted service account credentials to authenticate the command line tool and setup
+# volume mounts so credentials and tarballs are accessible to docker
+GCLOUD_CONFIG_CID=$( docker run -t -d \
+     -v ${HELM_DIR}/resources:/resources \
+     -v ${HELM_DIR}/docs:/docs \
+     --name gcloud-config ${CLOUD_SDK_IMAGE} \
+     gcloud auth activate-service-account --key-file /resources/es-repo-7c1fefe17951.json \
```
srikanthops: curious why the keyfile name is so cryptic..

marcoderama:
The `.travis.yml` mods are such that the `deploy-to-gcs.sh` script won't be invoked for this pull request.  For a look at how it works, see https://travis-ci.org/lightbend/helm-charts/builds/428839319.  That was a test run from before the push-only travis mod went in.

scripts/deploy-to-gcs.sh
marcoderama: I'd rather not, since this script could be called in a non-interactive fashion from Travis if we choose to go that route.  I guess I could check if the shell is interactive (or for a Travis env var).  Will consider...

scripts/deploy-to-gcs.sh
marcoderama: That's a temporary bucket name.  I don't want to use an 'official' one until all the other pieces are in place.  e.g. HTTPS redirect.  I don't know if we'll have staging and production versions.  (In the mean time, we'll publish to this private bucket.  Seemed okay to me.)

The project name was chosen by Karl.  Are you suggesting that should change, or just the bucket name?

scripts/deploy-to-gcs.sh
marcoderama: That's as generated by GCS.  I considered changing the name but figured there may be some value in leaving it as is.

marcoderama:
Note also that these two secure env vars have been added to the Travis configs:
```
encrypted_f01ffbb90c44_iv
encrypted_f01ffbb90c44_key
```

scripts/deploy-to-gcs.sh
srikanthops: what about executing -n dry run always so that we can later look at logs to see what got synced

marcoderama: Not sure I follow.  If you don't specify `-n`, the logs will still contain a list of what got synced.  See for e.g., https://github.com/lightbend/helm-charts/pull/96, where I artificially tweaked the `index.yaml` file to see if (only) it got picked up correctly.

Or did you mean something else?

.travis.yml
```diff
@@ -24,6 +24,14 @@ jobs:
   - name: Reactive-sandbox test
     script:
     - make test CHARTS=reactive-sandbox
+  - stage: deploy
+    name: Deploy to GCS
+    # Only deploy on a push to master
+    if: type = push
```
jsravn: I'm not sure this is right. See https://github.com/lightbend/es-console-spa/blob/master/.travis.yml#L24 for what I used in es-console-spa.

jsravn: Hmm I guess this is the same thing as `(NOT type IN (pull_request))` ?

scripts/deploy-to-gcs.sh
jsravn: Can we set this to something a bit more generic like `helm-charts-test`. Then in `.travis.yml` we can override it with `helm-charts` as the bucket name.

.travis.yml
marcoderama: I thought that test was appropriate, given 
https://docs.travis-ci.com/user/conditions-v1#integration
We only want to publish on a `push` right?  (I don't really know what`api` or `cron` Travis events are though.)

marcoderama:
@jsravn asks:
> I thought we wanted to remove docs folder completely, at least from version control. Are you trying to do things iteratively by preserving the old behavior?

I'll go with the 'iterative' option, even though that wan't intentional...  For now at least, this does preserve existing behavior.  If we yank `docs` from git, how do we backup our tarballs (in case someone somehow accidentally deletes everything in GCS)?  I guess we could publish to two buckets, and hope nobody blows away both...  Thoughts?

scripts/deploy-to-gcs.sh
marcoderama: Unfortunately bucket names need to be globally unique and `helm-charts-test` is already taken.  `lb-helm-charts-test` and `lightbend-helm-charts-test` are available.  I'll go with the latter unless someone suggests otherwise.

(For that matter, I guess it's not entirely 'correct' that these live in the `es-repo` project, since the charts are not necessarily `es`-related.  Not sure if I have the power to create a new repo or not.  Looks like it from the UI.  That doesn't seem to get much (any?) external exposure though so I'd guess we can probably live with `es-repo`.  $0.02)

scripts/deploy-to-gcs.sh
srikanthops: If the logs contain a list of files that got synced.. thats all I need thanks.

.travis.yml
jsravn: api is if someone triggers the build manually. Cron is if we enable cron runs. We probably want it building on manual trigger, right? In case we need to rerun it.

jsravn:
>  If we yank docs from git, how do we backup our tarballs (in case someone somehow accidentally deletes everything in GCS)? I guess we could publish to two buckets, and hope nobody blows away both... Thoughts?

@marcoderama Just enable versioning on the GCS bucket, it will save every version of the bucket then. Alternatively, we can always rebuild from version control. I think we need to get away from having to build the tarball and check it in, only then to have it publish externally - it's redundant/error prone.

marcoderama:
This should be ready for the big time now.  Most recent changes...

- Run the deploy script for `push` and `api` invocations.
- Changed to publish to the non-test bucket.  `gs://lightbend-helm-charts`
- Set versioning up on that bucket so we can recover from accidental deletes.
- Also setup an "object lifecycle rule" that should mean we keep no more than 3 versions of an object.  In general, we should only ever see multiple versions of the `index.yaml` file anyway and that's trivial to reproduce.
   - With versioning setup, you kind of have to go out of your way to delete all versions of an object so I think we're covered.  I tested using 'normal' deletion of an object (with only a single version).  It was trivial to recover it.

marcoderama:
@jsravn and I agreed that we can leave things in the `docs` folder for now.  This actually facilitates an easy transition from one to the other.  Once this PR is merged, the `gs://lightbend-helm-charts` bucket should track the `docs` folder (minus `all*.yaml` files).  We can then phase in the transition to installing from `https://repo.lightbend.com/helm-charts` instead of `https://lightbend.github.io/helm-charts`, eventually dropping the latter.

PR #97, opened 2018-09-18
tweaks on credential file, install-local

dbrinegar:
Cinnamon documents this file with a dot, not an underbar. This changes
to be consistent with that file name.

Also changes `install-local` to install from the folder contents instead
of a tarball, as what you want to install are your local file changes
and not some old tarball by mistake.  This also eliminates the constant
rewriting of a previous release.


dbrinegar:
didn't realize `install-local` was critical path to the travis test, prolly need to make a separate target

enterprise-suite/install-es.sh
```diff
@@ -64,7 +64,7 @@ function import_credentials() {
 # User overridable variables.
```
marcoderama: While we're at it, does it make sense to move this to `enterprise-suite/scripts`?

dbrinegar: yeah good question.  thoughts @jsravn ?

jsravn: Customers will be using this shell script, that's why I left it here. But we could put it in `scripts`, it's not a big deal I think.

jsravn: Did you end up moving it @dbrinegar?

marcoderama: He did not.  I could move it as part of https://github.com/lightbend/helm-charts/pull/100...

marcoderama: Created https://github.com/lightbend/helm-charts/pull/105 for moving the `install-es.sh` script.

PR #98, opened 2018-09-18
Update README.md

jsravn:


jsravn:
@marcoderama I suggest keeping this in the interest of greasing the wheels. I don't think it's ideal having a pointer to an internal doc in a public repo, either. We can remove it at some point.

PR #99, opened 2018-09-18
Allow for customization of baseurl

marcoderama:


marcoderama:
See more info at 
https://github.com/lightbend/es-backend/issues/273#issuecomment-421514838

marcoderama:
A test for this wouldn't fit well with the existing smoke tests.  Those run against a single instance of enterprise-suite.  To test this change, we'd have to shut down ES and reinstall it with `baseUrl` redefined.  In any case, to test this meaningfully, we should really be chasing down deeper links in the UI.  So...

I've suggested to @katsutoxin that he test this feature in the front-end tests he's wiring in now.  That works for him.  Thoughts @jsravn?

jsravn:
> A test for this wouldn't fit well with the existing smoke tests. Those run against a single instance of enterprise-suite. To test this change, we'd have to shut down ES and reinstall it with baseUrl redefined. In any case, to test this meaningfully, we should really be chasing down deeper links in the UI. So...

Why can't you just set the baseUrl for our tests? That way we exercise it.

And it affects all of our smoke tests - since we're relying on the es-console nginx proxy.

jsravn:
Nothing is being limited @dbrinegar. There is nothing stopping a user from doing `curl localhost:proxyport/baseURL/services/grafana`. Why would you want to support a dynamic path prefix? What use case is there for the user specifying any arbitrary path they want? I'm not even sure this is possible with nginx alone - prometheus and alertmanager both require knowing the baseURLs at runtime. I think grafana as well needs this for some stuff.

jsravn:
I think "support a dynamic baseURL/path prefix" is out of scope here anyways. All we're supporting now is a static, optional path customers can specify as part of the install - this supports ingress out of the box without a lot of work. If we want to also support the user typing in any arbitrary path to access the console, that needs its own story and prioritization.

jsravn:
I think you could do it by just modifying all the smoke tests to use the baseURL though. Wdyt @marcoderama.

marcoderama:
@jsravn Went with your other suggestion of reinstalling the Console, setting the baseUrl.  Works like a charm.

marcoderama:
And now that that's working, added tests for access to grafana/prom/es-monitor via ingress path.

The only problem is that the `smoke_ingress` test works in isolation but not when run at the same time as other smoke tests.  Gonna have to tweak `smoketest` to fix that.  sigh...

So anyone be merging this just yet.

marcoderama:
Tweaked `smoketest` to run `smoke_ingress` on its own.  Works fine locally.  Dies in Travis.  Will debug tomorrow...

c.f. https://travis-ci.org/lightbend/helm-charts/jobs/433305835 with some debug output around the failing ingress tests.  The three 'deeper' URLs just seem to timeout...

marcoderama:
Closing given @dbrinegar's work in https://github.com/lightbend/helm-charts/pull/114

PR #100, opened 2018-09-19
Use appropriate helm settings for install-dev target when in enterpri…

marcoderama:
…se-suite-latest

### DIFF
--- a/enterprise-suite-latest/Makefile
+++ b/enterprise-suite-latest/Makefile
@@ -23,7 +23,8 @@ include ../$(ES_CHART)/Makefile
 CHART_DIR = ../$(ES_CHART)
 COMPONENTS := $(wildcard ../$(ES_CHART)/*/.)
 SUBCOMPONENTS := $(wildcard ../$(ES_CHART)/*/*/.)
-
+# Override dev helm settings using values-latest.yaml file
+DEV_HELM_VALUES_FILE = values-latest.yaml
 
 # Special rule to build 'latest' chart
 $(HELM_CHARTS_DIR)/docs/$(filter %-latest,$(CHART))-$(VERSION).tgz: $(COMPONENTS) $(SUBCOMPONENTS)
--- a/enterprise-suite/Makefile
+++ b/enterprise-suite/Makefile
@@ -111,9 +111,13 @@ install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm  ## Install l
 	-ES_LOCAL_CHART=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz ES_FORCE_INSTALL=true \
 	    $(CHART_DIR)/scripts/install-es.sh --wait --set minikube=$(MINIKUBE),podUID=10001
 
+# helm settings for installing from local files
+# If set, DEV_HELM_VALUES_FILE will be added as '--values' value to install command, overriding default values.
+DEV_HELM_VALUES_FLAG = $(if $(DEV_HELM_VALUES_FILE), --values $(DEV_HELM_VALUES_FILE))
+
 install-dev: install-helm ## Install content of enterprise-suite/ folder
 	ES_LOCAL_CHART=$(CHART_DIR) ES_FORCE_INSTALL=true \
-	    $(CHART_DIR)/scripts/install-es.sh --wait --set minikube=$(MINIKUBE)
+	    $(CHART_DIR)/scripts/install-es.sh --wait --set minikube=$(MINIKUBE) $(DEV_HELM_VALUES_FLAG)
 
 help:  ## Print help for targets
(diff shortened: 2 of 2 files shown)


jsravn:
@marcoderama Careful w/ the PR summary, it overran into the comment box.

enterprise-suite-latest/Makefile
```diff
@@ -23,6 +23,13 @@ include ../$(ES_CHART)/Makefile
 CHART_DIR = ../$(ES_CHART)
 COMPONENTS := $(wildcard ../$(ES_CHART)/*/.)
 SUBCOMPONENTS := $(wildcard ../$(ES_CHART)/*/*/.)
+# Generate dev helm settings from enterprise-suite settings and values-latest.yaml file
+DEV_HELM_SETTINGS := $(DEV_HELM_SETTINGS)$(shell awk '\
+    BEGIN { FS = ": " } \
+    { \
+        settings = settings "," $$1 "=" $$2 \
+    } \
+    END { print settings }' values-latest.yaml)
```
jsravn: Wouldn't it be better to use `yq` vs `awk` hacking?

marcoderama: `awk` is my hammer!  What is this new-fangled `yq` that you speak of?

I'll take a look...

srikanthops: If it can be solved by yq, we should use, it, we are adding dependency on one more tool

marcoderama: I honestly think `awk` is fine for this job.  From what I can tell `yq` is a tool for converting between yaml and JSON.  That's not what I'm doing here.  I'm no `yq` whiz so came up with this:

`yq r -j values-latest.yaml | sed -e 's/[}{]//g' -e 's/"//g' -e 's/:/=/g'`

Is that preferable to the `awk` sol'n?  (Maybe it is.)  Is there a better way to do this with `yq`?

marcoderama: With respect to dependencies, I'd say that we can pretty much guarantee that `awk` is already installed.  We have to install `yq` in `scripts/setup-tools-for-ubuntu.sh`.  (And that's just one flavor of `yq`, others of which are incompatible.)

jsravn: We already rely on `yq` for creating the enterprise-suite-latest tarball. So installing it isn't an issue. The main advantage of `yq` is it will work regardless of how the yaml file is formatted - you're making assumptions with awk that the file is formatted a certain way. It's also difficult to understand compared to the `yq` syntax. `yq` is the right tool here, it's designed for reading and manipulating yaml files.

enterprise-suite-latest/Makefile
```diff
@@ -23,6 +23,8 @@ include ../$(ES_CHART)/Makefile
 CHART_DIR = ../$(ES_CHART)
 COMPONENTS := $(wildcard ../$(ES_CHART)/*/.)
 SUBCOMPONENTS := $(wildcard ../$(ES_CHART)/*/*/.)
+# Generate dev helm settings from enterprise-suite settings and values-latest.yaml file
+DEV_HELM_SETTINGS := $(DEV_HELM_SETTINGS),$(shell yq r -j values-latest.yaml | sed -e 's/[}{"]//g' -e 's/:/=/g')
```
jsravn: Why not just use `--set-file values-latest.yaml` ?

enterprise-suite-latest/Makefile
jsravn: Okay `yq` isn't as great as `jq` unfortunately... so it makes it a little harder to use `yq` here, admittedly. I wonder if you can use `--set-file`, which I think takes a `yaml` file directly.

enterprise-suite-latest/Makefile
marcoderama: Uhhh...  Cuz it never occurred to me for some reason...  Durn.  That's a much better solution.

srikanthops: I use it for minishift 
helm install --name=es-local-latest -f enterprise-suite-latest/values-latest.yaml  --debug --set minikube=true  enterprise-suite

marcoderama: Went with `--values` (same as `-f`).  Interestingly, I find that the `--set-file` flag is gone from `helm install` as of v 2.9.1 at least.  (Unfortunate that it's still documented at https://docs.helm.sh/helm/#helm-install.  But when I read the doco carefully, I see it wasn't right for the job anyway.  I've misunderstood that flag since forever...)

marcoderama: Aside:  Upgraded to helm 2.10.0 and the `--set-file` flag is there now.  Still not the right thing for this job but the mystery is solved.  :-)

marcoderama:
I did the squashing myself so I could control the commit comment.  Merging...

PR #101, opened 2018-09-19
Several improvements to the install script

jsravn:
- Use a more descriptive helm release name with less chance of
conflicts.
- Detect if chart is already installed, and perform an upgrade instead.
- Print out the helm commands that get run.
- Warn if version is unspecified.
- Add ES_FORCE_INSTALL which will delete the existing chart first
instead of upgrading.

I also removed part of NOTES.txt which is out of date, and replaced
it with a link to the documentation.

### DIFF
--- a/enterprise-suite/Makefile
+++ b/enterprise-suite/Makefile
@@ -92,17 +92,15 @@ install-helm:  ## Install required helm components into cluster
 	-kubectl create clusterrolebinding $(TILLER_NAMESPACE):tiller --clusterrole=cluster-admin --serviceaccount=$(TILLER_NAMESPACE):tiller
 	-helm init --wait --service-account tiller --tiller-namespace=$(TILLER_NAMESPACE)
 
-delete-local:  ## Delete chart from cluster with helm
-	-helm delete --purge $(RELEASE_NAME)
-
 # called from shell script for travis job:
-install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm delete-local  ## Install local tarball
+install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm  ## Install local tarball
 	# We set a podUID here for test purposes to ensure everything works as non-root.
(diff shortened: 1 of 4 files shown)


enterprise-suite/tests/install-es.bats
```diff
@@ -12,6 +12,8 @@ function setup {
     unset LIGHTBEND_COMMERCIAL_CREDENTIALS
     export LIGHTBEND_COMMERCIAL_USERNAME="myuser"
     export LIGHTBEND_COMMERCIAL_PASSWORD="mypass"
+    export ES_STUB_CHART_STATUS="1"
+    export ES_HELM_NAME="myhelmname"
```
jsravn: Just to test this works, as it currently shares the same name as the chart.

enterprise-suite/install-es.sh
```diff
@@ -25,7 +25,8 @@ function usage() {
     docvar ES_CHART "Chart name to install from the repository"
```
marcoderama: Is there a reason for controlling things with env vars as opposed to using command line flags?  Just curious.

enterprise-suite/Makefile
```diff
@@ -92,17 +92,15 @@ install-helm:  ## Install required helm components into cluster
 	-kubectl create clusterrolebinding $(TILLER_NAMESPACE):tiller --clusterrole=cluster-admin --serviceaccount=$(TILLER_NAMESPACE):tiller
 	-helm init --wait --service-account tiller --tiller-namespace=$(TILLER_NAMESPACE)
 
-delete-local:  ## Delete chart from cluster with helm
```
marcoderama: I know the makefile doesn't use it anymore but I'll probably still miss it...

enterprise-suite/install-es.sh
jsravn: 1. Parsing flags in bash is horrible and will add a lot of complexity to the script
2. I want to just pass through the arguments to helm so we don't need to duplicate helm - we can't do that easily if we use command line flags.

PR #103, opened 2018-09-19
Add serviceAccountName for node-exporter deployment

srikanthops:
to run in minishift node-exporter needs hostaccess. 

we can execute the below command to give hostaccess 
to serviceAccount "prometheus-kube-state-metrics" 

_oc adm policy add-scc-to-user -z prometheus-kube-state-metrics  hostaccess_

[Openshift node-exporter Reference implementation](https://github.com/openshift/origin/blob/master/examples/prometheus/node-exporter.yaml)

related to lightbend/es-backend/issues/320


enterprise-suite/templates/node-exporter.yaml
```diff
@@ -20,7 +18,7 @@ spec:
         app: prometheus
         component: node-exporter
     spec:
-      serviceAccountName: "default"
+      serviceAccountName: "prometheus-kube-state-metrics"
```
marcoderama: I really don't fully understand what's going on here, but should this be helm template variable?  (I mean, is it okay that this serviceAccountName be enshrined for everyone in the `.yaml`?

jsravn: Don't we need to create this serviceAccount in the first place @srikanthops ?

srikanthops: @jsravn  
We already have this serviceAccount, it is attached to "kube-state-metrics" 
I was piggybacking on the same serviceAccount.

We can surely create a new serviceAccount, let me do that to be on the safe side 

@marcoderama 
As mentioned above I am going to create a separate ServiceAccount and attach it to node-exporter-Deployment. 
If an empty default securityAssociation works fine with minikube, we don't need a helm template variable

srikanthops:
Steps I did 
>> create a new PodSecurityPolicy privileged
>> created a ClusterRole and ClusterRoleBinding that would allow privleged policy to ServiceAccount "prometheus-node-exporter"
>> this didn't work daemonSet prometheus-node-exporter kept on complaining it doesn't have enough permissions.

srikanthops:
Using "oc adm" commands in minishift/openshift environments is very common.
I will open a separate ticket to track this.
Can we close this as, I want to wrap up documenting minishift in es-docs

jsravn:
@srikanthops but it's an inconsistent experience. We create RoleBindings for other things, like prometheus. Why would we require them to use `oc` for just the node-exporter?

jsravn:
Okay, let's merge this for now. I'll see if I can get to the bottom of it.

PR #104, opened 2018-09-19
migrate frontend e2e test to helm-charts repo

katsutoxin:
connect to lightbend/console-home#343


enterprise-suite/Makefile
```diff
@@ -37,8 +37,21 @@ shell-tests:
...
 
+backend-tests: shell-tests minikube-tests
+
+frontend-tests:
+	$(CHART_DIR)/scripts/run-e2e-tests.sh
+
 # This target required by top-level helm-charts Makefile.  Tests release on Travis
-test: shell-tests minikube-tests
+#
+# The conditional bit is so we can control which set of tests are run given we're only
```
marcoderama: I suggested these Makefile (and .travis.yml) changes to Danny.  Not terribly pretty but allows us to define a `test` target for the helm-charts `Makefile` while allowing us to run the backend and frontend tests separately.

enterprise-suite/tests/e2e/cypress/config/README.md
```diff
@@ -0,0 +1,11 @@
+## Code flow for env specific config file
+
+1. in package.json, pass confie file setting to cypress, such as ```cypress run --env configFile=minikube```
```
marcoderama: confie -> config

enterprise-suite/tests/e2e/cypress/config/minikube.json.orig
```diff
@@ -0,0 +1,3 @@
+{
```
marcoderama: Did you mean to include this file?  Looks like a temp file for dev work.

enterprise-suite/tests/e2e/cypress/integration/create-monitor.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+// TODO: need to validate convert-to-rate in the future
+
+describe('Create Monitor Test', () => {
+
+  beforeEach(() => {
+    // Util.deleteMonitorsForWorkload('es-demo');
+  });
+
+
+  // NOTE: group by actor is flacky due to the following two issues
```
marcoderama: flacky -> flaky

enterprise-suite/tests/e2e/package.json.old
```diff
@@ -0,0 +1,34 @@
+{
```
marcoderama: Did you mean to include this file?

enterprise-suite/tests/e2e/scripts/check_e2e_results.sh
```diff
@@ -0,0 +1,40 @@
...
+echo "PASS=$PASS"
+echo "FAIL=$FAIL"
+echo "PENDING=$PENDING"
+echo "SKIP=$SKIP"
+echo ""
+
+
+if [[ "$PASS" -lt "$EXPECTED_PASS" ]]
+then
+  echo "pass count $PASS should not less then $EXPECTED_PASS"
```
marcoderama: ...should not be less than...

enterprise-suite/tests/e2e/scripts/check_e2e_results.sh
```diff
@@ -0,0 +1,40 @@
...
+  exit -1
+fi
+
+if [[ "$FAIL" -gt "$EXPECTED_FAIL" ]]
+then
+  echo "fail count $FAIL should not greater then $EXPECTED_FAIL"
```
marcoderama: ...should not be greater than...

enterprise-suite/tests/e2e/scripts/check_e2e_results.sh
katsutoxin: oh~~  this script is not really use so far.    (we might use it in the future, but not right now)
The script was designed for handling flaky test or known errors....  Now we decide to turn off all  flaky tests.

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: go
 sudo: required
 dist: xenial
+node_js: 9.4.0
```
marcoderama: Not sure you need it at this level.

.travis.yml
```diff
@@ -22,6 +23,7 @@ jobs:
     - scripts/setup-minikube-for-linux.sh
     - make test CHARTS=enterprise-suite-latest
   - name: Minikube - Frontend
+    node_js: 9.4.0
```
marcoderama: Wondering if that should be
`node_js: "9.4.0"`
based on the doco I see.  Not sure it makes a difference though.

.travis.yml
```diff
@@ -21,6 +21,15 @@ jobs:
     script:
     - scripts/setup-minikube-for-linux.sh
     - make test CHARTS=enterprise-suite-latest
+  - name: Minikube - Frontend
+    cache:
+      directories:
+        - ~/.npm
+        - ~/.cache
```
jsravn: I think the cache is unnecessary, since it only helps if you are building in stages.

.travis.yml
```diff
@@ -21,6 +21,15 @@ jobs:
...
+    node_js: 9.4.0
+    script:
+    - scripts/setup-minikube-for-linux.sh
+    - make test CHARTS=enterprise-suite-latest ES_TEST=frontend
```
jsravn: We need to test the stable version (not `-latest`) as well, so we have confidence in our releases.

.travis.yml
katsutoxin: The cache setting is recommend by official cypress-travis document https://docs.cypress.io/guides/guides/continuous-integration.html#Travis
It is to cache both npm modules and Cypress binary

 NOTE:  Cypress is downloaded to `~/.cache/Cypress` in linux
https://docs.cypress.io/guides/getting-started/installing-cypress.html#Overriding-the-Binary-Cache-Folder

.travis.yml
katsutoxin: just added :)

katsutoxin: Hey @jsravn ,  actually we only run the test for latest image is on purpose.  @dbrinegar , @marcoderama and I had a meeting to discuss which image we should run e2e test. We think that there are some feature gap between stable and latest image, so it cannot share the same e2e test because some tests only works for new feature. The right way should be generate/tag the release image from the latest image which pass the test. 

Take https://travis-ci.org/lightbend/helm-charts/builds/431063603 for example,  we found that #333.4 (stable image) is failed. but #333.5 (latest image) is passed.  The reason is that the test is design for latest image.

katsutoxin: oh~~  I checked it with @dbrinegar ,  he said "if your tests don't work against any released version of a thing, we need a release before it can be merged".  I am going to create a release for es-console.

jsravn: Yup you should depend on a released es-console in helm - the point of which is to tie together the versions. `latest` is just a weak signal of problems, and shouldn't block releases/PRs.

jsravn: Let's make this step only run in the nightly cronjob. Add:

```
if: type in (cron)
```

PR #106, opened 2018-09-24
Only build latest on the daily cron job

jsravn:
So we don't:
* Block PRs if the latest image happens to change and break things.
* Break the master branch release if latest image happnes to change and
break things.

We will run latest as part of the daily cron job, and get alerted if it
breaks.

Also turn down slack notifications to only alert if master build fails
or goes from failure to success.


.travis.yml
```diff
@@ -21,6 +21,8 @@ jobs:
     script:
     - scripts/setup-minikube-for-linux.sh
     - make test CHARTS=enterprise-suite-latest
+    # Only build latest in the daily cron - don't block PRs or master releases.
+    if: type in (cron)
```
marcoderama: Can this go higher in the definition?  I thought it applied to the Reactive-sandbox test at first blush.  (Not worth holding anything up for this though.  Ignore as you see fit.)

PR #107, opened 2018-09-24
akka_http monitor fixes

jsravn:
- Use a rate for 5xx monitor so we don't go permanently unhealthy.
- Remove duplicate graphs in the play-exporter.json
- Specify percentile and use ms for the akka_http response times.

For https://github.com/lightbend/es-backend/issues/345

### DIFF
--- a/enterprise-suite/es-monitor-api/bootstrap-monitors.json
+++ b/enterprise-suite/es-monitor-api/bootstrap-monitors.json
@@ -608,11 +608,11 @@
       "model": "threshold",
       "parameters": {
         "name": "akka_http_server_5xx",
-        "metric": "akka_http_http_server_responses_5xx",
+        "metric": "akka_http_http_server_responses_5xx_rate",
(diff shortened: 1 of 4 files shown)


enterprise-suite/es-monitor-api/bootstrap-monitors.json
```diff
@@ -608,11 +608,11 @@
       "model": "threshold",
       "parameters": {
         "name": "akka_http_server_5xx",
-        "metric": "akka_http_http_server_responses_5xx",
+        "metric": "akka_http_http_server_responses_5xx_rate",
         "comparator": ">",
         "threshold": "0",
         "window": "5m",
-        "confidence": "1",
+        "confidence": "0.25",
```
dbrinegar: this makes it slow to resolve, so those annoying errors at the beginning will linger for 5m.  setting to "1" makes it go green as soon as the 5xx errors stop

PR #111, opened 2018-09-25
Move run-minikube-test.sh stuff into Makefile

marcoderama:
It just seemed odd to me that the Makefile called a script that then invoked make targets.  I think it's cleaner without the script.


enterprise-suite/Makefile
```diff
@@ -34,8 +34,10 @@ package: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz
 shell-tests:
 	$(CHART_DIR)/tests/lib/bats/bats $(CHART_DIR)/tests
 
-minikube-tests:
-	$(CHART_DIR)/scripts/run-minikube-tests.sh
+minikube-tests: build install-local
+	@echo Installed Console from helm charts: $(CHART)
+	@echo Running tests
+	( cd $(CHART_DIR)/tests ; ./smoketest ; )
```
jsravn: Why not just run `tests/smoketest`?

marcoderama: cd'ed into the `tests` directory only cuz that's what the script did.  As it turns out, it's required atm.  We can fix that later...  :-)

PR #112, opened 2018-09-26
Use commercial repo for es-console

jsravn:
part of https://github.com/lightbend/es-backend/issues/110

### DIFF
--- a/enterprise-suite/templates/es-console.yaml
+++ b/enterprise-suite/templates/es-console.yaml
@@ -16,7 +16,7 @@ spec:
       {{ end }}
       containers:
       - name: es-console
-        image: lightbend-docker-registry.bintray.io/enterprise-suite/es-console:{{ .Values.esConsoleVersion }}
+        image: l{{ .Values.imageCredentials.registry }}/enterprise-suite/es-console:{{ .Values.esConsoleVersion }}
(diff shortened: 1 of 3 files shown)


enterprise-suite/templates/es-console.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       {{ end }}
       containers:
       - name: es-console
-        image: lightbend-docker-registry.bintray.io/enterprise-suite/es-console:{{ .Values.esConsoleVersion }}
+        image: l{{ .Values.imageCredentials.registry }}/enterprise-suite/es-console:{{ .Values.esConsoleVersion }}
```
marcoderama: That leading 'l' character looks wrong...
`l{{ .Values.imageCredentials.registry }}...`

enterprise-suite/values.yaml
```diff
@@ -2,7 +2,7 @@
 # Update these versions for a release.
 #
 # lightbend-docker-registry.bintray.io/enterprise-suite/es-console
```
marcoderama: We should change these comments now too where appropriate.  (Add the `commercial` bit.)

enterprise-suite/templates/es-console.yaml
jsravn: Yeah whoops!!

PR #113, opened 2018-09-26
WIP: Proxy to alertmanager, use envsubst for service IPs

jsravn:
part of https://github.com/lightbend/es-backend/issues/102

### DIFF
--- a/enterprise-suite/templates/es-console.yaml
+++ b/enterprise-suite/templates/es-console.yaml
@@ -14,6 +14,25 @@ spec:
       securityContext:
         runAsUser: {{ .Values.podUID }}
       {{ end }}
+
+      initContainers:
+      - name: nginx-conf-envsubst
+        image: alpine
+        resources:
+          requests:
+            cpu: {{ default .Values.defaultCPURequest }}
+            memory: {{ default .Values.defaultMemoryRequest }}
+        command:
+        - /bin/sh
+        - -c
+        args:
+        - apk add gettext libintl && cat /etc/es-console/default.conf.tmpl | envsubst > /etc/nginx/conf.d/default.conf
+        volumeMounts:
+        - name: config-volume
+          mountPath: /etc/es-console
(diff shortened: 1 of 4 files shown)


jsravn:
I'm not sure why, but this change causes nginx to start redirecting to port 8080. I'm sort of baffled :thinking: .

enterprise-suite/templates/es-grafana.yaml
```diff
@@ -50,7 +50,7 @@ metadata:
   labels:
     app: grafana
     component: "server"
-  name: grafana-server
+  name: grafana
```
marcoderama: Do we need to let people know about this change (and the prometheus one)?

enterprise-suite/templates/es-grafana.yaml
```diff
@@ -84,7 +84,7 @@ spec:
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
-  name: grafana-server
+  name: grafana
```
marcoderama: Does line 106 need to change too?  
    `name: grafana-server`
What about the comment at line 11?
    `# 2. add prometheus data source to grafana-server.`

enterprise-suite/templates/es-console.yaml
```diff
@@ -14,6 +14,25 @@ spec:
...
+        image: alpine
+        resources:
+          requests:
+            cpu: {{ default .Values.defaultCPURequest }}
+            memory: {{ default .Values.defaultMemoryRequest }}
+        command:
+        - /bin/sh
+        - -c
+        args:
+        - apk add gettext libintl && cat /etc/es-console/default.conf.tmpl | envsubst > /etc/nginx/conf.d/default.conf
```
marcoderama: Just curious.  How is the environment initially setup so that `envsubst` has something to work with?

jsravn:
Going to close this since its redundant with #114

PR #114, opened 2018-09-27
resilient nginx

dbrinegar:
1. service name lookups are not critical errors, but rather just cause
nginx to serve 502's until the service name is available, at which point
nginx will switch over to it automatically.  This fixes the issue of
nginx crashing if prometheus has not yet started, as well as fixes the
issue of nginx failing if prometheus is moved.

2. base paths are dynamic.  However you arrive at nginx, it preserves
the base path for the Console SPA, Grafana, Prometheus, and
es-monitor-api.  So one user can be using a port-forward while another
is using an ingress while another is using kube proxy, and it'll all
just work.

3. busts browser caching for the Console SPA entry point (index.html) as
well as our Grafana dashboard plugin (exporter-async.js). This should
stop the problem of browsers caching older versions and not pickup up
updates.

4. routes Grafana session cookies properly, so session store doesn't get
stuffed with new sessions for every request

5. all services are directly reachable, so /service/prometheus in the
browser will open prometheus and it'll just work

fixes lightbend/es-backend#342
part of lightbend/es-backend#311 (needs docs)
fixes lightbend/es-backend#353


dbrinegar:
cc @katsutoxin @pcraddick

dbrinegar:
@marcoderama Travis test is failing with 

```
Waiting for smoke_ingress to finish... error: error executing template "{{ (index (index .spec.template.spec.containers 0).ports 0).containerPort }}": template: output:1:4: executing "output" at <index (index .spec.t...>: error calling index: index of untyped nil
Error: unknown flag: --verbose
```

look familiar?

enterprise-suite/templates/es-console.yaml
```diff
@@ -15,6 +15,13 @@ spec:
         runAsUser: {{ .Values.podUID }}
       {{ end }}
       containers:
+      - name: dnsmasq
+        image: "janeczku/go-dnsmasq:release-1.0.7"
```
jsravn: Is there a reason to not use dnsmasq? This golang version looks relatively new and not widely used. It's also unmaintained - the last commit was 2 years ago. I would trust the native dnsmasq a lot more.

jsravn: Also thinking in terms of: security issues, redhat certification. Using official dnsmasq will better meet these needs.

enterprise-suite/templates/es-console.yaml
```diff
@@ -72,22 +79,100 @@ data:
...
-        root  /usr/share/nginx/html;
-        index index.html index.htm;
-        try_files $uri$args $uri$args/ /index.html;
+      # use external resolver to lookup backends, cache for 30 seconds
+
+      resolver 127.0.0.1:5353 ipv6=off valid=30s;
+
+      set $prometheus "prometheus-server.{{.Release.Namespace}}.svc.cluster.local.";
+      set $monitorapi "es-monitor-api.{{.Release.Namespace}}.svc.cluster.local.";
+      set $grafana "grafana-server.{{.Release.Namespace}}.svc.cluster.local.";
```
jsravn: This isn't going to work outside of minikube. The fqdn here will change based on the cluster setup. We need to rely on the search domains that k8s installs.

enterprise-suite/templates/es-console.yaml
```diff
@@ -72,22 +79,100 @@ data:
...
+
+      # proxy_pass backends with dynamic base url
+      # (redirects to ' $1/' with a space prefix tells nginx to not add server and port)
+
+      sub_filter_once off;
+
+      location ~ (.*/service/prometheus)(/.*) {
+        proxy_pass http://$prometheus$2$is_args$args;
```
jsravn: Does this style work with `upstream` directives in nginx? Just thinking if we ever wanted to support keepalives or more than one backend, or the other config in `upstream`. I guess we can cross that bridge when we come to it.

jsravn:
What do you think about adding a lot more commenting in nginx.conf, for people not super familiar with all the nginx rewriting syntax?

enterprise-suite/templates/es-console.yaml
```diff
@@ -72,22 +79,100 @@ data:
...
+        proxy_redirect '/' ' $1/';
+      }
+
+      location ~ (.*/service/es-monitor-api)(/.*) {
+        proxy_pass http://$monitorapi$2$is_args$args;
+        proxy_redirect '/' ' $1/';
+      }
+
+      location ~ (.*/service/grafana)(/.*) {
+        proxy_pass http://$grafana:3000$2$is_args$args;
```
jsravn: It'd be nicer if we could use service discovery for the port, rather than hardcoding it. Not a super big deal though. It's available in the dns srv record and env vars, but I'm guessing nginx can't easily use either.

enterprise-suite/templates/es-console.yaml
dbrinegar: Good points. This one was instead of me writing an init container to parse /etc/resolv.conf, but maybe we should think about that option. Needing it is totally silly because nginx requires an external resolver in the mode where it does lookups at run time, but what we want is just use the built in resolver, no special resolver needed.

I think the go-dnsmasq repo could also be inactive because it is "done" meaning there's nothing else needed, it is truly simple. But yeah, not a great look since at least there could have been a go runtime update along the way.

In other words, all this does is provide a go runtime that does a dns lookup, as a dns service.

I was tempted to use "kube-dns.kube-system" as the "resolver" as an example, but steered off of it as there seems a bunch of work in this area with non-kube-dns alternatives emerging.

One alternative is to have an init container that parses the resolver out of /etc/resolv.conf and gives that to nginx in a config file.

enterprise-suite/templates/es-console.yaml
dbrinegar: omg.

Oh!  At this means the "resolver" needs to do search, so need to test nginx as I don't think it does this in external resolver mode. gah

enterprise-suite/templates/es-console.yaml
dbrinegar: The proxy_pass feature has different modes, the lookup with resolver mode being triggered by the host being a variable. If we wanted it to go to an upstream, we'd use the upstream key here instead of `$prometheus` but leave the rest as-is.

```
     upstream prompool {
       ...
     }

      location ~ (.*/service/prometheus)(/.*) {
        proxy_pass http://prompool$2$is_args$args;
        ...
```

For upstreams, this lookup mode is triggered by the http block having a resolver defined. 🙄

enterprise-suite/templates/es-console.yaml
dbrinegar: I have no idea why we defined a service on a funky port instead of port 80. It might be a Grafana-ism, which wants to write fully qualified things in places which might include the port.

Or it might be someone copy and pasted random things from the internet.... :-|

dbrinegar: @katsutoxin @srikanthops do either of you know why we picked port 3000 for Grafana, instead of port 80 like everything else?

enterprise-suite/templates/es-console.yaml
dbrinegar: @jsravn what do you say to the idea of going forward with this now, make a follow on ticket to noodle out how to improve the external resolver situation?  The point being that this fixes a bunch of problems we're having pretty regularly, and the resolver thing is quite fiddly / needs some care.

enterprise-suite/templates/es-console.yaml
dbrinegar: fixed, but .. bummer

dbrinegar:
ok @jsravn addressed comments except for the resolver.  would like to carve that out separately so we can benefit from these changes in tomorrow's release

dbrinegar:
@marcoderama need your help with the busted tests here.  do I need to merge in a fix from somewhere?

enterprise-suite/templates/es-console.yaml
```diff
@@ -72,22 +79,111 @@ data:
...
+      # bookmarks / reloads in the /workloads/ path
+      # (for our single page app, these should reload index.html but keep the path)
+
+      location ~ (.*/)(workloads/.*) {
+        alias $base/index.html;
+        default_type text/html;
+        sub_filter 'base href="/"' 'base href="$1"';
+        # kill cache
+        add_header Last-Modified $date_gmt;
+        add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
```
katsutoxin: For the line `add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0'` , we duplicated it in 3 different places.  Could we set a variable for it?

dbrinegar: ya nginx doesn't seem to have very good facilities for composing configuration, but we can certainly define a common string.  I couldn't find a good way to have the whole section common, so there is some copy paste

enterprise-suite/templates/es-console.yaml
katsutoxin: @dbrinegar  port 3000 is the default grafana port.  Please refer http://docs.grafana.org/installation/configuration/#http-port

enterprise-suite/templates/es-console.yaml
jsravn: Ah okay, so go-dnsmasq actually reads the search domains out of `resolv.conf`. I missed that. Yeah, I don't think there would be a good way to handle this with dnsmasq.

Like you said, you could just use the kube-dns resolver - but that could bypass a customer's custom setup, like if they wanted to have local dnsmasq caching things on the host, or one of the other options. `/etc/resolv.conf` is the handoff point for apps - so we would have to use that one way or another to find the correct DNS server to use and search domains. It's pretty irritating that nginx doesn't support /etc/resolv.conf out of the box.

jsravn:
Sorry @dbrinegar - I completely forgot to ask if you can provide a smoketest for this? It just needs to hit with some arbitrary path to check the various proxies are working. Everything else lgtm now.

jsravn:
Approved, pending test, so you don't have to wait for me until Monday.

PR #115, opened 2018-09-27
Proxy alertmanager, add example config, make it customizable

jsravn:
This adds additional support for configuring alertmanager:
- Helm parameter for selecting a user provided ConfigMap for the alertmanager config
- Greatly expanded the example to help users get started quicker
- Proxy alertmanager via the console nginx

### DIFF
--- a/enterprise-suite/Makefile
+++ b/enterprise-suite/Makefile
@@ -105,12 +105,12 @@ install-helm:  ## Install required helm components into cluster
 	-kubectl create namespace $(TILLER_NAMESPACE)
 	-kubectl create serviceaccount --namespace $(TILLER_NAMESPACE) tiller
 	-kubectl create clusterrolebinding $(TILLER_NAMESPACE):tiller --clusterrole=cluster-admin --serviceaccount=$(TILLER_NAMESPACE):tiller
-	-helm init --wait --service-account tiller --tiller-namespace=$(TILLER_NAMESPACE)
+	-helm init --wait --service-account tiller --upgrade --tiller-namespace=$(TILLER_NAMESPACE)
 
 # called from shell script for travis job:
 install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm  ## Install local tarball
 	# We set a podUID here for test purposes to ensure everything works as non-root.
-	-ES_LOCAL_CHART=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz ES_FORCE_INSTALL=true \
+	-ES_LOCAL_CHART=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz \
 	    $(CHART_DIR)/scripts/install-es.sh --wait --set minikube=$(MINIKUBE),podUID=10001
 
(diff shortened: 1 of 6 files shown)


enterprise-suite/Makefile
```diff
@@ -105,20 +105,20 @@ install-helm:  ## Install required helm components into cluster
 	-kubectl create namespace $(TILLER_NAMESPACE)
 	-kubectl create serviceaccount --namespace $(TILLER_NAMESPACE) tiller
 	-kubectl create clusterrolebinding $(TILLER_NAMESPACE):tiller --clusterrole=cluster-admin --serviceaccount=$(TILLER_NAMESPACE):tiller
-	-helm init --wait --service-account tiller --tiller-namespace=$(TILLER_NAMESPACE)
+	-helm init --wait --service-account tiller --upgrade --tiller-namespace=$(TILLER_NAMESPACE)
 
 # called from shell script for travis job:
 install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm  ## Install local tarball
 	# We set a podUID here for test purposes to ensure everything works as non-root.
-	-ES_LOCAL_CHART=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz ES_FORCE_INSTALL=true \
```
marcoderama: Interesting.  Why the change?  (Just curious.)

enterprise-suite/alertmanager/alertmanager.yml
```diff
@@ -0,0 +1,55 @@
+# Example Lightbend Console Alertmanager config.
+
+global:
+
+# Routing rules.
+route:
+  receiver: slack
+  # Resend unresolved alerts every 4h.
```
marcoderama: I prefer something more generic like "Resend interval for unresolved alerts".  That way if we ever change the value, we don't risk the doco getting out of date.  Just sayin'.

enterprise-suite/alertmanager/alertmanager.yml
```diff
@@ -0,0 +1,55 @@
...
+templates:
+# Slack notification templates.
+- '/etc/config/slack.tmpl'
+
+# Receivers.
+receivers:
+- name: slack
+  slack_configs:
+  # Use Slack integrations to find the webhook.
+  - api_url: https://slack.localhost/services/somewhere
```
marcoderama: Does this need to be templatized or is that an actual value?

enterprise-suite/Makefile
jsravn: I was just getting frustrated with the 60s+ redeploy / dev cycle :). Removing force install and using upgrade works fine 99% of the time and only takes seconds. Maybe it's controversial? Not sure. I mentioned this in slack and no one replied, so I assume it's okay...

enterprise-suite/alertmanager/alertmanager.yml
jsravn: This file isn't intended to be used as-is - see my docs change https://github.com/lightbend/es-docs/pull/55. It's an example that users should copy and edit, then load into their own configmap.

PR #116, opened 2018-09-28
Update make-release.sh to work with helm-charts changes.

marcoderama:
- Uses version in Chart.yaml and post-increments.
- Version numbers in `enterprise-suite` and `reactive-sandbox` incremented so the next build will have the right version number.  (No change for `sample-project` because it hasn't been released yet.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,17 +67,26 @@ Then run the release script on a clean master checkout:
     scripts/make-release.sh enterprise-suite
     git push --follow-tags
     
-This will increment the chart version, package it, and make a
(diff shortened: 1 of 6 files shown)


marcoderama:
Most definitely meant to make those version number changes.  (See 
https://github.com/lightbend/helm-charts/pull/116#issue-218820607)  With these changes, `Chart.yaml` needs to have the _next_ version number.  Without those `Chart.yaml` mods, we'd overwrite the existing tar balls.

marcoderama:
"Dismissing" a review seems so harsh...

PR #118, opened 2018-09-28
remove test for k8sIcon from workload e2e test

craigupson:


enterprise-suite/tests/e2e/cypress/integration/workload-page.spec.ts
```diff
@@ -17,7 +17,7 @@ describe('Workload Page Test', () => {
 
   it('validate control icons', () => {
     Util.validateControlIconContains('Grafana');
-    // Util.validateControlIconContains('Kubernetes');
+    //  Util.validateControlIconContains('Kubernetes');
```
jsravn: Doesn't look like anything changed?

craigupson: I mistakenly pushed a commit to master (removing this kubernetes icon test) then backed it out (via reset --hard 4ffccd1   <- my commit id) so this might be an artifact of that.

craigupson:
retest this please

enterprise-suite/tests/e2e/cypress/integration/workload-page.spec.ts
jsravn: Well master has this change already - see https://github.com/lightbend/helm-charts/blob/master/enterprise-suite/tests/e2e/cypress/integration/workload-page.spec.ts

Surprised you could push to master, thought we disabled that!

craigupson:
a screw up on my part - which I thought I had backed out.. sorry.

PR #120, opened 2018-09-29
Fix make-release.sh

marcoderama:
- Don't increment version too early
- Minor doco tweaks

### DIFF
--- a/enterprise-suite/README.md
+++ b/enterprise-suite/README.md
@@ -71,10 +71,6 @@ These assume you're using minikube.
 
 ### Release ES images
 
-See the Releases section of the
-[Enterprise Suite Build and Release Process](https://github.com/lightbend/helm-charts/tree/master/enterprise-suite)
-documentation.
-
 ### Release Charts
 
 Install [yq](https://github.com/mikefarah/yq) if you don't have it yet:
--- a/enterprise-suite/README.md
+++ b/enterprise-suite/README.md
@@ -69,17 +69,11 @@ These assume you're using minikube.
 
 ## Cutting a Release / Publishing Charts
 
-_Is this still correct?_
-
 ### Release ES images
 
-Release the images in Jenkins:
-1. Go to <https://ci.lightbend.com/view/EntSuite/job/es-release-all/build>.
-2. Click 'Build'. This will increment the version for each image and release it.
-3. Open the [console](https://ci.lightbend.com/view/EntSuite/job/es-release-all/lastBuild/console), scroll to the bottom and you can see the versions of each image.
-4. Update [enterprise-suite/values.yaml](enterprise-suite/values.yaml) with the new image versions and commit the changes.
-
-See [es-build](https://github.com/lightbend/es-build) for more details.
+See the Releases section of the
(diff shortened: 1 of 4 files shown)


enterprise-suite/README.md
```diff
@@ -89,17 +83,19 @@ Install [yq](https://github.com/mikefarah/yq) if you don't have it yet:
     # or
     brew install yq                  
 
+Make sure the image versions in `enterprise-suite/values.yaml` are as desired.
+
 Then run the release script on a clean master checkout:
 
-    scripts/make-release.sh enterprise-suite
+    ./scripts/make-release.sh enterprise-suite
```
jsravn: Why is `./` needed?

enterprise-suite/README.md
```diff
@@ -69,17 +69,11 @@ These assume you're using minikube.
...
-Release the images in Jenkins:
-1. Go to <https://ci.lightbend.com/view/EntSuite/job/es-release-all/build>.
-2. Click 'Build'. This will increment the version for each image and release it.
-3. Open the [console](https://ci.lightbend.com/view/EntSuite/job/es-release-all/lastBuild/console), scroll to the bottom and you can see the versions of each image.
-4. Update [enterprise-suite/values.yaml](enterprise-suite/values.yaml) with the new image versions and commit the changes.
-
-See [es-build](https://github.com/lightbend/es-build) for more details.
+See the Releases section of the
+[Enterprise Suite Build and Release Process](https://github.com/lightbend/helm-charts/tree/master/enterprise-suite)
+documentation.
```
jsravn: We should probably pull this, because it's in the public repo. I already had someone random try to access the doc and got an email about it.

enterprise-suite/README.md
```diff
@@ -89,17 +83,19 @@ Install [yq](https://github.com/mikefarah/yq) if you don't have it yet:
...
+
 Then run the release script on a clean master checkout:
 
-    scripts/make-release.sh enterprise-suite
+    ./scripts/make-release.sh enterprise-suite
     git push --follow-tags
     
 This will increment the chart version, package it, and make a
-commit. Finally, `git push --tags` will publish the release and git tag.
+commit. Finally, `git push --follow-tags` will publish the release and git tag.
```
jsravn: This isn't true anymore, it publishes before the version is incremented. Should make it clear here.

enterprise-suite/README.md
marcoderama: Not really needed.  I just put it there cuz a lot of people don't have '.' in their path.  I think I tried to cut'n'paste that line and it failed for me.

Leaving for now.  We can change later if need be.  (Cuz I think I made that tweak in other places too.)

enterprise-suite/README.md
marcoderama: Good point.  I fixed by tagging before making the change to the version in Chart.yaml.

enterprise-suite/README.md
jsravn: You don't need `.` in your PATH. `scripts/make-release.sh` should be valid in any `sh` based shell.

PR #122, opened 2018-09-30
relabel container and pod metrics with workload

dbrinegar:
Following the naming conventions used for Pod Template Hashes and the
SimpleNameGenerator, as well as the reduced SafeString character set, we
pattern match for workload names out of pods with these patterns:

1. workload-hash-simple
2. workload-simple
3. workload-number
4. workload

While not perfect, it seems to work quite well in the main. Is is
certainly possible to incorrectly match "hash" or "simple" with plain
names that have those patterns, but given the size constraints and the
fact that vowels are not allowed reduces the chance of error.

With this we also swapped in correct metrics for the Grafana Kubernetes
section, since we can match container stats with ContextTags.

fixes lightbend/es-backend#310


enterprise-suite/es-grafana/kubernetes-exporter.json
```diff
@@ -1,7 +1,32 @@
...
+            "sum without(container_name, id, image, name, interface)(irate(container_network_receive_bytes_total{ContextTags}[1m]))"
+        ]
+    },
+    {
+        "graphName": "Bytes Transmitted Per Second",
+        "promQL": [
+            "sum without (container_name,id, image, name, interface)(irate(container_network_transmit_bytes_total{ContextTags}[1m]))"
+        ]
+    }
 ]
```
jsravn: Will this work with the top-level (cluster view) grafana dashboard?

My idea is we should have two separate dashboards - one is the cluster dashboard, that the console home page links to, and the workspace scoped container dashboard (this one here).

jsravn: Ah right, that is in a separate exporter. Nevermind.

dbrinegar:
Couple other cleanup changes after the initial review

enterprise-suite/es-grafana/kubernetes-exporter.json
```diff
@@ -1,7 +1,38 @@
...
+    {
+        "graphName": "CPU Percent",
+        "promQL": [
+            "sum without(container_name,id,image,name)((irate(container_cpu_user_seconds_total{ContextTags}[1m]) + irate(container_cpu_system_seconds_total[1m])) * 100)"
+        ]
+    },
+    {
+        "graphName": "Active Memory Used Percent",
+        "promQL": [
+            "sum without(container_name,id,image,name)((container_memory_working_set_bytes{ContextTags} / container_spec_memory_limit_bytes < +Inf or container_memory_working_set_bytes{ContextTags} / on (node_ip) group_left node_memory_MemTotal) * 100)"
```
jsravn: I think RSS is far more useful than working_set_bytes. People want to know how close their pod is to being OOMKilled - RSS will tell you that. working_set_bytes - I'm not sure to be honest, it will be useful in some narrow contexts for perf analysis. With a JVM process I imagine working_set_bytes will be close to RSS always anyways.

Can we have RSS here instead? Or include both of them? RSS + file caches are the most important memory metrics on a Linux system for monitoring. It also answers the question "Why is my total usage larger than my RSS?" - because RSS + file_cache = total usage.

enterprise-suite/es-grafana/kubernetes-exporter.json
```diff
@@ -1,7 +1,38 @@
...
+        ]
+    },
+    {
+        "graphName": "Total Memory Used Percent",
```
jsravn: Should we mention here this is out of the limit or out of the node? Can we set a graph description perhaps?

enterprise-suite/es-grafana/kubernetes-exporter.json
```diff
@@ -1,7 +1,38 @@
...
+        "promQL": [
+            "sum without(container_name,id,image,name)((container_memory_usage_bytes{ContextTags} / container_spec_memory_limit_bytes < +Inf or container_memory_usage_bytes{ContextTags} / on (node_ip) group_left node_memory_MemTotal) * 100)"
+        ]
+    },
+    {
+        "graphName": "File System Used Percent",
+        "promQL": [
+            "sum without(container_name, id, image, name)(container_fs_usage_bytes{ContextTags} / container_fs_limit_bytes * 100)"
```
jsravn: Isn't this going to produce a metric for every filesystem on the node? I seem to remember having to filter this down somehow.

enterprise-suite/es-grafana/kubernetes-exporter.json
```diff
@@ -1,7 +1,38 @@
...
+        ]
+    },
+    {
+        "graphName": "Bytes Received Per Second",
```
jsravn: Can we make it KBytes so it's more readable? Or set the unit on the graph if possible here.

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -71,6 +71,22 @@ scrape_configs:
       - source_labels: [__name__]
         regex: container_(network_tcp_usage_total|tasks_state|cpu_usage_seconds_total|memory_failures_total|network_udp_usage_total)
         action: drop
+      # extract workload from workload-hash-rand
+      - source_labels: [kubernetes_pod_name, es_workload]
+        regex: (.*)-[456789bcdf]{4,10}-[bcdfghjklmnpqrstvwxz2456789]{5};
```
marcoderama: Just curious why no `123` in `[456789bcdf]`

dbrinegar: this is the SafeString set, full explanation in the issue.  The "hash" is from a string of digits, so there's only `[0-9]` which maps in SafeString to `[456789bcdf]`

enterprise-suite/es-grafana/kubernetes-exporter.json
dbrinegar: I'm happy to use RSS here, but just want to pause on the rationale.

Rationale: I added this based on your good suggestion to monitor for container pressure (when will I be oom'd). Now I haven't worked in kernel vm code for a long time, but from what little I do know, predicting when the kernel will be unable to fulfill a slab request is crazy hard. It is also my understanding that "working_set" is an attempt to do better than RSS for this very purpose, as it better captures the "pressure" aspect, meaning it measures what has the kernel not been able to put down, what is actively on wire, regardless of purpose. In this circumstance, slab allocations over the remainder will be at least non-trivial to handle and very likely to oom.

In any event, setting aside the theory, the kubelet itself is using working_set as the high water line for kube memory pressure / eviction triggers:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/eviction/memory_threshold_notifier.go#L80-L112

That's the same data point we're collecting, and the reason for using it.

Anwyay, the idea of using this is to more accurately reflect the OOM line and not potentially underestimate it by only considering RSS.

enterprise-suite/es-grafana/kubernetes-exporter.json
dbrinegar: There's nothing in the current config to let us do that, but I have a change coming that would let us set the legend with a template.  Not sure if it can do an expression or just label swaps..would have to dig a little.  We can / should certainly document our graphs.

enterprise-suite/es-grafana/kubernetes-exporter.json
dbrinegar: The cadvisor collection seems to be pretty clean and sticking to real devices used in the container.

node_exporter definitely has a bunch of cruft to sweep through but we do it this way

https://github.com/lightbend/helm-charts/blob/master/enterprise-suite/templates/node-exporter.yaml#L42-L43

enterprise-suite/es-grafana/kubernetes-exporter.json
dbrinegar: units coming, not in this PR but the next

dbrinegar:
@jsravn let's chat about working_set and make a call...

enterprise-suite/es-grafana/kubernetes-exporter.json
dbrinegar: more evidence, FYI:

> The better metric is container_memory_working_set_bytes as this is what the OOM killer is watching for.

https://blog.freshtracks.io/a-deep-dive-into-kubernetes-metrics-part-3-container-resource-metrics-361c5ee46e66

> The kubelet exludes inactive_file (i.e. # of bytes of file-backed memory on inactive LRU list) from its calculation as it assumes that memory is reclaimable under pressure.

https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/

> script reproduces what the kubelet does... mem_avail = capacity - working_set

https://github.com/kubernetes/website/pull/2892/files#diff-e8b9869d9d8022398a9f32d68250d792

dbrinegar: and here's a bunch of folks "wishing it wasn't so"

https://github.com/kubernetes/kubernetes/issues/43916

although I think they're perhaps up against other problems and getting confused.

Anyway, this is the line we should be watching, pretty sure!

jsravn: https://blog.freshtracks.io/a-deep-dive-into-kubernetes-metrics-part-3-container-resource-metrics-361c5ee46e66 is just wrong I think. The OOMKiller considers reclaimable memory. On a system without swap, the only reclaimable pages are file system cache - so RSS is the value to look at. I'm not even sure OOMKiller will consider inactive anon pages - it takes too long to swap to disk.

Working set bytes _might_ (big IF) be used if disk swap is available - the kernel can then swap out those unused pages under memory pressure. But no reasonable production system should have swap on. Especially with Kubernetes! In the vast majority of cases, working set bytes is going to be almost meaningless, unless it happens to be close to RSS.

jsravn: I fully understand the difference between reclaimable pages vs anon pages btw. So I agree with all the links you provide, except for that first blogpost. We shouldn't count filesystem pages towards memory pressure (with some exceptions, like databases which heavily rely on fs caches).

My issue is with this cadvisor metric working_memory_bytes - which has _nothing_ to do with the OOMKiller or actual physical memory (anon pages) used as far as I can tell. Instead, it looks like some kernel book keeping of recently used pages in the process. I've never seen anyone use this in actual monitoring, because as far as I can tell, it's not easily available in /proc either. I'm not actually sure where cadvisor is getting this number from. I can probably dig a bit more into the kernel code / cadvisor to see what exactly this value is used for. It might be useful in some limited contexts, but RSS and file caches are the important values.

jsravn: Okay, I did the digging.

You can see how cadvisor calculates this value (which is not from Linux, so has nothing to do with OOMKiller - it's cAdvisor's own creation): https://github.com/google/cadvisor/blob/master/container/libcontainer/handler.go#L530

It's calculated as:

     total_usage - total_inactive_file

`total_inactive_file` is the sum of `inactive_file` for the process. `inactive_file` is defined at https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt:

    inactive_file	- # of bytes of file-backed memory on inactive LRU list.

So this cadvisor metric working_set_bytes means: "Total memory usage, not including inactive file system cache". So it's actually:

    working_set_bytes = rss + total_active_file

So it's showing resident memory and all the recently used file system pages. That is actually a pretty nice metric - it's showing the "real" memory profile. For something like prometheus, this value will probably include almost all of the filesystem cache. For something that is JVM based (and not a db), it will probably exclude almost all of the filesystem cache.

So it's a useful metric. But it's a bit fuzzier than the other metrics, like RSS and file_cache. My conclusion is:

* RSS will show you the actual physical page usage that is not reclaimable. If RSS exceeds your memory limit, you get OOMKilled.
* file_cache shows all the reclaimable pages. The OS will reclaim all of these before the OOMKiller is triggered.
* cAdvisor's working_memory_bytes is a derived value. It gives an idea of the amount of "extra" memory needed beyond anon pages, to keep your process performing well. This would be very useful with any database that relies heavily on the file_cache.

I think RSS+file_cache are still the most important to monitor, working_memory_bytes will be useful as a performance metric (keep your total memory limit > working_set_bytes for best performance).

jsravn: So  I completely missed in David's comment that kubelet is actually using working_set_bytes as part of its eviction calculations. So if a user enables kubelet eviction policies (https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/#eviction-policy), they'll run into this.

dbrinegar:
summary after discussion:

1. Resident Set Size is nice for understanding App vs Cache because it adds up to Total. Is also used by OOM killer when it is picking things to knock down to save the kernel.

2. Working Set Size is nice for understanding memory pressure and risk of eviction, if the Kubernetes cluster has an eviction policy.

For Akka apps they're probably very close to each other. We'll go with #1 as is easier to reconcile, for example when comparing to Cache and Total, or when looking at memory budgets across JVM, Container, Kubelet, Host.

PR #124, opened 2018-10-02
Archive cypress videos in s3

jsravn:


jsravn:
Build isn't triggering.

jsravn:
Ah, https://travis-ci.org/lightbend/helm-charts/requests reveals all.

.travis.yml
```diff
@@ -5,25 +5,30 @@ dist: xenial
...
 
 jobs:
   include:
   - stage: Tests
     name: Validate
     script:
     - make lint
-  - name: Minikube
+
+  - name: ES Smoke Tests
```
marcoderama: Glad to see the name change.  Was thinking about that the other day...

.travis.yml
```diff
@@ -42,16 +51,19 @@ jobs:
...
     - make test CHARTS=reactive-sandbox
+
   - stage: deploy
     name: Deploy to GCS
     # Only deploy for pushes to master (and if manually invoked)
     if: type IN (push, api)
     script:
     # Decrypt credentials for GCS system account helmchart@es-repo.iam.gserviceaccount.com
-    - openssl aes-256-cbc -K $encrypted_f01ffbb90c44_key -iv $encrypted_f01ffbb90c44_iv -in resources/es-repo-7c1fefe17951.json.enc -out resources/es-repo-7c1fefe17951.json -d
+    - mkdir -p /tmp/resources
```
marcoderama: I was considering using `$TMPDIR` instead of `/tmp` for something else.  Probably doesn't matter but was curious what you thought.

.travis.yml
```diff
@@ -42,16 +51,19 @@ jobs:
...
+    - openssl aes-256-cbc -K $encrypted_f01ffbb90c44_key -iv $encrypted_f01ffbb90c44_iv -in resources/es-repo-7c1fefe17951.json.enc -out /tmp/resources/es-repo-7c1fefe17951.json -d
```
marcoderama: Curious about this change.  Is this just to keep source and generated artifacts separate?  Is there a need to put things in `/tmp` given that everything is going to get blown away anyway once the test runs to completion?  (I suspect I'm missing some subtlety...)

jsravn: It's so we don't accidentally upload the encrypted file to S3. By default the artifacts tool will upload anything generated inside the git repo. So this is just being "better safe than sorry".

.travis.yml
jsravn: Not too worried about portability here since this is travis specific, so we can rely on the build environment. (Famous last words?)

PR #127, opened 2018-10-04
Fix unset var check

JustinPihony:
See https://stackoverflow.com/a/13864829/779513 for what is expected

### DIFF
--- a/enterprise-suite/scripts/install-es.sh
+++ b/enterprise-suite/scripts/install-es.sh
@@ -49,7 +49,7 @@ function usage() {
 # If not using Travis, we stash credentials in a file.  Seems not to work with Travis.
 # This prevents the credentials being written to a log.  (Travis obfuscates the values.)
 function set_credentials_arg() {
-    if [ -z "$TRAVIS" ] ; then
+    if [ -z "${TRAVIS+x}" ] ; then
         CREDS=$(mktemp -t creds.XXXXXX)
         # write creds to file for use by helm
         printf '%s\n' "imageCredentials.username: $1" "imageCredentials.password: $2" >"$CREDS"


dbrinegar:
what does this solve?  if someone has ```export TRAVIS=``` then this will skip credentials where before it would install them.

marcoderama:
I think I understand the change but I don't understand how it will solve anyone's install problems.  Can you elaborate @JustinPihony?

The intent of that check was to determine whether the script was being run within a Travis context or not.  After looking at https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash I see that if Travis sets `TRAVIS` to an empty value, then the check (as is) would assume we're _not_ running in Travis.  The only difference I can see that making though is that Travis builds would fail (which I don't think they are).  So while the proposed change is correct (I think), I don't see how it's going to solve anyones install issues...

Unless I'm misunderstanding something...

marcoderama:
According to https://docs.travis-ci.com/user/environment-variables/, Travis sets `TRAVIS=true`.  I guess I need to confirm but that suggests it's not an empty string, unless someone is setting it that way on purpose.  Regardless, the install script should work for a user no matter which path is chosen there.  (If there's an issue, it should only break things in Travis.)

marcoderama:
@dbrinegar This isn't an issue of skipping credentials.  We check credentials no matter what.  The intent of the changes that got us here was to prevent the credentials showing up in logs when run by a user.  The 'solution' was:
```

JustinPihony:
Misunderstood the path - and the path it was going WAS correct

jsravn:
As an aside, I think having a conditional check for Travis isn't ideal. Seems like a smell to me that maybe the credential file stuff isn't as it should be. I'll dig and see.

jsravn:
Nevermind, see that was done in https://github.com/lightbend/helm-charts/pull/128.

PR #128, opened 2018-10-04
Properly write credentials for helm

marcoderama:


enterprise-suite/scripts/install-es.sh
```diff
@@ -46,17 +45,12 @@ function usage() {
...
-        CREDS=$(mktemp -t creds.XXXXXX)
-        # write creds to file for use by helm
-        printf '%s\n' "imageCredentials.username: $1" "imageCredentials.password: $2" >"$CREDS"
-        HELM_CREDENTIALS_ARG="--values $CREDS"
-    else
-        HELM_CREDENTIALS_ARG="--set imageCredentials.username=$1,imageCredentials.password=$2"
-    fi
+    CREDS=$(mktemp -t creds.XXXXXX)
+    # write creds to file for use by helm
+    printf '%s\n' "imageCredentials:" "   username: $1" "   password: $2" >"$CREDS"
```
JustinPihony: YAML is whitespae sensitive, correct? I see 3 spaces instead of 2. My local was 2..not sure if it'll make a difference

marcoderama: 2 is apparently recommended but 3 etc. will work.

PR #129, opened 2018-10-04
Use build-helm-charts for generating the index

jsravn:
1. Put our artifacts in the new bucket `lightbend-console-charts`
2. Trigger a build on https://travis-ci.com/lightbend/build-helm-charts
   to generate the index.yaml.

See https://github.com/lightbend/build-helm-charts for more details.

### DIFF
--- a/scripts/deploy-to-gcs.sh
+++ b/scripts/deploy-to-gcs.sh
@@ -24,7 +24,7 @@ usage () {
 # GCS project
 GCS_PROJECT=es-repo
 # GCS bucket within project
-: "${GCS_BUCKET:=lightbend-helm-charts}"
+: "${GCS_BUCKET:=lightbend-console-charts}"
 
 # HELM_DIR is helm-charts directory.  Defaults to cwd but can be overridden.
 : "${HELM_DIR:=$(pwd)}"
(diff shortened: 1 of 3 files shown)


jsravn:
After this is merged we can swap `build-helm-charts` to use the prod bucket.

scripts/deploy-to-gcs.sh
```diff
@@ -24,7 +24,7 @@ usage () {
 # GCS project
 GCS_PROJECT=es-repo
 # GCS bucket within project
-: "${GCS_BUCKET:=lightbend-helm-charts}"
+: "${GCS_BUCKET:=lightbend-console-charts}"
```
marcoderama: Do we want to setup versioning on that bucket?  (I see that Lifecycle management is not currently enabled for this bucket.)

jsravn: Sure.

jsravn: @marcoderama can you do this, since you did it already? And please document the process in the Build and Release drive doc so we know how it was setup.

marcoderama: Versioning is now setup.  I've set it to keep at most three versions of any object.

Still on me to update doco.

PR #130, opened 2018-10-04
Update alertmanager.yml example with more details

jsravn:
It now shows how to route to different teams.


jsravn:
Ugh another flake.

jsravn:
Okay not a flake, legit failure :). But it hangs forever.

PR #132, opened 2018-10-05
install-es.sh copes with lacking newline

marcoderama:
Boris couldn't get `install-es.sh` to work.  Turns out it was because his credentials file didn't have a trailing newline character.


enterprise-suite/scripts/install-es.sh
```diff
@@ -63,7 +63,8 @@ function import_credentials() {
         repo_username=$LIGHTBEND_COMMERCIAL_USERNAME
         repo_password=$LIGHTBEND_COMMERCIAL_PASSWORD
     elif [ -e "$LIGHTBEND_COMMERCIAL_CREDENTIALS" ]; then
-        while IFS='=' read -r key value; do
+        while IFS= read -r prop || [ -n "$prop" ]; do
+            read key value <<< $( echo $prop | awk 'BEGIN { FS = ":" } { print $1 " " $2 }' )
```
srikanthops: shouldn't FS = '='

PR #133, opened 2018-10-08
Add persistent volumes for the various pods

jsravn:
* Prometheus
* es-monitor-api
* alertmanager

### DIFF
--- a/enterprise-suite/templates/alertmanager.yaml
+++ b/enterprise-suite/templates/alertmanager.yaml
@@ -10,8 +10,6 @@ spec:
   resources:
     requests:
       storage: {{ .Values.alertmanagerVolumeSize }}
-  parameters:
-    fstype: xfs
 
 ---
 apiVersion: apps/v1beta2
--- a/enterprise-suite/templates/alertmanager.yaml
+++ b/enterprise-suite/templates/alertmanager.yaml
@@ -1,4 +1,19 @@
 ---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: alertmanager-storage
+spec:
+  storageClassName: {{ default .Values.defaultStorageClass .Values.esAlertmanagerStorageClass }}
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ .Values.alertmanagerVolumeSize }}
+  parameters:
+    fstype: xfs
+
+---
 apiVersion: apps/v1beta2
 kind: Deployment
 metadata:
@@ -74,7 +89,8 @@ spec:
           configMap:
             name: {{ .Values.alertManagerConfigMap }}
         - name: data-volume
-          emptydir: {}
+          persistentVolumeClaim:
+            claimName: alertmanager-storage
 
 {{ if .Values.minikube }}
 ---
--- a/enterprise-suite/templates/es-grafana.yaml
+++ b/enterprise-suite/templates/es-grafana.yaml
@@ -81,6 +81,21 @@ spec:
 {{ end }}
 
 ---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: es-grafana-storage
+spec:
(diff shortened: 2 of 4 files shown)


enterprise-suite/templates/alertmanager.yaml
```diff
@@ -1,3 +1,18 @@
...
+  name: alertmanager-storage
+spec:
+  storageClassName: {{ default .Values.defaultStorageClass .Values.esAlertmanagerStorageClass }}
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: {{ .Values.alertmanagerVolumeSize }}
+  parameters:
+    fstype: xfs
```
dbrinegar: This seems an odd detail to have at this level. What's the idea behind specifying it here instead of upstream in the storage class?

dbrinegar: Is the "prometheus uses lots of inodes" problem still true? The TSD has been radically re-written in 2.x, I thought the "block per timeseries" thing was gone and backing store uses a small number of large files, indexing, etc.

jsravn: As far as I know you can't specify fstype at the StorageClass. fstype is used by kubelet when formatting the volume, rather than the provisioning of storage.

jsravn: Based on my experience with running out of inodes on ext4, I think xfs is the safest option regardless. It's also the default filesystem in RHEL if you're concerned about its stability.

jsravn: Okay take that back - seems like you can put it in either place. But we don't have control of the StorageClass - it's expected users will provide that. So I think it's safer to specify in the PVC, rather than relying on users.

jsravn: Also, ext4 has reliability issues on RHEL in my experience.

jsravn:
I'll look into es-grafana, but for next time I think it's okay to approve a PR if it's missing functionality. That can be added into another PR. I'd rather merge often than postpone for more functionality.

enterprise-suite/templates/alertmanager.yaml
dbrinegar: But... if this fstype parameter is incompatible with the user defined storage class, what happens? Why would we force it here, instead of recommend it in the docs for certain kinds of StorageClasses? What if a customer wants some cloud provider blob store thingy we never heard of that doesn't support xfs? Does everything fail at this point? Or is our fstype parameter ignored?

On face this looks like a detail that pierces the abstraction: provide us a StorageClass, then we'll mutate it for you so it has xfs...  see what I mean?

dbrinegar: xfs is cool, I just want to leave this selection up to the user. I don't think we need to force it and can let users try other things if they want. Prometheus 2.x shouldn't have the same urgent problem with inodes, as it only has a few files. Prometheus 1.x had a huge problem here, with a file per timeseries, but we aren't using that.

My minikube install, just poking around, has 19 files in prometheus /data, for 4,895 timeseries. The kernel has 100k inodes in use for all the kubernetes things, but prometheus is using only 28 of those. It seems a tiny amount.

```
root@prometheus-server-fffb9b9cf-rj2tn:/# ls /proc/1/root/proc/1/fd | wc -l
28
```

jsravn: Maybe, I think you can view it either way. I think the support for fstype on StorageClass is recent. You used to have to specify it on the PersistentVolume. The StorageClass defines the backing storage. File system is independent of that, and the user can define what they want.

jsravn: I doubt a user will care until their volume dies.

jsravn:
I don't think it's a productive debate, XFS vs ext4, so I'll just remove XFS in the interest of moving forward.

PR #135, opened 2018-10-09
Provide name to console server port

srikanthops:
/lightbend/es-backend/issues/269

I didn't capture it, but on openshift there was an error message, providing a name resolved it.


jsravn:
Crap, this broke the `kubectl proxy` URL.

jsravn:
Going to rename it so it's not so long in the URL. https://github.com/lightbend/helm-charts/pull/148

PR #137, opened 2018-10-10
Added POD_IP env var to the alertmanager pod template

thinkmorestupidless:
this allows the pod ip to be used by the prometheus operator when setting the 'cluster.listen-address' setting, allowing cluster to form correctly when running on DC/OS

### DIFF
--- a/enterprise-suite/templates/alertmanager.yaml
+++ b/enterprise-suite/templates/alertmanager.yaml
@@ -31,6 +31,10 @@ spec:
               cpu: {{ default .Values.defaultCPURequest .Values.alertManagerCPURequest }}
               memory: {{ default .Values.defaultMemoryRequest .Values.alertManagerMemoryRequest }}
           env:
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
           args:
             - --config.file=/etc/config/alertmanager.yml
             - --storage.path=/data


enterprise-suite/templates/alertmanager.yaml
```diff
@@ -31,6 +31,10 @@ spec:
               cpu: {{ default .Values.defaultCPURequest .Values.alertManagerCPURequest }}
               memory: {{ default .Values.defaultMemoryRequest .Values.alertManagerMemoryRequest }}
           env:
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
           args:
             - --config.file=/etc/config/alertmanager.yml
             - --storage.path=/data
```
jsravn: Also need to pass in the alertmanager argument here with the podIP.

jsravn:
@thinkmorestupidless sorry, do you mind pushing this as a branch rather than a fork? Because the tests require commercial credentials, forks can't run them for security reasons.

enterprise-suite/templates/alertmanager.yaml
thinkmorestupidless: sorry, don't understand what you mean by this

thinkmorestupidless: oh, i get it - i'm an idiot and haven't actually used the IP in the container `--cluster.listen-address=$(POD_IP):6783`

jsravn:
Closing in favour of https://github.com/lightbend/helm-charts/pull/139.

PR #138, opened 2018-10-11
Add target to lint all.yaml

marcoderama:


marcoderama:
No need to fix `alertmanager.yaml` anymore.  Fixed in another PR.

PR #139, opened 2018-10-11
Set the IP address explicitly for the alertmanager

thinkmorestupidless:
sets an env var and then uses that to set --cluster.listen-address


jsravn:
Awesome, thanks!

PR #142, opened 2018-10-11
Wait longer

dbrinegar:
fixes lightbend/es-backend#361

We can also focus our prometheus health rules on our working job / namespace / etc but will do that separately and post refactoring API so don't have to do twice


dbrinegar:
ya, we really don't want this to ever scrape. Without a scrape in queue though, Prometheus will exit and cause Pod thrash..

PR #144, opened 2018-10-11
fix frontend e2e installation

katsutoxin:


.travis.yml
```diff
@@ -29,6 +29,11 @@ jobs:
     - make test CHARTS=enterprise-suite-latest
 
   - name: ES Frontend E2E
+    addons:
+    apt:
```
marcoderama: Doesn't the `apt` block need to be indented?

katsutoxin: good catch!

PR #146, opened 2018-10-12
Do Not Merge Fixes for OpenShift 3.6 support

srikanthops:
This is a place holder to discuss OpenShift 3.6 support.

### DIFF
--- a/enterprise-suite/templates/node-exporter.yaml
+++ b/enterprise-suite/templates/node-exporter.yaml
@@ -3,7 +3,7 @@ kind: ServiceAccount
 metadata:
   name: prometheus-node-exporter
 ---
-apiVersion: apps/v1beta2
+apiVersion: extensions/v1beta1
 kind: DaemonSet
 metadata:
   labels:
--- a/enterprise-suite/templates/prometheus.yaml
+++ b/enterprise-suite/templates/prometheus.yaml
@@ -1,24 +1,3 @@
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: prometheus-server
-rules:
-- apiGroups: [""]
-  resources:
-  - nodes
-  - nodes/proxy
-  - services
-  - endpoints
-  - pods
-  verbs: ["get", "list", "watch"]
-- apiGroups:
-  - extensions
-  resources:
-  - ingresses
-  verbs: ["get", "list", "watch"]
-- nonResourceURLs: ["/metrics"]
-  verbs: ["get"]
 
 ---
 apiVersion: v1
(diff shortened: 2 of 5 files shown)


enterprise-suite/templates/prometheus.yaml
```diff
@@ -60,7 +74,9 @@ apiVersion: v1
 metadata:
   name: prometheus-storage
 spec:
+  {{ if not .Values.openShift36 }}
   storageClassName: {{ default .Values.defaultStorageClass .Values.prometheusStorageClass }}
+  {{  end }}
```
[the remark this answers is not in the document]
srikanthops: PVs and PVCs work fine, the issue was with StorageClassName "standard" I couldn't find what is the equivalent in minishift, 
--set minishift36=true,useEmptyDirVolumes=true

enterprise-suite/templates/node-exporter.yaml
```diff
@@ -3,7 +3,11 @@ kind: ServiceAccount
 metadata:
   name: prometheus-node-exporter
 ---
-apiVersion: apps/v1beta2
+{{ if .Values.openshift36 }}
+  apiVersion: extensions/v1beta1
```
marcoderama: The new lines are indented relative to the old one.  Is that an issue?

enterprise-suite/scripts/install-es.sh
```diff
@@ -41,6 +41,7 @@ function usage() {
     docvar ES_HELM_NAME "Helm release name"
     docvar ES_FORCE_INSTALL "Set to true to delete an existing install first, instead of upgrading"
     docvar DRY_RUN "Set to true to dry run the install script"
+    docvar ES_OPENSHIFT36 "set to true to use api versions for openShift 3.6"
```
marcoderama: Are we likely to see other contexts with different versions down the road?  If so, I might prefer something like:
`ES_KUBERNETES=openshift36` (so we can later have something like `ES_KUBERNETES=minishift`)
(Or whatever is an appropriate name for `ES_KUBERNETES`.)

That way, when the time comes, we won't have to create a new env var or deal with conflicting settings.

srikanthops: agree changed it ES_KUBERNETES

enterprise-suite/scripts/install-es.sh
```diff
@@ -152,13 +154,22 @@ else
     should_upgrade=false
 fi
 
+# OPENSHIFT_V_3.6 Support
+OPENSHIFT36_APIVERSIONS=
+if [ "openshift36" == "$ES_KUBERNETES" ]; then
+    OPENSHIFT36_APIVERSIONS="--set rbacApiVersion=authorization.openshift.io/v1,deploymentApiVersion=apps/v1beta1,daemonSetApiVersion=extensions/v1beta1"
+fi
```
jsravn: Do you think we'll need this? Or can we just document the versions to use?

srikanthops: Removed changes to installer script

srikanthops:
@marcoderama I removed changes to installer script, we will document how to install on OpenShift-3.6 version.
Kindly approve these changes

PR #149, opened 2018-10-16
update test to reflect new URL structure for 'namespace-scoping/paul' branch

pcraddick:
Paired with this PR:

https://github.com/lightbend/es-console-spa/pull/321

It is expected that currently this branch fails to build, because it is running against the current master; (this is a downside of having the code and tests living in separate repos).

The plan is to merge the above PR to master, despite it failing the current (i.e. old) tests. Once in master, we will cut a new image, and configure the test code here to point to it; in which case this PR build will pass, and can itself be merged to master

### DIFF
--- a/enterprise-suite/templates/es-console.yaml
+++ b/enterprise-suite/templates/es-console.yaml
@@ -176,7 +176,7 @@ data:
       # bookmarks / reloads in the /workloads/ path
       # (for our single page app, these should reload index.html but keep the path)
 
-      location ~ (.*/)(workloads/.*) {
+      location ~ (.*/)(namespaces/.*) {
         alias $base/index.html;
         default_type text/html;
         sub_filter 'base href="/"' 'base href="$1"';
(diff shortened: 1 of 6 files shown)


katsutoxin:
You might also want to modify file `enterprise-suite/tests/e2e/cypress/support/navigation.ts`
Change the following url path to new format  :)
```
    static goMonitorPage(workloadId: string, monitorId: string) {
        cy.visit(`/workloads/${workloadId}/monitors/${monitorId}`);
    }
```

enterprise-suite/templates/es-console.yaml
```diff
@@ -176,7 +176,7 @@ data:
       # bookmarks / reloads in the /workloads/ path
       # (for our single page app, these should reload index.html but keep the path)
 
-      location ~ (.*/)(workloads/.*) {
+      location ~ (.*/)(namespaces/.*) {
```
jsravn: I wonder if this would break if the workspace itself has the word "namespaces" in it? It's pretty edge case I suppose. To fix, we could think about including `workloads` too?
```suggestion
      location ~ (.*/)(namespaces/[^/]+/workloads/.*) {
```

dbrinegar: es_workload="namespaces"
namespace="workloads"

evil. That's going in the tests. :-)

PR #150, opened 2018-10-16
install-es.sh can create resource yaml file (aka all.yaml)

marcoderama:
This allows someone that doesn't want to use tiller to generate the resource yaml and then install it themselves.  e.g.

```
ES_EXPORT_FILE=myall.yaml ./scripts/install-es.sh --version=0.10.3 --set minikube=true
kubectl --namespace=lightbend apply -f myall.yaml --validate=false
```

This pulls down the appropriate chartball and then generates the "all.yaml" file from that, modifying things as possibly dictated by `--set` values.

(They only need the `--validate=false` bit for 0.10.3 cuz we have a typo in one of the yaml files.)


marcoderama:
I can generate the yaml file with or without credentials.  Creating with credentials made sense to me.  The generated yaml will be exactly the same as our normal `all.yaml` file with the exception that it also includes something like:

```
# Source: enterprise-suite/templates/commercial-credentials.yaml                                                                     
                                                                                                                                     
---                                                                                                                                  
apiVersion: v1                                                                                                                       
kind: Secret                                                                                                                         
metadata:                                                                                                                            
  name: commercial-credentials                                                                                                       
type: "kubernetes.io/dockerconfigjson"                                                                                               
data:                                                                                                                                
  .dockerconfigjson: eyJhdXRocyI6IHsibGlnaHRiZW5kLWRvY2tlci1jb21tZXJjaWFsLXJlZ2lzdHJ5LmJpbnRyYXkuaW8iOiB7ImF1dGgiOiAiYldGeVkyOWtaWEpoYldFNk1ERm1NR1V4WW1abVptTXpaakEzWmpVeFptSTFPV1EwTldKaE1tWTNNVGMzTkdWaVpXTmxNdz09In19fQ==                                            
---                                                                                                                                  
```

I'm curious to know if people think that's a good thing or needs to be backed out.

dbrinegar:
I think is good to leave all.yaml without a secret. You just install the secret ahead of time and all good.

marcoderama:
I can generate the yaml file with or without credentials. Creating with credentials made sense to me. The generated yaml will be exactly the same as our normal all.yaml file with the exception that it also includes something like:

```
# Source: enterprise-suite/templates/commercial-credentials.yaml

---
apiVersion: v1
kind: Secret
metadata:
  name: commercial-credentials
type: "kubernetes.io/dockerconfigjson"
data:
  .dockerconfigjson: _base64-encoded-credentials==_
---
```
                                                                                                                              
I'm curious to know if people think that's a good thing or needs to be backed out.

marcoderama:
@dbrinegar Fixed credentials issue in latest commit.

enterprise-suite/scripts/install-es.sh
```diff
@@ -40,14 +44,15 @@ function usage() {
...
+    docvar ES_EXPORT_FILE "Export resource yaml to this file.  Does not install in this case."
     docvar DRY_RUN "Set to true to dry run the install script"
     exit 1
 }
 
 # Create arg for the helm install/upgrade lines.  $1 is username.  $2 is password.
 # This prevents the credentials being written to a log.
 function set_credentials_arg() {
-    CREDS=$(mktemp -t creds.XXXXXX)
+    CREDS=$( mktemp -t creds.XXXXXX )
```
jsravn: How come?

enterprise-suite/scripts/install-es.sh
```diff
@@ -151,13 +161,49 @@ else
...
+            esac
+        done
+        set -- "${REST[@]}" # restore the other positional parameters
+    fi
+
+    TDIR=$( mktemp -d 2>/dev/null || mktemp -d -t 'install-es-tdir' )
+    debug helm fetch -d $TDIR "$VERSION_ARG" "$chart_ref"
+    debug helm template --name "$ES_HELM_NAME" --namespace "$ES_NAMESPACE" \
+        "$@" \
+        $TDIR/${ES_CHART}*.tgz > "$ES_EXPORT_FILE"
```
jsravn: It's getting quite gnarly so I think we should stick this all into a function.

enterprise-suite/scripts/install-es.sh
```diff
@@ -151,13 +161,49 @@ else
...
 else
-    debug helm install "$chart_ref" --name="$ES_HELM_NAME" --namespace="$ES_NAMESPACE" \
-        "$HELM_CREDENTIALS_ARG" \
-        $@
+    # Get credentials
+    import_credentials
```
jsravn: This means the resources won't have the secret they need. Is that intentional? Why not just include the secret for them to use?

enterprise-suite/scripts/install-es.sh
marcoderama: No reason.  A style thing I'm trying out...  You don't like?

enterprise-suite/scripts/install-es.sh
marcoderama: Very intentional after getting knuckles rapped by @dbrinegar when I _did_ include them.  Turns out that the credentials are easily decipherable with a few base64 decodes.

My plan is to add a flag or YAEV (yet another env var) to generate the appropriate `kubectl` command.  That way the user could do something like:
`eval $(install-es.sh -c)``
to get their credentials installed in the cluster without having to embed them in the yaml.  Thoughts?

marcoderama:
And yeah, I started exporting to a particular file before I changed `debug` to send to stderr.  I can have the export dump to stdout now.

enterprise-suite/scripts/install-es.sh
marcoderama: Shred the gnar dude!

I tried moving that stuff to a function but it got even hairier cuz I was modifying $@ (which changed inside the function of course).  Leaving as is for now.

enterprise-suite/scripts/install-es.sh
marcoderama: See commit https://github.com/lightbend/helm-charts/pull/150/commits/7c5bc0b3bef0c7319f7ee82b080cd1808e28e90b

The intent is that a user would do:
` eval $(ES_EXPORT_CREDS=true  ./scripts/install-es.sh)`
(After which they could apply the yaml from ES_EXPORT_YAML.)

jsravn: Sorry, I still don't understand the problem. What's wrong with printing out the secret resource for users? It's not being published anywhere. It seems really convenient to provide the entire resource set.

jsravn: Requiring users to use `kubectl` is not ideal - it requires a Kubernetes cluster to be up. It would be better if they had the yaml file they could apply or store in whatever way they desire. Idea here is for maximum flexibility for the user. If we're worried that users don't understand that a `Secret` contains secure info, then we can document that as a note.

enterprise-suite/scripts/install-es.sh
jsravn: Yeah I prefer without spaces - it's more idiomatic.

dbrinegar:
My caveman thought on all of this is around the distinction between "preparing" vs "installing"

In "preparing" you do a dry-run and print all the things out, put them in files, review them, modify, revision them in a git repo, branch, push, etc. This could be as simple as cloning helm-charts or something more complex like a dry-run into files.

In "preparing" you would not want to print your passwords. You generally don't want passwords to be in github, as a way of thinking about it.

Passwords should be done in the "install" step. Install secret in your various clusters could indeed be a one-time thing done once-per-cluster, before you do anything else. It could be done by the admin team, who have their own keybase thing on the side to manage things, or vault, or who knows what. But they aren't putting secrets in github, and are not scala devs, and have more kubectl tooling than they have sbt things. We should consider that context, I think.

enterprise-suite/scripts/install-es.sh
marcoderama: See general comments in this PR for response to your last two comments here.  In a nutshell:
- Can now export either 'console' yaml or 'creds' yaml.
- No more `kubectl` stuff.

marcoderama:
Okay.  Here's the solution to make everyone happy!  Got it?

User can specify either 
 - `ES_EXPORT_YAML=console` to generate resource yaml that does not include explicit credentials.  This output matches our current `all.yaml` exactly.
 - `ES_EXPORT_YAML=creds`to generate resource yaml for `commercial-credentials.yaml` that contains the users credentials in the base64 encoded secret form.

They can `kubectl -apply` the output of either command.  Output of the latter invocation contains insecure credentials but at least they're _aware_ it's a credentials file and should be handled carefully.

I got rid of the code that generate the `kubectl` command.  It didn't work properly in the case that the user had already installed credentials.  Didn't seem worth it to generate all the magic required to cope with all cases.  They can now use the `ES_EXPORT_YAML=creds` thing and are responsible for doing whatever's required, including getting around things like already installed creds.

marcoderama:
So a typical scenario might look like:

```
ES_EXPORT_YAML=creds ./install-es.sh >creds.yaml
kubectl --namespace=lightbend apply -f creds.yaml
chmod go-rw creds.yaml  # or  rm creds.yaml
ES_EXPORT_YAML=console ./install-es.sh --version=0.10.3 --set minikube=true >console.yaml
kubectl --namespace=lightbend apply -f console.yaml
# Maybe enshrine console.yaml in a code repository here
```

They do the creds thing once, and can then generate/tweak/apply the console yaml as many times as they want.

PR #151, opened 2018-10-16
show monitor name instead of fubar annotations

dbrinegar:
since the UI doesn't let one see or edit the annotations

### DIFF
--- a/enterprise-suite/alertmanager/slack.tmpl
+++ b/enterprise-suite/alertmanager/slack.tmpl
@@ -1,4 +1,3 @@
 {{ define "title.slack" }}{{ .GroupLabels.namespace}}/{{ .GroupLabels.es_workload }} [{{ .Status | toUpper }}][{{ .CommonLabels.severity | toUpper }}]{{ end }}
-{{ define "text.slack" }}{{ range $index, $element := .Alerts }}_{{ .Annotations.summary }}_
-    {{ .Annotations.description }}
+{{ define "text.slack" }}{{ range $index, $element := .Alerts }}_{{ .CommonLabels.name }}_
 {{ end }}{{- end -}}


enterprise-suite/alertmanager/slack.tmpl
```diff
@@ -1,4 +1,3 @@
 {{ define "title.slack" }}{{ .GroupLabels.namespace}}/{{ .GroupLabels.es_workload }} [{{ .Status | toUpper }}][{{ .CommonLabels.severity | toUpper }}]{{ end }}
-{{ define "text.slack" }}{{ range $index, $element := .Alerts }}_{{ .Annotations.summary }}_
-    {{ .Annotations.description }}
+{{ define "text.slack" }}{{ range $index, $element := .Alerts }}_{{ .CommonLabels.name }}_
```
alanngai: I assume .CommonLabels.name is the monitor name?  If so, lgtm

jsravn:
@dbrinegar we really wanted to drop the alert description/summary? I thought we just wanted to add the alert name.

dbrinegar:
The UI doesn't show or allow editing of those fields, and the Monitor API doesn't provide generic defaults, so their content at the moment is random and probably not suitable. I think if the UI were to submit them blank and the Monitor API were to come up with generic defaults suitable for the kind of monitor, it would be reasonable to include those messages in the alerts.

enterprise-suite/alertmanager/slack.tmpl
jsravn: Seems like it's not? At least, @srikanthops is having issues with it.

jsravn: Oh I see the issue - .CommonLabels isn't available in the range. It should be `.Labels.name`.

PR #152, opened 2018-10-17
Reload monitor-api on config changes

jsravn:


dbrinegar:
@katsutoxin is working on fixing the UI tests to not require 2 containers in the prometheus pod

enterprise-suite/templates/prometheus.yaml
```diff
@@ -158,6 +158,19 @@ spec:
...
             initialDelaySeconds: 30
 
+        - name: es-monitor-api-configmap-reload
+          image: jimmidyson/configmap-reload:{{ .Values.configMapReloadVersion }}
+          resources:
+            requests:
+              cpu: {{ default .Values.defaultCPURequest .Values.configmapReloadCPURequest }}
+              memory: {{ default .Values.defaultMemoryRequest .Values.configmapReloadMemoryRequest }}
+          args:
+          - --volume-dir=/etc/config
```
dbrinegar: this is the shared config emptyDir for sharing config files between monitor api and prometheus.

The configMap is es-monitor-api-volume

dbrinegar: Shutting down on configMap updates seems ...ok... but there's this scenario where the reloader triggers every time after restart and we get in a death loop. How confident are you that can never happen? Just thinking out loud, a /reload endpoint would be softer as it wouldn't trigger pod recreation. I think all that endpoint would have to do is `	api.scheduleWriteConfig()`

dbrinegar:
hmm I don't see the "merge master" button, but you'll need to merge master to pick up Danny's change to the tests that are otherwise blocking this change

enterprise-suite/templates/prometheus.yaml
jsravn: ConfigMaps don't sync more than every 60s in the first place. So I'm not worried about a death loop type deal. Reason I did it this way was to accommodate for changes to bootstrap monitors as well, without having to do a larger code change in es-monitor-api.

jsravn: Whoops "config-volume" name confused my brain.

jsravn:
Updated to use /reload.

jsravn:
Not clear why the build is failing - looks like prometheus isn't starting up. But I haven't touched it here.

enterprise-suite/templates/prometheus.yaml
```diff
@@ -158,6 +158,19 @@ spec:
...
 
+        - name: es-monitor-api-configmap-reload
+          image: jimmidyson/configmap-reload:{{ .Values.configMapReloadVersion }}
+          resources:
+            requests:
+              cpu: {{ default .Values.defaultCPURequest .Values.configmapReloadCPURequest }}
+              memory: {{ default .Values.defaultMemoryRequest .Values.configmapReloadMemoryRequest }}
+          args:
+          - --volume-dir=/monitor-data
+          - --webhook-url=http://localhost:8180/reload
```
dbrinegar: `127.0.0.1` instead of `localhost` would avoid a host lookup, not that we expect that to matter in this environment, but we've seen customers with non-kubernetes environments where localhost lookup was fubar.

jsravn: Agreed, I've also seen places erroneously map `localhost` to actual IPs.

jsravn:
Urgh I messed up the configmap mount on the configmap reloader. Will fix it in #167.

PR #153, opened 2018-10-17
Fix multi-port service discovery for pods/services

jsravn:
- Fix regex for port replacement - it was broken when the address doesn't
include a port (such as for port 80).

- Only scrape ports that end in `metrics` for Service endpoints, same as we do
for pods. Minikube added the `prometheus.io/scrape` annotation to a bunch of
services, so this is necessary. `prometheus.io/port` is still supported.


dbrinegar:
lgtm. is there a reasonable way to smoke test these?

PR #166, opened 2018-10-31
Enable emptyDir volumes by default

jsravn:
To make it easier to get started on development clusters and minishift.

Also changes flag to `usePersistentVolumes`.

### DIFF
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -58,6 +58,8 @@ prometheusMemoryRequest: 250Mi
 #################################################
 # Persistent Volumes
 #
+# Deprecated value - set to false to disable emptyDir and use persistent volumes. Prefer `usePersistentVolumes`.
+useEmptyDirVolumes: true
 # Set to true to use persistent volumes, otherwise emptyDir volumes are used.
 usePersistentVolumes: false
 # Default StorageClass to use for persistent volumes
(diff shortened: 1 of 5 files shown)


enterprise-suite/values.yaml
```diff
@@ -59,7 +59,7 @@ prometheusMemoryRequest: 250Mi
 # Persistent Volumes
 #
 # Set to true to use emptyDir for all volumes, disabling persistent volumes.
-useEmptyDirVolumes: false
+useEmptyDirVolumes: true
 # Default StorageClass to use for persistent volumes
```
marcoderama: Does the StorageClass comment need something along the lines of "ignored if `useEmptyDirVolumes` is `true`"?

jsravn:
Please re-review as I've changed the flag name.

jsravn:
@marcoderama yes I suppose it's possible. I was thinking we're still "release candidate" so it's okay to make breaking changes like this.

jsravn:
It's not particularly easy to preserve the old flag either (which would take precedence?).

jsravn:
Okay, I've changed it to preserve `useEmptyDirVolumes`. `usePersistentVolumes` takes precedence.

I also modified it so the PVCs are not deleted - which would delete the underlying storage - as a precaution.

jsravn:
@marcoderama I don't think you pushed the approve button? :)

PR #167, opened 2018-10-31
WIP Test that all scrapes are working

jsravn:
This helps verify the relabel rules are working as expected.


enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -237,15 +237,15 @@ scrape_configs:
       - source_labels: [__address__, __meta_kubernetes_pod_annotation_{{ dot2underscore .PrometheusDomain }}_port]
         action: replace
         # Extract IP from address. Address might not contain a port in some cases.
-        regex: ([^:]+):?(?:\d+);(\d+)
+        regex: ([^:]+):?(?:\d*);(\d+)
```
marcoderama: Not that it really matters if this works, but is this working the way you expect? The second group ((?:\d+)) is non-capturing so only there to walk over otherwise non-matched digits in front of the semicolon. I'm not sure exactly what the input to this regex looks like but you're getting the IP address from the first group (([^:]+)). The second group isn't actually doing anything as is. (It matched at least one digit before.)

For example, using https://regex-golang.appspot.com (or regex101.com) to test, I find that the regex
([^:]+);(\d+)
does just as good a job with the input https:172.17.0.17;57 as does your version. (Is that sample input reasonable?) My sense is that it's expected the first group deals with stuff before an optionally present colon, the second group matches the IP, and the third group matches the port.

Same comment applies to the similar changes in this file. Please ignore if I've completely missed the point...

jsravn: Here are possible inputs:

1. `172.17.0.17:9090;9090`
2. `172.17.0.17;9090`

`([^:]+);(\d+)` will not match the first one correctly (it will extract `9090:9090`).

jsravn: `*` is necessary here, otherwise the second group always matches at least one digit, which isn't what we want. We only want it to match the port if it exists. The first group is greedy and should consume everything it can.

jsravn: All ears if you can think of a nicer way to do this :).

marcoderama: Still approved.  :-)  I mis-guessed what the input strings looked like.  Given your examples, regex looks good to me.

marcoderama: _Might_ be worth putting example input in the comment...  Or not.

PR #171, opened 2018-11-02
Attempt, again, to get past smoke_ingress flakiness

marcoderama:
It appears the ingress is reporting as ready but still needs a bit more time in some cases.

### DIFF
--- a/enterprise-suite/tests/smoke_ingress
+++ b/enterprise-suite/tests/smoke_ingress
@@ -2,7 +2,7 @@
 
 ## Test access to the console over an ingress setup
 
-#set -x
+set -x
 
 source smokecommon
 
@@ -32,8 +32,10 @@ ingress_ready() {
     # Debug
     # kubectl describe ing/ingress-test -n lightbend
     # ingress can take up to 1m to get ready
-    sleep 2
     kubectl describe ing/ingress-test -n lightbend | awk 'END { exit !( $1 == "Normal"  &&  $2 == "UPDATE" ) }'
+    status=$?
+    sleep 2
+    return $status
 }
 
 # Makes sure expose-es-console is ready


marcoderama:
This change buys us a couple seconds after the ingress _says_ it's ready.  Worked for me on repeated tests.  Let's see how it goes.

enterprise-suite/tests/smoke_ingress
```diff
@@ -32,8 +32,10 @@ ingress_ready() {
     # Debug
     # kubectl describe ing/ingress-test -n lightbend
     # ingress can take up to 1m to get ready
-    sleep 2
     kubectl describe ing/ingress-test -n lightbend | awk 'END { exit !( $1 == "Normal"  &&  $2 == "UPDATE" ) }'
+    status=$?
+    sleep 2
```
dbrinegar: why move the sleep?

dbrinegar: sleeping more after we have the answer doesn't... change the answer or anything else about this function. debug leftover?

marcoderama: We have _an_ answer but it seems even that was marginally premature.  With the sleep after, the stated condition appears to be reality.

jsravn: Everything is async in k8s so I can see why this might be necessary, as the status value in the apiserver might not reflect reality (but that would be pretty crap).

dbrinegar: ah! totally agree on looking for a better condition if this one is not yet correct. busy wait until we can curl the ingress?

marcoderama: Yeah, I was looking for a test that only involved kubectl things.  After this test passes, indicating the ingress is ready, we test by curling the ingress.  Seemed a bit silly to only test curling the ingress only after determining it was up by curling the ingress...  But if the test I chose doesn't work then there may be no choice.

marcoderama:
@jsravn Yeah, I reran the build maybe 6 or 8 times.  All were successful.

enterprise-suite/tests/smoke_ingress
dbrinegar: maybe instead of sleep 2 here you could busy_wait in test_es_console_responding

something somewhere has to keep trying the curl thing until it works, if we think it may be timing sensitive from this step. sleep 2 is just going to work until we needed sleep 3, should busy_wait instead of sleep

marcoderama:
The change can't hurt so I'm going to give it a spin.

PR #172, opened 2018-11-05
Install troubleshooting script

mitkus:
For https://github.com/lightbend/es-backend/issues/391

`lbc.py` takes care of running Lightbend Console in k8s cluster and diagnosing any problems that might happen.

There are three modes now:
- `install` does what `install-es.sh` did + checks the environment for the required versions of tools, credentials, etc.
- `verify` inspects the running console and compares to the expected list of deployments
- `debug-dump` gathers list of k8s resources in console namespace and the logs of all the container into a zipfile

Things left to do:
- [x] Flag which produces zipfile of k8s info and logs for support
- [x] Make sure the right messages end up in stderr vs. stdout
- [x] Testing


enterprise-suite/scripts/install-es.py
```diff
@@ -0,0 +1,231 @@
...
+    print(cmd)
+    if not args.dry_run:
+        stdout, returncode = run(cmd)
+        print(stdout)
+        if not can_fail and returncode != 0:
+            sys.exit("Command '" + cmd + "' failed:\n" + stdout)
+        return returncode
+    return 0
+
+version_re = re.compile(r'([0-9]+\.[0-9]+(\.[0-9]+)?)')
```
marcoderama: Does this work with versions like 1.0.0-rc.3?  (Not entirely sure how you're using it.)

enterprise-suite/scripts/install-es.py
```diff
@@ -0,0 +1,231 @@
...
+            else:
+                sys.exit("Installed version of '" + name + "' is too old. Found: {}, required: {}"
+                    .format(current, required)) 
+
+    # Non-critical warning
+    print("warning: unable to determine installed version of '" + name + "'")
+
+def preflight_check(creds, minikube=False):
+    # Check versions
+    require_version("docker version -f '{{.Client.Version}}'", '17.06')
```
marcoderama: Can the version numbers be defined as constants?

enterprise-suite/scripts/install-es.py
```diff
@@ -0,0 +1,231 @@
...
+
+def setup_args():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--dry-run', help='only print out the commands that will be executed',
+                        action='store_true')
+    parser.add_argument('--force-install', help='set to true to delete an existing install first, instead of upgrading',
+                        action='store_true')
+    parser.add_argument('--export-yaml', help='export resource yaml to stdout, forces dry_run if not false',
+                        choices=['creds', 'console', 'false'], default='false')
```
marcoderama: Maybe "none" instead of "false" since alternative is not "true"?

marcoderama: Fuhget my suggestion here.  I see I made that silly choice originally and it's probably not a good idea to change things now that the install script is in the wild.

enterprise-suite/scripts/install-es.py
mitkus: It treats 1.0.0-rc.3 same as 1.0.0. Not ideal, but fine for purposes of checking installed versions of kubectl, etc.?

enterprise-suite/scripts/install-es.py
mitkus: Sure, changed that

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
mitkus: I imagined this script would run both in python2 and python3. It doesn't! So I changed to `python2` here.

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,250 @@
...
+        if args.version == None:
+            print(("warning: --version has not been set, helm will use the latest available version. "
+                "It is recommended to use an explicit version."))
+
+        with tempfile.NamedTemporaryFile('w') as creds_tempfile:
+            write_temp_credentials(creds_tempfile, creds)
+            install_helm_chart(args, creds_tempfile.name)
+
+if __name__ == '__main__':
+    main()
```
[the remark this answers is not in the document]
mitkus: done

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+def preinstall_check(creds, minikube=False, minishift=False):
+    assert minikube == False or minishift == False, 'Did not expect both minikube and minishift running'
+
+    check_helm()
+    check_kubectl()
+
+    require_version("docker version -f '{{.Client.Version}}'", REQ_VER_DOCKER)
+
+    if minikube:
+        require_version('minikube version', REQ_VER_MINIKUBE)
```
jsravn: We shouldn't assume because minikube is running that we're deploying to it. You'd need to use `kubectl` to verify that (`kubectl config current-context` == `minikube`).

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    stdout, returncode = run('kubectl version', DEFAULT_TIMEOUT)
+    if returncode != 0:
+        msg = 'Cannot reach cluster with kubectl'
+        if minishift:
+            # Minishift needs special configuration for kubectl to work
+            msg = msg + ". Did you do 'eval $(minishift oc-env)'?"
+        sys.exit(msg)
+
+def preinstall_check(creds, minikube=False, minishift=False):
+    assert minikube == False or minishift == False, 'Did not expect both minikube and minishift running'
```
jsravn: This is possible though, if unlikely. I don't think the install should fail.

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+        check_install(args)
+    
+    if args.subcommand == 'install':
+        creds = import_credentials(args)
+
+        if not args.skip_checks:
+            if args.export_yaml == None:
+                minikube = is_running_minikube()
+                minishift = is_running_minishift()
+                preinstall_check(creds, minikube, minishift)
```
jsravn: Don't we still need to check if helm is installed when doing export_yaml?

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    stdout, returncode = run('helm version', DEFAULT_TIMEOUT)
+    if returncode != 0:
+        sys.exit('Cannot get helm status. Did you set up helm inside your cluster?')
+
+    # TODO: Check if RBAC rules for tiller are set up
+
+    # Check credentials
+    registry = 'lightbend-docker-commercial-registry.bintray.io'
+    stdout, returncode = run('docker login -u {} --password-stdin {}'.format(creds[0], registry),
+                             6, creds[1])
```
jsravn: Unfort. this is not good enough, we also need to try checking an image (perhaps via curl on the tags?). Older lightbend credentials can login but get 404 when they try to pull the commercial registry images. In both cases we should also provide a link to our docs on how to set it up.

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    if returncode != 0:
+        sys.exit('Unable to login to lightbend docker registry. Please check your credentials.')
+    else:
+        run('docker logout ' + registry, DEFAULT_TIMEOUT)
+    
+    # TODO: Try to pull docker image from lighbend registry
+    
+    # At the moment when I try 'docker pull {registry}/enterprise-suite/es-monitor-api:latest' I get:
+    # Error response from daemon: Get https://lightbend-commercial-registry.bintray.io/v2/: 
+    # x509: certificate is valid for *.bintray.com, bintray.com, not lightbend-commercial-registry.bintray.io
```
jsravn: Bintray is broken?

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    install.add_argument('--version', help='es-console version to install', type=str)
+
+    install.add_argument('helm', help='arguments to be passed to helm', nargs=argparse.REMAINDER)
+
+    # Common arguments for all subparsers
+    for subparser in [install, verify, diagnose]:
+        subparser.add_argument('--skip-checks', help='skip environment checks',
+                            action='store_true')
+        subparser.add_argument('--creds', help='credentials file', default='~/.lightbend/commercial.credentials')
+        subparser.add_argument('--namespace', help='namespace to install es-console into/where it is installed', default='lightbend')
```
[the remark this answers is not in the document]
mitkus: Changed

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    
+    printerr('Lightbend Console diagnostic data written to {}'.format(filename))
+
+def setup_args():
+    parser = argparse.ArgumentParser()
+    subparsers = parser.add_subparsers(dest='subcommand', help='sub-command help')
+
+    install = subparsers.add_parser('install', help='install es-console')
+    verify = subparsers.add_parser('verify', help='verify es-console installation')
+    diagnose = subparsers.add_parser('diagnose', help='make an archive with k8s status info for debugging and diagnostic purposes')
```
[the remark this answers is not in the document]
mitkus: Done

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    install.add_argument('--repo', help='helm chart repository', default='https://repo.lightbend.com/helm-charts')
+    install.add_argument('--version', help='es-console version to install', type=str)
+
+    install.add_argument('helm', help='arguments to be passed to helm', nargs=argparse.REMAINDER)
+
+    # Common arguments for all subparsers
+    for subparser in [install, verify, diagnose]:
+        subparser.add_argument('--skip-checks', help='skip environment checks',
+                            action='store_true')
+        subparser.add_argument('--creds', help='credentials file', default='~/.lightbend/commercial.credentials')
```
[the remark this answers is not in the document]
mitkus: Some checks are still done before diagnose or verify (kubectl). I also used to check credentials, but not any more so moved it to install.

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+            archive.writestr(filename, stdout)
+        else:
+            sys.exit(failure_msg + 'unable to get logs for container {} in a pod {}'
+                     .format(container, pod))
+
+    timestamp = datetime.datetime.now().strftime('%Y-%m-%d-%H-%M-%S')
+    filename = 'es-console-diagnostics-{}.zip'.format(timestamp)
+    with zipfile.ZipFile(filename, 'w') as archive:
+        # List all resources in our namespace
+        stdout, returncode = run('kubectl --namespace {} get all'.format(args.namespace), show_stderr=False)
```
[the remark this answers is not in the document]
mitkus: Added `describe all`

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def write_log(archive, pod, container):
+        stdout, returncode = run('kubectl --namespace {} logs {} -c {}'
+                                 .format(args.namespace, pod, container))
+        if returncode == 0:
+            filename = '{}+{}.log'.format(pod, container)
+            archive.writestr(filename, stdout)
+        else:
+            sys.exit(failure_msg + 'unable to get logs for container {} in a pod {}'
+                     .format(container, pod))
```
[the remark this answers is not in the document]
mitkus: Done

enterprise-suite/scripts/lbc.py
mitkus: That's done in the else branch few lines below

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+        else:
+            printerr('Unable to check deployment {} status'.format(name))
+        return False
+
+    status_ok = True
+    status_ok &= deployment_running('es-console')
+    status_ok &= deployment_running('grafana-server')
+    status_ok &= deployment_running('prometheus-alertmanager')
+    status_ok &= deployment_running('prometheus-kube-state-metrics')
+    status_ok &= deployment_running('prometheus-server')
```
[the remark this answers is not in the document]
mitkus: Done

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+
+    if minishift:
+        require_version('minishift version', REQ_VER_MINISHIFT)
+        require_version('oc version', REQ_VER_OC)
+
+    if minikube or minishift:
+        # Check if docker is pointing to a cluster
+        if os.environ.get('DOCKER_HOST') == None:
+            sys.exit('Docker CLI is not pointing to a cluster. Did you run "eval $({} docker-env)"?'
+                     .format('minikube' if minikube else 'minishift'))
```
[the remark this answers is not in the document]
mitkus: Removed

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+            msg = msg + ". Did you do 'eval $(minishift oc-env)'?"
+        sys.exit(msg)
+
+def preinstall_check(creds, minikube=False, minishift=False):
+    assert minikube == False or minishift == False, 'Did not expect both minikube and minishift running'
+
+    check_helm()
+    check_kubectl()
+
+    require_version("docker version -f '{{.Client.Version}}'", REQ_VER_DOCKER)
```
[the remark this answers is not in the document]
mitkus: Removed docker cli dependency. Was using it to do `docker login` credentials checking, but now it's not needed after changing to docker api check.

enterprise-suite/scripts/install-es.py
marcoderama: Sounds good.

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+    install.add_argument('--export-yaml', help='export resource yaml to stdout',
+                        choices=['creds', 'console'])
+    install.add_argument('--helm-name', help='helm release name', default='enterprise-suite')
+    install.add_argument('--local-chart', help='set to location of local chart tarball')
+    install.add_argument('--chart', help='chart name to install from the repository', default='enterprise-suite')
+    install.add_argument('--repo', help='helm chart repository', default='https://repo.lightbend.com/helm-charts')
+    install.add_argument('--version', help='es-console version to install', type=str)
+
+    install.add_argument('helm', help='arguments to be passed to helm', nargs=argparse.REMAINDER)
+
```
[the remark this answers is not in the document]
mitkus: Done

enterprise-suite/scripts/lbc.py
mitkus: I changed creds verification to use Docker API, should be fine now

enterprise-suite/scripts/lbc.py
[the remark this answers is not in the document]
mitkus: Organization is now done by putting different outputs into separate files. When printing to stdout, they are just separated by `=== File: {filename}.log ===` though.

enterprise-suite/scripts/lbc.py
mitkus: Not relevant anymore, authentication uses Docker API now

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,404 @@
...
+                    for container in get_pod_containers(pod):
+                        write_log(archive, pod, container)
+        else:
+            sys.exit(failure_msg + 'unable to pods in the namespace {}'.format(args.namespace))
+    
+    printerr('Lightbend Console diagnostic data written to {}'.format(filename))
+
+def setup_args():
+    parser = argparse.ArgumentParser()
```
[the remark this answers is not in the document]
mitkus: I agree that this should be improved, it's probably not too hard but I'd have to dig deeper into `argparse` lib. I would like to do this together with writing documentation for `lbc.py`, outside of this PR.

enterprise-suite/scripts/lbc.py
mitkus: Added kubectl context verification to the minikube check. I want to keep this particular version check because old minikube version was a problem I encountered myself about a week ago.

enterprise-suite/scripts/lbc.py
mitkus: Removed, shouldn't be possible now with kubectl context checks.

enterprise-suite/scripts/lbc.py
```diff
@@ -0,0 +1,462 @@
...
+
+        creds_exec_arg = ''
+        if args.export_yaml == 'creds':
+            creds_exec_arg = '--execute templates/commercial-credentials.yaml ' + creds_arg
+            printerr('warning: credentials in yaml are not encrypted, only base64 encoded. Handle appropriately.')
+        
+        try:
+            tempdir = make_tempdir()
+            execute('helm fetch -d {} {} {}'
+                .format(tempdir, version_arg, chart_ref))
```
jsravn: This has the same bug as install-es.sh that we fixed recently. This shouldn't be run if `--local-chart` is set.

enterprise-suite/scripts/lbc.py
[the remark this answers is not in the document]
jsravn: We could probably just put the default value in the help message.

enterprise-suite/scripts/lbc.py
mitkus: Done

PR #173, opened 2018-11-05
Fix persistent volumes for deployed clusters

jsravn:
Fix several bugs around the PVC setup:
- PVCs can get mounted as 755, so only root users can write to them. On non-openshift clusters, prometheus runs with uid:gid 65534:65534, so gets permission denied when trying to write to the volume. I've added an init container that sets the permissions as needed. We can't use `fsGroup` because it doesn't work on Openshift.
- Can't have more than one PVC per pod - as they could be created on different availability zones, which would prevent the pod from being scheduled. So I've had to consolidate the monitor-data-volume and prometheus-data-volume into a single PVC.
- Add annotation to prevent helm from deleting the PVCs, to prevent data loss for customers.

Unfortunately this is a breaking change due to the need to consolidate the PVCs. But there is no way around this as far as I can see. Given the broken state of PVCs right now I think it's okay/necessary.

### DIFF
--- a/enterprise-suite/templates/alertmanager.yaml
+++ b/enterprise-suite/templates/alertmanager.yaml
@@ -3,6 +3,8 @@ kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: alertmanager-storage
+  annotations:
+    "helm.sh/resource-policy": keep
 spec:
   storageClassName: {{ default .Values.defaultStorageClass .Values.esAlertmanagerStorageClass }}
   accessModes:
--- a/enterprise-suite/templates/prometheus.yaml
+++ b/enterprise-suite/templates/prometheus.yaml
@@ -91,10 +91,12 @@ spec:
     spec:
       serviceAccountName: prometheus-server
 
-      {{ if .Values.podUID }}
       securityContext:
+      {{ if .Values.podUID }}
         runAsUser: {{ .Values.podUID }}
       {{ end }}
+        # Needed so the prometheus data volume is write-able. The official prometheus image runs with gid 65534.
+        fsGroup: 65534
 
       initContainers:
         - name: setup
@@ -171,29 +173,35 @@ spec:
 
         - name: prometheus-server
           image: prom/prometheus:{{ .Values.prometheusVersion }}
+
           resources:
             requests:
               cpu: {{ default .Values.defaultCPURequest .Values.prometheusCPURequest }}
               memory: {{ default .Values.defaultMemoryRequest .Values.prometheusMemoryRequest }}
+
           args:
             - --config.file=/etc/config/prometheus.yml
             - --storage.tsdb.path=/data
             - --web.console.libraries=/etc/prometheus/console_libraries
             - --web.console.templates=/etc/prometheus/consoles
             - --web.enable-lifecycle
+
           ports:
             - containerPort: 9090
+
           readinessProbe:
             httpGet:
               path: /-/ready
               port: 9090
             timeoutSeconds: 30
+
           livenessProbe:
             httpGet:
               path: /-/healthy
               port: 9090
             timeoutSeconds: 30
             initialDelaySeconds: 30
+
           volumeMounts:
             - name: config-volume
               mountPath: /etc/config
--- a/enterprise-suite/templates/prometheus.yaml
+++ b/enterprise-suite/templates/prometheus.yaml
@@ -45,6 +45,8 @@ kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: prometheus-storage
+  annotations:
+    "helm.sh/resource-policy": keep
 spec:
   storageClassName: {{ default .Values.defaultStorageClass .Values.prometheusStorageClass }}
   accessModes:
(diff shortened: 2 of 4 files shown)


enterprise-suite/templates/prometheus.yaml
```diff
@@ -91,10 +91,12 @@ spec:
     spec:
       serviceAccountName: prometheus-server
 
-      {{ if .Values.podUID }}
       securityContext:
+      {{ if .Values.podUID }}
         runAsUser: {{ .Values.podUID }}
       {{ end }}
+        # Needed so the prometheus data volume is write-able. The official prometheus image runs with gid 65534.
+        fsGroup: 65534
```
seglo: @jsravn I'm curious how you discovered what group the prometheus image runs as?

jsravn: `kubectl exec <pod> -- groups` -> shows `nogroup`
`kubectl exec <pod> -- grep nogroup /etc/groups` -> shows 65534

I'd rather the prometheus pod ran as the root user though - I think the usage of non-root is going to cause complications with openshift (yet to be tested).

seglo: Cool.  I understand running as root is the easiest most cross-platform compatible way, but I also understand the gaping security hole it opens up.  I haven't spent a lot of time understanding how to secure container deployments, but simply running as nobody, or worse, some random user/UID is obviously going to break depending on your environment.

I like the user namespace feature in Docker (https://docs.docker.com/engine/security/userns-remap/), though I've only read about it.  I'm not sure how or if it's that portable across K8s deploys since it relies on docker daemon config.

seglo: Also, how did you troubleshoot the prometheus image by running commands in the pod?  Did you run those commands on an environment that worked (i.e. minikube), or did you do something to keep the container running long enough to execute it on GKE?

jsravn: @seglo Openshift automatically maps the container user to an isolated UID (1000+), so running root containers in Openshift is fine.

In other environments it's not ideal - but there isn't really a great way to handle this - since the container has no way of knowing how that UID is used on the host system, and uids are not namespaced on Linux systems. It would be really nice if Openshift upstreamed this functionality.

The userns remapping is a possibility but not natively supported in K8s yet as far as I know. What you can do though is set `runAsUser` in the security context for a container - which is a good middle ground (but again, it depends on the host system's UIDs, so in a helm chart we can't default that to anything meaingful).

jsravn: > Also, how did you troubleshoot the prometheus image by running commands in the pod? Did you run those commands on an environment that worked (i.e. minikube), or did you do something to keep the container running long enough to execute it on GKE?

I did a `kubectl edit deployment` on `prometheus-server`, and attached the prometheus mount to inspect how it was getting mounted. To inspect the prometheus container, I just looked at my local minikube - but I think you could as easily do a `docker run` and inspect it that way.

jsravn:
Unfortunately this doesn't work on Openshift:
```
27s         1m           14        prometheus-server-6fdc68f5c6.15648ca52781f435                     ReplicaSet                                                               Warning   FailedCreate             replicaset-controller                  Error creating: pods "prometheus-server-6fdc68f5c6-" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{65534}: 65534 is not an allowed group]                     
```

Openshift really wants to manage the security contexts. This is going to be tough to fix generically.

jsravn:
Next problem: having more than one PVC doesn't work because they can get spread across multiple zones. Doh!

enterprise-suite/templates/prometheus.yaml
seglo: Thanks @jsravn !

jsravn:
I've fixed the multi PVC issue - unfortunately it means a breaking change, in that we need to combine two PVCs into a single one, and we were mounting them at `/` rather than using a `subPath`.

enterprise-suite/values.yaml
```diff
@@ -66,8 +66,6 @@ usePersistentVolumes: false
 defaultStorageClass: standard
 # Prometheus volume size - used for storing metrics data.
 prometheusVolumeSize: 256Gi
-# ES Monitor API volume size - used for storing monitors in a git repo.
-esMonitorVolumeSize: 32Gi
```
srikanthops: @jsravn We should remove from Docs..

enterprise-suite/templates/alertmanager.yaml
```diff
@@ -3,6 +3,8 @@ kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: alertmanager-storage
+  annotations:
+    "helm.sh/resource-policy": keep
```
srikanthops: is this annotation necessary when installing without helm

jsravn: Not necessary but it doesn't hurt. I don't think we should make a flag for it.

enterprise-suite/values.yaml
jsravn: yep!

jsravn:
Might have merged prematurely :). Can you guys also take a look?

srikanthops:
Changes look fine. But lets try on different env(s)

PR #174, opened 2018-11-05
Allow for both ES_EXPORT_YAML and ES_LOCAL_CHART

marcoderama:
Also no longer warn about missing `--version` when using` ES_LOCAL_CHART` as that doesn't make sense.  (Actually the opposite.  Now warn if both are specified.)


jsravn:
Thanks!

PR #176, opened 2018-11-07
Add --devel flag so helm can see rc.x versions

marcoderama:
Unbelievably (to me at least) helm ignores by default versions that look like `a.b.c.d-e`.  This means that the install script is currently ignoring our release candidates (e.g. `1.0.0-rc.3`).

c.f. https://github.com/helm/helm/issues/2407

Adding the `--devel` flag will pick them up.

### DIFF
--- a/enterprise-suite/scripts/install-es.sh
+++ b/enterprise-suite/scripts/install-es.sh
@@ -196,7 +196,7 @@ if [ "false" != "$ES_EXPORT_YAML" ]; then
 
     if [ -z "$ES_LOCAL_CHART" ]; then
         TDIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'install-es-tdir')
-        debug helm fetch -d $TDIR "$VERSION_ARG" "$chart_ref"
+        debug helm fetch -d $TDIR --devel "$VERSION_ARG" "$chart_ref"
         chart_ref=$TDIR/${ES_CHART}*.tgz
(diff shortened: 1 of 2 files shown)


enterprise-suite/scripts/install-es.sh
```diff
@@ -196,7 +196,7 @@ if [ "false" != "$ES_EXPORT_YAML" ]; then
 
     if [ -z "$ES_LOCAL_CHART" ]; then
         TDIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'install-es-tdir')
-        debug helm fetch -d $TDIR "$VERSION_ARG" "$chart_ref"
+        debug helm fetch -d $TDIR --devel "$VERSION_ARG" "$chart_ref"
```
srikanthops: according to man page --devel option is ignored when version is specified.

marcoderama: Yeah.  `$VERSION_ARG` will have a value if and only if the user specifies a value on the `install-es.sh` command line.  If they do, `--devel` will be ignored.  If they don't, it'll be empty and `--devel` will come into play.

marcoderama:
And now that I think about it, we may want to change the install script to allow for installation of "stable" or "bleeding edge" versions.  If "stable", then we don't consider `-rc.x` versions and drop the `--devel` flag internally.  Later...

PR #178, opened 2018-11-09
Don't use metric_relabel_rules

jsravn:
They don't capture automatic metrics (up, scrape_duration_time, etc.).
As a result, we can't create custom monitors on any of these metrics -
as they don't get `es_monitor_type`.

We are using metric_relabel_rules as a way to avoid labeling automatic
metrics in some cases (cadvisor/kube_state_metrics), so I've kept these
in place. However, I think it might be better if we just fixed the rules
to handle automatic metrics as it's not obvious. For now I've added a
comment.

For https://github.com/lightbend/es-backend/issues/408

### DIFF
--- a/enterprise-suite/es-monitor-api/prometheus.yml
+++ b/enterprise-suite/es-monitor-api/prometheus.yml
@@ -50,6 +50,7 @@ scrape_configs:
       - source_labels: [__meta_kubernetes_node_address_InternalIP]
         target_label: node_ip
 
+    # These are explicitly metric_label_configs so they don't affect automatic metrics (up, scrape_time_duration, etc.).
     metric_relabel_configs:
       - source_labels: [pod_name]
         target_label: kubernetes_pod_name
@@ -117,7 +118,10 @@ scrape_configs:
         action: replace
         target_label: node_ip
 
+{{ .MonitorTypeRules | indent 6 }}
+
     # es_workload labels:
+    # These are explicitly metric_label_configs so they don't affect automatic metrics (up, scrape_time_duration, etc.).
     metric_relabel_configs:
       - source_labels: [created_by_kind, pod, es_workload]
         regex: 'ReplicaSet;(.+)-[^-]+-[^-]+;'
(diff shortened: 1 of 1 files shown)


jsravn:
This also illustrates the big gap we have in our testing. We don't test or verify any of our monitors work correctly. Regressions are basically impossible to avoid in this scenario. We should start working on that asap post-1.0. 

- https://github.com/lightbend/es-backend/issues/271
- https://github.com/lightbend/es-backend/issues/286 is a start

jsravn:
I'll think if there's a way we can add this to the smoke tests. Bash is unpleasant though for this kind of more complex testing.

dbrinegar:
I don't see how this could work, as you need to determine the workload first. monitor type has to come last

dbrinegar:
the "why use one or the other" is explained here https://www.robustperception.io/relabel_configs-vs-metric_relabel_configs

jsravn:
> monitor type has to come last

Ah whoops - yes, I need to fix this for the cadvisor and kube-state-metrics relabeling. That's a pain because they don't currently work with automatic metrics (up, scrape_duration_time, etc.) as-is.

> the "why use one or the other" is explained here 

Yep, I understand the difference I believe. It's why it's broken for the automatic metrics. It's not super clear to me though why we put them in `metric_relabel_config` in the first place though. My reverse engineering of it is: so it skips the automatic metrics when extracting the workload in cadvisor/kube-state-metrics relabeling.

jsravn:
I think we have two different usecases:

1. Pod/service scrapes. Here it makes sense to define es_workload and es_monitor_type in normal relabel rules, since it's prescrape and based on the target info.
2. cadvisor (kubelet)/kube-state-metrics. The labels come from the `/metrics` page itself, so we need to handle it in `metric_relabel_config`.

I fixed it for (1), but inadvertently broke it for (2). So I need to handle case (2).

jsravn:
Okay I fixed the kube-state-metrics. The pod and service rules should be good as is.

jsravn:
Please have another look @dbrinegar.

jsravn:
Yeah we really do need a test for this, and the other monitor bugs we've been finding.

I'm really hesitant to expand the smoketests significantly though - I'm worried about the maintainability of a large bash smoke test suite, especially as we add more complexity.

jsravn:
Going to merge then work on a test.

PR #179, opened 2018-11-12
Add ability to skip creation of PVCs

jsravn:
As we no longer delete pre-existing PVCs to prevent data loss.

This allows a user to do a force-install (reinstall) of console and use
the prior PVCs.


enterprise-suite/values.yaml
```diff
@@ -62,6 +62,8 @@ prometheusMemoryRequest: 250Mi
 useEmptyDirVolumes: true
 # Set to true to use persistent volumes, otherwise emptyDir volumes are used.
 usePersistentVolumes: false
+# If true, the chart creates persistent volumes using PVCs. Set to false to reuse previously created persistent volumes.
+createPersistentVolumes: true
```
marcoderama: I fear people will be confused by "usePersistentVolumes" and "createPersistentVolumes".  As it is, `createPersistentVolumes` defaults to true, suggesting we'll create PVs, but since `usePersistentVolumes` is false, that won't actually happen.

It's even trickier to grasp if one sees `useEmptyDirVolumes` up there.  I think we should have a section at the bottom for deprecated settings and move it there.  Maybe...

Not sure I have a suggestion though...

jsravn: I originally called it `reusePersistentVolumes` which I think was a lot more confusing :).

marcoderama: Had one thought on this.  What about indenting the `createPersistentVolumes` setting under that for `usePersistentVolumes`?  Might make it clearer it's subordinate.  (Or might look like we screwed up the indentation...)

jsravn: We could do that, like `persistentVolumes.create` and `persistentVolumes.use` but I'm not sure it is making it clearer.

jsravn: Can you give me an example? Maybe you have something else in mind.

marcoderama: I mean physically indent in the file but I realize now that's a non-starter given it's a yaml file.

PR #181, opened 2018-11-12
Make images configurable in a consistent way

jsravn:
This is per https://docs.helm.sh/chart_best_practices/, but also has
other benefits:
- For airgapped installs, we can specify a differerent location for all
of the images.
- For the openshift registry we can provide different image names.
- Makes it clear all of our image dependencies in one place.

### DIFF
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -20,9 +20,9 @@ nodeExporterVersion: v0.15.2
 # janeczku/go-dnsmasq
 goDnsmasqVersion: release-1.0.7
 # alpine
-alpineVersion: latest
+alpineVersion: 3.8
 # busybox
-busyboxVersion: latest
+busyboxVersion: 1.29
 
 #################################################
 # Image names to use. Combined with the versions above, gives the fully qualified docker image name.
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -17,6 +17,38 @@ configMapReloadVersion: v0.2.2
 kubeStateMetricsVersion: v1.2.0
 # prom/node-exporter
 nodeExporterVersion: v0.15.2
+# janeczku/go-dnsmasq
+goDnsmasqVersion: release-1.0.7
+# alpine
+alpineVersion: latest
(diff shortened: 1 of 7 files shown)


enterprise-suite/values.yaml
```diff
@@ -17,6 +17,38 @@ configMapReloadVersion: v0.2.2
...
 nodeExporterVersion: v0.15.2
+# janeczku/go-dnsmasq
+goDnsmasqVersion: release-1.0.7
+# alpine
+alpineVersion: latest
+# busybox
+busyboxVersion: latest
+
+#################################################
+# Image names to use. Combined with the versions above, gives the fully qualified docker image name.
```
marcoderama: Do we want to list the names before the images?

enterprise-suite/values.yaml
```diff
@@ -17,6 +17,38 @@ configMapReloadVersion: v0.2.2
...
+# configmap-reload image
+configMapReloadImage: jimmidyson/configmap-reload
+# kube-state-metrics image
+kubeStateMetricsImage: gcr.io/google_containers/kube-state-metrics
+# node-exporter image
+nodeExporterImage: prom/node-exporter
+# go-dnsmasq image
+goDnsmasqImage: janeczku/go-dnsmasq
+# alpine image
+alpineImage: alpine
```
marcoderama: I guess we use latest for this and busybox now, but do we want to consider specifying particular versions?

jsravn: Yes we should!

enterprise-suite/values.yaml
jsravn: I'm not sure I follow. Can you give me an example?

enterprise-suite/values.yaml
jsravn: I'll make that change.

enterprise-suite/values.yaml
marcoderama: We list versions starting at line 2, and then later (line 28) list names.  I just wondered if it made more sense to list the names before the versions.

PR #182, opened 2018-11-12
Timeout tests after 90s (default)

marcoderama:
Added a `timeout` function for running tests.  Default is 90s.  If nothing times out, then behavior is unchanged.  If there's a timeout, then the FAILURES are incremented, and also tracked as a timeout failure.

For example, I hacked the following change to `smokecommon`:
```
@@ -152,10 +152,11 @@ test_es_console_responding() {
 # Grafana access test
 test_grafana_responding() {
   grafana_responding() {
+      sleep 4
       results=$(curl_query $1 "api/org" | jq '.id')
       ! [[ $results -eq 0 ]]
   }
-  timeout grafana_responding "$@"
+  timeout 1 grafana_responding "$@"
   T $? grafana service accessible via $1
 }
```

That forced the `test_grafana_responding` test to take at least 4 seconds.  Note also that I specified a non-default timeout of 1 second for this test.  The output from running `smoke_services` looked like:

```
PASS: Lightbend Console accessible via http://192.168.99.100:30080
*** TIMEOUT: grafana service accessible via http://192.168.99.100:30030
*** FAIL: grafana service accessible via http://192.168.99.100:30030
./smoke_services: line 23: 84374 Terminated: 15          "$@"
PASS: prometheus service accessible via http://192.168.99.100:30090
PASS: alertmanager accessible via http://192.168.99.100:30093
*** TIMEOUT: grafana service accessible via http://192.168.99.100:30080/service/grafana
*** FAIL: grafana service accessible via http://192.168.99.100:30080/service/grafana
./smoke_services: line 27: 84392 Terminated: 15          "$@"
PASS: prometheus service accessible via http://192.168.99.100:30080/service/prometheus
PASS: es-monitor-api service accessible via http://192.168.99.100:30080/service/es-monitor-api
Some smoke_services tests failed!  PASS: 5  FAIL: 2  (TIMEOUTS: 2)
```

### DIFF
--- a/enterprise-suite/tests/smokecommon
+++ b/enterprise-suite/tests/smokecommon
@@ -1,5 +1,3 @@
-#!/bin/bash
-
 # Note that parts of this depend on minikube at this time.
 
 #set -x
@@ -8,6 +6,8 @@
 #set -e
 set -o pipefail
 set -o nounset
+
+source timeout.fn
 
 : "${CURRENT_CONTEXT:=minikube}"
 
@@ -32,8 +32,37 @@ busy_wait() {
   done
 }
 
-TEST_SUCCESSES=0
-TEST_FAILURES=0
+timeout() {
+    # default timeout is 90s.  If $1 is a number, it will be used as timeout value
+    timeout=90
+    re='^[0-9]+$'
+    if [[ "$1" =~ $re ]] ; then
+        timeout=$1
+        shift
+    fi
+    timeout_function $timeout "$@"
+}
+
+
+# Setup fds for writing/reading success/failure counts.  Allows things to work across subshells
+SUCCESSES=$(mktemp -t pipe.XXXXX)
+exec 3>$SUCCESSES
+exec 4<$SUCCESSES
+# unlink so we don't have to clean up later
+rm $SUCCESSES
+
+FAILURES=$(mktemp -t pipe.XXXXX)
+exec 6>$FAILURES
+exec 7<$FAILURES
+# unlink so we don't have to clean up later
+rm $FAILURES
+
+TIMEOUTS=$(mktemp -t pipe.XXXXX)
+exec 8>$TIMEOUTS
+exec 9<$TIMEOUTS
+# unlink so we don't have to clean up later
+rm $TIMEOUTS
+
 
 # test runner
 # command to test &&:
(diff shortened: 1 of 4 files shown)


enterprise-suite/tests/timeout.fn
```diff
@@ -0,0 +1,68 @@
...
+# Execute a command with a timeout
+
+# This has been modified so that it's a function instead of a command.
+# Also return 142 on timeout to match "kill SIGALARM" return value.
+
+# License: LGPLv2
+# Author:
+#    http://www.pixelbeat.org/
+# Notes:
+#    Note there is a timeout command packaged with coreutils since v7.0
```
jsravn: @marcoderama is there a reason not to use the coreutils timeout command?

enterprise-suite/tests/timeout.fn
```diff
@@ -0,0 +1,68 @@
...
+a=
+
+timeout_cleanup()
+{
+	trap - ALRM                #reset handler to default
+    kill -ALRM $a 2>/dev/null  #stop timer subshell if running
+    last=$!
+	kill $last 2>/dev/null &&  #kill last job
+      return 142               #exit with 142 if it was running
+}
```
jsravn: Formatting seems messed up here.

enterprise-suite/tests/timeout.fn
```diff
@@ -0,0 +1,68 @@
...
+
+timeout_watchit()
+{
+    trap "timeout_cleanup" ALRM
+	sleep $1 & wait
+	kill -ALRM $$
+}
```
jsravn: Same - `trap` line is not indented the same.

enterprise-suite/tests/timeout.fn
mitkus: Can only depend on coreutils on linux. Mac OS, for example, doesn't have `timeout`.

enterprise-suite/tests/smokecommon
```diff
@@ -32,8 +32,37 @@ busy_wait() {
...
+exec 7<$FAILURES
+# unlink so we don't have to clean up later
+rm $FAILURES
+
+TIMEOUTS=$(mktemp -t pipe.XXXXX)
+exec 8>$TIMEOUTS
+exec 9<$TIMEOUTS
+# unlink so we don't have to clean up later
+rm $TIMEOUTS
+
```
jsravn: Would this be simpler if we used `mkfifo` instead? Then we can write/read from the named pipes rather than anonymous file descriptors, which I find pretty hard to follow.

marcoderama: I originally tried with mkfifo but it just wouldn't work.  I would close it for writing but that would close it for reading at the same time and it would disappear.  I tried various stunts but all to no avail.  It just worked as soon as I moved to an actual file.

I also tried to use variables for the file descriptors but that got even uglier fast.  (cf. https://stackoverflow.com/questions/8295908/how-to-use-a-variable-to-indicate-a-file-descriptor-in-bash)  And I skipped 5 cuz some page suggested that was a good idea...

So it is what it is.

marcoderama:
And yeah @jsravn, I think we've pushed the shell scripting as far as we want to here.

PR #183, opened 2018-11-13
Improvements for lbc.py

mitkus:
This PR contains a number of smaller changes and improvements to `lbc.py`:
- Make it work on systems that don't have `python2`, refuse to run with an error message on systems where `python` is `python3`.
- Add `uninstall` subcommand which basically does `helm delete --purge` in a safer way
- Add `--set` argument (https://github.com/lightbend/es-backend/issues/391)
- Add `--wait` argument
- Handle `helm status` output better (https://github.com/lightbend/es-backend/issues/433)
- Add `--reuse-resources` flag which should help with installing multiple consoles in different namespaces (https://github.com/lightbend/es-backend/issues/432). Note: this will work in the next release which will include @jsravn changes (https://github.com/lightbend/helm-charts/pull/179, https://github.com/lightbend/helm-charts/pull/180), so it is behind a flag for now.
- Change `enterprise-suite/Makefile` to use `lbc.py` for running minikube tests
- Fix a bug where script would crash when trying to run unavailable command in linux (eg. when running `minishift status` when minishift is not installed)

### DIFF
--- a/enterprise-suite/Makefile
+++ b/enterprise-suite/Makefile
@@ -121,12 +121,12 @@ HELM_VALUES_FLAG=$(if $(HELM_VALUES_FILE),--values $(HELM_VALUES_FILE))
 install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm  ## Install local tarball
 	# We set a podUID here for test purposes to ensure everything works as non-root.
 	# We usePersistentVolumes to ensure PVCs work as expected.
-	-ES_LOCAL_CHART=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz \
-	    $(CHART_DIR)/scripts/lbc.py install --wait --set minikube=$(MINIKUBE),podUID=10001,usePersistentVolumes=true -- $(HELM_VALUES_FLAG)
+	-$(CHART_DIR)/scripts/lbc.py install --local-chart=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz \
+		--wait --set minikube=$(MINIKUBE),podUID=10001,usePersistentVolumes=true -- $(HELM_VALUES_FLAG)
 
 install-dev: install-helm ## Install content of enterprise-suite/ folder
-	ES_LOCAL_CHART=$(CHART_DIR) \
-	    $(CHART_DIR)/scripts/lbc.py install --wait --set minikube=$(MINIKUBE) -- $(HELM_VALUES_FLAG)
+	-$(CHART_DIR)/scripts/lbc.py install --local-chart=$(CHART_DIR) --wait \
+		--set minikube=$(MINIKUBE) -- $(HELM_VALUES_FLAG)
 
 help:  ## Print help for targets
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(lastword $(MAKEFILE_LIST)) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
--- a/enterprise-suite/Makefile
+++ b/enterprise-suite/Makefile
@@ -33,7 +33,7 @@ package: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz  ## Build chart tarball
 
 script-tests:
 	$(CHART_DIR)/tests/lib/bats/bats $(CHART_DIR)/tests
-	python2 $(CHART_DIR)/scripts/lbc_test.py
+	python $(CHART_DIR)/scripts/lbc_test.py
 
 minikube-tests: build install-local
 	@echo Installed Console from helm charts: $(CHART)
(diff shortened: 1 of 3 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -51,14 +80,14 @@ def make_tempdir():
 # Returns (stdout, returncode) tuple. If timeout
 # occured, returncode will be negative (-9 on macOS).
```
marcoderama: nitpick:  "occurred"

enterprise-suite/scripts/lbc.py
```diff
@@ -184,11 +213,81 @@ def preinstall_check(creds, minikube=False, minishift=False):
...
+        all_found = True
+        found_crs = []
+        for cr in CONSOLE_CLUSTER_ROLES:
+            if cr in stdout:
+                found_crs.append(cr)
+            else:
+                all_found = False
+
+        if not all_found and len(found_crs) > 0:
+            fail('Found some cluster roles from previous console install, but not all: {}. Please clean up manually.'
```
marcoderama: Nitpick alert!  "...clean THEM up manually"?  (Would match the same line in `are_pvcs_created`.)

enterprise-suite/scripts/lbc.py
```diff
@@ -184,11 +213,81 @@ def preinstall_check(creds, minikube=False, minishift=False):
...
+                 .format(str(found_crs)))
+
+        return all_found
+    return False
+
+def install(creds_file):
     creds_arg = '--values ' + creds_file
+    version_arg = ('--version ' + args.version) if args.version != None else '--devel'
```
marcoderama: The `install-es-sh` script would allow `--version=blah` as well as `--version blah`.  (Only cuz helm does.)  Does this do the same?  (I'm no python expert.)  I don't think so, and frankly I think that's okay.  I only mention it in case someone does think it matters.

enterprise-suite/scripts/lbc.py
```diff
@@ -184,11 +213,81 @@ def preinstall_check(creds, minikube=False, minishift=False):
...
     helm_args = ' '.join([arg for arg in args.helm if arg != '--'])
-    version_arg = ('--version ' + args.version) if args.version != None else '--devel'
+
+    # Add '--set' arguments to helm_args
+    if args.set != None:
+        helm_args += ' '.join(['--set ' + keyval for keyval in args.set])
+
+    # Helper to dynamically add new helm args
+    def helm_add_arg(helm_args, arg):
+        if arg not in helm_args:
```
marcoderama: I can't find proof atm so you should probably ignore this comment, but I have this feeling helm tolerates some args specified multiple times, and then uses the last value.  Not sure this would behave the same way.

And again, I'm not sure about this so feel free to ignore this comment.

enterprise-suite/scripts/lbc.py
```diff
@@ -232,17 +331,39 @@ def install_helm_chart(creds_file):
...
+        if args.wait:
+            helm_args = helm_add_arg(helm_args, '--wait')
+
+        if args.reuse_resources:
+            # Reuse PVCs if present
+            if are_pvcs_created(args.namespace):
+                printerr('warning: found existing PVCs from previous console installation, will reuse them')
+                helm_args = helm_add_arg(helm_args, '--set createPersistentVolumes=false')
+
+            # Reuse closter roles if present
```
marcoderama: closter -> cluster

enterprise-suite/scripts/lbc.py
```diff
@@ -410,29 +544,47 @@ def setup_args(argv):
...
 
     # Install arguments
-    install.add_argument('--dry-run', help='only print out the commands that will be executed',
-                        action='store_true')
     install.add_argument('--force-install', help='set to true to delete an existing install first, instead of upgrading',
                         action='store_true')
     install.add_argument('--export-yaml', help='export resource yaml to stdout',
                         choices=['creds', 'console'])
-    install.add_argument('--helm-name', help='helm release name', default='enterprise-suite')
+    install.add_argument('--reuse-resources', help='try to reuse PVCs and cluster roles from a previous install',
```
marcoderama: May not be realistic, but is it conceivable someone will only want to reuse only one of PVCs or roles?  How would that work?

marcoderama:
It may be a bit late in the game for this but...  Can we change the name of this script?  In particular I mean drop the `.py` suffix.  That it's written in python is (should be) irrelevant to the user.

@jsravn may feel otherwise I suspect.  :-)  I thought of this when reviewing the doco changes.  Things like
`./lbc.py install`
just look inelegant.  imho of course.

enterprise-suite/scripts/lbc.py
mitkus: That should work - when this flag is set, it checks for existing PVCs and cluster roles separately. If they don't exist, it still creates them. So it could, for example, reuse PVCs and create cluster roles. If no reusable resources exist, flag does nothing besides checking.

enterprise-suite/scripts/lbc.py
mitkus: Just tried it out - yes, it supports both.

enterprise-suite/Makefile
```diff
@@ -33,7 +33,7 @@ package: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz  ## Build chart tarball
 
 script-tests:
 	$(CHART_DIR)/tests/lib/bats/bats $(CHART_DIR)/tests
-	python2 $(CHART_DIR)/scripts/lbc_test.py
+	python $(CHART_DIR)/scripts/lbc_test.py
```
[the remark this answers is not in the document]
mitkus: I didn't want to stomp over any existing tests. Let's keep the bats tests while we still have `install-es.sh` in the repo, can remove both once we're sure nobody is using the old script.

enterprise-suite/scripts/lbc.py
mitkus: Thanks, fixed!

enterprise-suite/scripts/lbc.py
mitkus: Fixed

enterprise-suite/scripts/lbc.py
mitkus: Made the help message more clear about this.

enterprise-suite/scripts/lbc.py
mitkus: Changed

enterprise-suite/scripts/lbc.py
mitkus: Right, I checked and helm uses the last value. Simplified this so that we're doing the same.

mitkus:
I actually agree that `./lbc install` would be nicer. Anyway, I'll make a separate PR for changing the name if nobody is against it.

jsravn:
Yeah, I prefer keeping the suffix, this is pretty standard practice. Without a suffix it looks like a binary to me. Also for us devs, lack of a prefix means the IDE can't determine the filetype.

marcoderama:
The convention seems cargo-culty to me.  (Your fault for introducing me to that term @jsravn)  It serves no purpose in unix-land.  And emacs will happily determine that it's a python file so I'm not sure what you mean about IDE problems.  ;-)    $0.02.

PR #184, opened 2018-11-14
Add exposeServices flag, deprecate minikube flag

mitkus:
For https://github.com/lightbend/es-backend/issues/410

I'm a total noob with Helm charts. Does this look ok?

Tests on minikube seem to work:

```
scripts/lbc.py install --local-chart=. --set exposeServices=NodePort
```
```
scripts/lbc.py install --local-chart=. --set exposeServices=LoadBalancer
```

### DIFF
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -81,6 +81,15 @@ alertManagerConfigMap: alertmanager-default
 # Service type for exposing Console outside the cluster. Can be "NodePort", "LoadBalancer" or "false".
 exposeServices: false
 
+# If exposeServices is set to "NodePort" or "LoadBalancer", Grafana will be exposed on this port.
+esGrafanaExposePort: 30030
+# If exposeServices is set to "NodePort" or "LoadBalancer", Console will be exposed on this port.
+esConsoleExposePort: 30080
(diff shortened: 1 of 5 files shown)


enterprise-suite/values.yaml
```diff
@@ -78,6 +78,8 @@ imageCredentials:
 # Alertmanager ConfigMap. Set to the name of a ConfigMap, with a file `alertmanager.yml`.
 alertManagerConfigMap: alertmanager-default
 
+# Service type for exposing Console outside the cluster. Can be "NodePort", "LoadBalancer" or "false".
```
marcoderama: Do we want to document `LoadBalancer` yet, cuz I'm assuming there's more to making that work than just setting `exposeServices: LoadBalancer`.  I could be wrong though...

mitkus: Yeah, it would also need an Ingress resource. Not sure we can set it up in our chart if exposeService==LoadBalancer, it needs to be configured for different network setups.

mitkus:
@marcoderama I agree it would be good to get rid of deprecated values in the templates. On the other hand, that complicates `values.yaml` which is very nice and readable now. There's also deprecated `useEmptyDirVolumes` handled in the same way. I guess we can do a separate PR later to change how deprecated values are handled?

PR #189, opened 2018-11-19
Make `./lbc.py install` try to reuse resources from previous failed install, also better python3 behaviour

mitkus:
For https://github.com/lightbend/es-backend/issues/433

In some cases, I get failed helm install where some resources still get created. After that, `./lbc.py install` tries to clean that up with `helm delete --purge`, but our policy for PVCs prevents them from being deleted. This PR makes `lbc.py` check for existing PVCs if previous install failed, just like when `--reuse-resources` flag is provided.

Also, this PR slightly improves running on python3 - instead of just complaining and quitting, now we try to spawn python2 process. Tested only on Mac OS so far.

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -4,18 +4,20 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import sys 
+import subprocess
 
 # Note: this script expects to run on python2. Some systems map 'python'
 # executable name to python3, detect that and complain.
 if sys.version_info >= (3, 0):
-    sys.exit(("It appears 'python' in your PATH is Python 3.X, please "
-              "run this script using Python 2: 'python2 lbc.py'"))
+    print(("It appears 'python' in your PATH is Python 3.X, please "
+           "run this script using Python 2: 'python2 lbc.py'\n"))
+    proc = subprocess.run(['python2'] + sys.argv)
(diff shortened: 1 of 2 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -360,7 +348,21 @@ def install(creds_file):
...
                  .format(args.helm_name, status))
-    
+
+        if args.wait:
+            helm_args += ' --wait'
+
+        if args.reuse_resources or status == 'failed':
+            # Reuse PVCs if present
+            if are_pvcs_created(args.namespace):
+                printerr('warning: found existing PVCs from previous console installation, will reuse them')
```
jsravn: I'm not sure it should be a warning - since it might be okay and expected behaviour.

enterprise-suite/scripts/lbc.py
```diff
@@ -360,7 +348,21 @@ def install(creds_file):
...
+                helm_args += ' --set createPersistentVolumes=false'
+
+            # Reuse cluster roles if present
+            if are_cluster_roles_created():
+                printerr('warning: found existing cluster roles from previous console installation, will reuse them')
```
jsravn: Same as my prior comment.

enterprise-suite/scripts/lbc.py
mitkus: Changed to `info:`

PR #194, opened 2018-11-20
Values for hooking up external alertmanager

mitkus:
For https://github.com/lightbend/es-backend/issues/340

Depends on https://github.com/lightbend/es-monitor-api/pull/77

This adds two new values to our helm chart: `alertManagerAddress` and `alertManagerPort`. It allows to provide an existing alertmanager which es-console prometheus will report to. If `alertManagerAddress` is not set, it behaves like previously and deploys new alertmanager.

This also fixes https://github.com/lightbend/es-backend/issues/438


mitkus:
Tested with `es-monitor-api` change https://github.com/lightbend/es-monitor-api/pull/77:

```
scripts/lbc.py install --local-chart=. --reuse-resources --namespace=monitoring --set exposeServices=NodePort --set esMonitorVersion=latest
```

And with a previously deployed alertmanager:

```
scripts/lbc.py install --local-chart=. --reuse-resources --namespace=monitoring --set exposeServices=NodePort --set esMonitorVersion=latest --set alertManagerAddress=alertmanager2.monitoring.svc.cluster.local
```

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -13,12 +13,9 @@ rule_files:
 
 alerting:
   alertmanagers:
-  - kubernetes_sd_configs:
-    - role: pod
-    relabel_configs:
-    - source_labels: [__meta_kubernetes_namespace,__meta_kubernetes_pod_label_app,__meta_kubernetes_pod_label_component,__meta_kubernetes_pod_container_port_number]
-      regex: lightbend;prometheus;alertmanager;[0-9]+
-      action: keep
+  - static_configs:
```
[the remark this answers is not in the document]
mitkus: `kubernetes_sd_configs: role: pod` lists all pods and relabel config is used as a filter to find the alertmanager pod (`action: keep`). I think this has nothing to do with alertmanager metrics.

enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -13,12 +13,9 @@ rule_files:
...
+    - targets:
+      - {{ .AlertmanagerAddress }}
```
jsravn: I wonder if we need to ever support multiple alert managers. How does HA alertmanager work? Do you need to provide the address of each alertmanager? In which case, we probably need to support passing in multiple values.

enterprise-suite/templates/prometheus.yaml
```diff
@@ -139,6 +139,11 @@ spec:
             - --defaultMonitorsFile=/etc/es-monitor-api/default-monitors.json
             - --prometheusTemplate=/etc/es-monitor-api/prometheus.yml
             - --prometheusDomain={{ .Values.prometheusDomain }}
+            {{- if .Values.alertmanagerAddress}}
+            - --alertmanagerAddress={{ .Values.alertManagerAddress }}:{{ .Values.alertManagerPort }}
+            {{ else }}
+            - --alertmanagerAddress=alertmanager:{{ .Values.alertManagerPort }}
+            {{ end }}
```
jsravn: Wouldn't it be better to have a flag "createAlertManager" that we can use as the gate for creating the internal alertmanager instead?

Then you can just set `alertManagerAddress` to the default value `alertmanager`, and we can get rid of the conditional logic that depends on whether the address is nil or not.

I think this is also a bit more explicit/clearer to users. It's not as obvious that setting the address will cause the internal alertmanager to not be created.

enterprise-suite/es-monitor-api/prometheus.yml
mitkus: Yes, HA alertmanagers work by providing multiple addresses to prometheus.

enterprise-suite/templates/alertmanager.yaml
```diff
@@ -58,16 +59,16 @@ spec:
             - --cluster.listen-address=$(POD_IP):6783
 
           ports:
-            - containerPort: 9093
+            - containerPort: {{ .Values.alertManagerPort }}
           readinessProbe:
             httpGet:
               path: /-/ready
-              port: 9093
+              port: {{ .Values.alertManagerPort }}
```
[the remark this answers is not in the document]
mitkus: I think probe is just another HTTP endpoint accessible on the same port as other alertmanager stuff. So ... probably no need for different ports?

enterprise-suite/templates/alertmanager.yaml
```diff
@@ -146,3 +147,4 @@ metadata:
   name: alertmanager-default
 data:
 {{ (.Files.Glob "alertmanager/*").AsConfig | indent 2}}
+{{ end }}
```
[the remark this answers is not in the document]
mitkus: Done

enterprise-suite/templates/prometheus.yaml
mitkus: Changed it to `createAlertManager` + `alertManagers` for addresses.

enterprise-suite/templates/prometheus.yaml
```diff
@@ -1,4 +1,21 @@
...
+# {{ tuple .Values.test "9000" | include "serviceList" }}
+#
+# Evaluates to:
+# service-00:9000,service-01:9000
+{{- define "serviceList" -}}
+{{- $suffix := index . 1 -}}
+{{- $local := dict "first" true -}}
+{{- range $k, $v := index . 0 -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}:{{- $suffix -}}{{- $_ := set $local "first" false -}}{{- end -}}
+{{- end -}}
+
```
jsravn: Can you put this into `_helpers.tpl`, that's the helm convention.

enterprise-suite/values.yaml
```diff
@@ -75,6 +75,16 @@ imageCredentials:
...
 
+# If true, an alertmanager will be deployed. Set to false and modify `alertManagers` list to use existing alertmanager(s).
+createAlertManager: true
+
+# List of alertmanager addresses (can be k8s local DNS service names)
+alertManagers: 
+  - alertmanager
+
+# Port to use for accessing alertmanager services
+alertManagerPort: 9093
```
jsravn: How come the port is separate from the hostname? Wouldn't it be simpler to include the port on the address?
```
alergManagers
- alertmanager:9093
- alertmanager2:9093
- alertmanager3:4000 # external
```

enterprise-suite/templates/alertmanager.yaml
```diff
@@ -111,9 +112,9 @@ metadata:
   name: expose-alertmanager
```
jsravn: I think we could just delete this service. We shouldn't need these expose services anymore.

enterprise-suite/values.yaml
jsravn: Okay, I see we do this to make the implementation simpler (like the configmap reloader stuff). But I think this should be optimized for users, so it should contain the port on the list of alertmanagers and not have it be a separate var.

For the alertmanager we create, I think it's fine to hardcode 9093 on the config reloader since we tightly control it and know what port it's listening on.

enterprise-suite/values.yaml
```diff
@@ -75,6 +75,16 @@ imageCredentials:
 #  username: setme
 #  password: setme
 
+# If true, an alertmanager will be deployed. Set to false and modify `alertManagers` list to use existing alertmanager(s).
+createAlertManager: true
+
+# List of alertmanager addresses (can be k8s local DNS service names)
+alertManagers: 
+  - alertmanager
```
jsravn: I'm not sure this should be a list. Doesn't that make it awkward to specify on the command line? I think a single var that is comma separated would be fine. Since this gets passed directly to the es-monitor-api as a command line option.

enterprise-suite/templates/es-console.yaml
```diff
@@ -150,7 +151,7 @@ data:
       }
 
       location ~ (.*/service/alertmanager)(/.*) {
-        proxy_pass http://$alertmanager:9093$2$is_args$args;
+        proxy_pass http://$alertmanager:$alertmanagerport$2$is_args$args;
         proxy_redirect '/' ' $1/';
       }
```
jsravn: Does this make sense if the user has an external alertmanager? We could probably disable this section in that case (it's up to the user to manage access to their external AM).

mitkus: It still works. In case of external alertmanager, we need to know its address anyway, so we can setup forwarding here. That is used at least for alertmanager UI button in Console.

The only quirk I can think of is that in case of HA alertmanager cluster, this would always forward to the first alertmanager.

enterprise-suite/templates/alertmanager.yaml
mitkus: Yup, done this in a separate PR: https://github.com/lightbend/helm-charts/pull/197

enterprise-suite/values.yaml
mitkus: Making it a list has advantage of easier manipulation in template logic. For example, I need to get first item for setting up nginx forwarding. Also, I feel it's more idomatic since YAML has list data structure? That makes it nicer if you want to specify values in a YAML file instead of passing everything as commandline arguments.

Commandline looks like this now:
`--set alertManagers={alertmanager-00,alertmanager-01,alertmanager-02}`
vs. comma separated list:
`--set alertManagers=alertmanager-00,alertmanager-01,alertmanager-02`

enterprise-suite/templates/prometheus.yaml
mitkus: Done

enterprise-suite/values.yaml
mitkus: Good point, I changed `alertManagers` to include ports in each address.

mitkus:
Now tested with HA alertmanager cluster! K8s yaml for setting it up: https://gist.github.com/mitkus/bc4484115f584fbac9c6e3220c23ef4f. Very copy-paste heavy since they need unique service names and I don't know how to do it better.

Launching Console that will use this cluster:
`
scripts/lbc.py install --local-chart=. --namespace=monitoring --set exposeServices=NodePort --set "alertManagers={alertmanager-00:9093,alertmanager-01:9093,alertmanager-02:9093}" --set createAlertManager=false --set esMonitorVersion=latest
`

Alert mgr button in Console UI now links to first alertmanager in the cluster.

enterprise-suite/es-monitor-api/prometheus.yml
mitkus: Now works with HA alertmanager clusters

PR #196, opened 2018-11-20
Prefix ClusterRole with namespace of Console install

jsravn:
This allows multiple Console installs without any extra work from users.
Prior to that they would have had to set `CreateClusterRoles=false`.


mitkus:
Should we also remove `createClusterRoles` value with this change?

jsravn:
@mitkus I thought about it. But think it may be useful for customers that want to restrict the ClusterRole to specific namespaces, that sort of thing. Or is this YAGNI?

jsravn:
I'll remove it in another PR.

PR #197, opened 2018-11-21
Remove expose-* services for everything but console

mitkus:
For https://github.com/lightbend/es-backend/issues/410

I understand these should be accessed through console now, so this PR removes `expose-prometheus`, `expose-grafana`, `expose-alertmanager` services.

### DIFF
--- a/enterprise-suite/templates/alertmanager.yaml
+++ b/enterprise-suite/templates/alertmanager.yaml
@@ -103,24 +103,6 @@ spec:
           emptyDir: {}
           {{ end }}
 
-{{ if or .Values.exposeServices .Values.minikube }}
----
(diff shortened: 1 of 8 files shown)


dbrinegar:
I think I'm in favor of removing unnecessary options like this (not typically needed, easy for user to add their own such resource, nothing special about integrating in our helm-chart) as it reduces the number of permutations to support.

I think you'll need to fiddle the tests to use nginx to get to things (the console expose) as they're hitting these other things directly. So prometheus is on the console url under `/service/prometheus` etc

jsravn:
Please also announce in #applied-console, since there are probably still a few people using the nodeports for everything.

PR #201, opened 2018-11-27
do not merge: test rc

dbrinegar:
This is refactor-api plus some rc images, to see if Travis is green.

### DIFF
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -2,9 +2,9 @@
 # Update these versions for a release.
 #
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-console
-esConsoleVersion: v0.1.7
+esConsoleVersion: rc
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-monitor-api
-esMonitorVersion: v0.2.1
+esMonitorVersion: rc
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-grafana
 esGrafanaVersion: v0.1.4
 # prom/prometheus


enterprise-suite/tests/smoke_prometheus
```diff
@@ -9,9 +9,20 @@ source smokecommon
...
 done
+
+cleanup() {
+  for app in resources/app*.yaml; do
+    kubectl delete -f "$app" -n $NAMESPACE
+  done
+  kubectl delete namespace $NAMESPACE
+}
+
+trap cleanup 0
```
jsravn: Is there a reason to create it in a separate namespace to where we install the console? This is okay on minikube, but on openshift or more restricted environments that control namespace creation it could be problematic. I think it's better to keep the apps inside of the console namespace.

enterprise-suite/tests/smoke_prometheus
```diff
@@ -237,17 +248,40 @@ busy_wait app_data
...
+            "threshold": "1"
+          }
+        },
+        "summary": "summ",
+        "description": "desc"
+      }
+    }
+EOF
+}
+
```
jsravn: Nice!

enterprise-suite/values.yaml
```diff
@@ -2,9 +2,9 @@
...
 #
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-console
-esConsoleVersion: v0.1.7
+esConsoleVersion: rc
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-monitor-api
-esMonitorVersion: v0.2.1
+esMonitorVersion: rc
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-grafana
 esGrafanaVersion: v0.1.4
 # prom/prometheus
```
jsravn: I'm guessing we'll change these versions before the PR is merged, right?

enterprise-suite/tests/smoke_prometheus
dbrinegar: Ah good point. I was hoping to "delete namespace" and clean everything up after a test, but deleting things turned up yet new edge cases so still pondering how to best clean up the test.

enterprise-suite/values.yaml
dbrinegar: this is just temporary staging, to be thrown away once we merge to master

PR #202, opened 2018-11-28
Fix minishift error message

srikanthops:
lightbend/es-backend/issues/464

noticed, we don't handle missing command, so added catch exception for OSError 
https://docs.python.org/2/library/subprocess.html#exceptions

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -107,7 +107,7 @@ def run(cmd, timeout=None, stdin=None, show_stderr=True):
 # prints it to stdout or stderr, handles failure status
 # codes by exiting with and error if can_fail=False.
 def execute(cmd, can_fail=False, print_to_stdout=False):
-    printerr(cmr)
+    printerr(cmd)
     if not args.dry_run:
         stdout, returncode = run(cmd)
         if print_to_stdout:
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -92,6 +92,12 @@ def run(cmd, timeout=None, stdin=None, show_stderr=True):
         if len(stderr) > 0 and show_stderr:
             printerr(stderr)
         returncode = proc.returncode
+    except OSError, o:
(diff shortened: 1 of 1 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -92,6 +92,12 @@ def run(cmd, timeout=None, stdin=None, show_stderr=True):
         if len(stderr) > 0 and show_stderr:
             printerr(stderr)
         returncode = proc.returncode
+    except OSError, o:
+        stdout=o.strerror
+        returncode=o.errno
+    except Exception, e:
+        stdout=str(e)
+        returncode=1
```
jsravn: I don't think we should catch the exception here every time. For example if we are trying to run the `helm` command, then it should fail with an exception.

Instead in is_running_minishift, we should catch OSError in case it doesn't exist. WDYT?

srikanthops: We have a return in the finally block, so the exceptions are never raised, 
I choose this way, to catch the exception and return a non-zero error code, and leave it for the caller to handle it.. 
Note: Lot us didn't have minishift command installed, and there was no exception (errno 2) for that.

enterprise-suite/scripts/lbc.py
```diff
@@ -101,7 +107,7 @@ def run(cmd, timeout=None, stdin=None, show_stderr=True):
 # prints it to stdout or stderr, handles failure status
 # codes by exiting with and error if can_fail=False.
 def execute(cmd, can_fail=False, print_to_stdout=False):
-    printerr(cmd)
+    printinfo(cmd)
```
mitkus: This should be `printerr`, so it goes to `stderr`. Idea is to be able to pipe output to a useful file `lbc.py install --export-yaml=console > console.yaml`

jsravn: That seems like a bug, why did you approve the PR? :)

srikanthops: ok got it..

PR #203, opened 2018-11-28
Fix 'Service' service discovery to use service/endpoint metadata only

jsravn:
This was relying on there being pods under the hood to work, which is
wrong. It should only use the service and endpoint values.

Fix https://github.com/lightbend/es-backend/issues/478


dbrinegar:
This was all changed initially due to those odd dns pod endpoints in minikube? Do we have a regression test for that issue, or have you verified that those are not coming back?

jsravn:
I don't think I added a test case specifically for that, but have added many test cases for the service discovery itself since then so I have reasonable confidence. And I added a testcase here for the endpoint usecase specifically. It could be much better though - we are not testing many of the variations, especially around annotations.

jsravn:
Possibly a test case checking for failed ups (although it would fail now, because of this change)...

jsravn:
Ah completely forgot - already added a testcase for it in #167.

PR #206, opened 2018-11-29
Fix alertmanager examples

jsravn:
- Group on severity as well, so we don't mix and duplicate
warning/critical alerts in the same groups.
- Inhibit correctly on namespace and workload, so we don't inhibit
alerts in other workloads!

### DIFF
--- a/enterprise-suite/alertmanager/alertmanager.yml
+++ b/enterprise-suite/alertmanager/alertmanager.yml
@@ -13,7 +13,7 @@ route:
   # Wait before sending a new alert for a group.
   group_interval: 5m
   # Group alerts by namespace and workload.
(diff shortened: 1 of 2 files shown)


enterprise-suite/alertmanager/alertmanager.yml
```diff
@@ -13,7 +13,7 @@ route:
   # Wait before sending a new alert for a group.
   group_interval: 5m
   # Group alerts by namespace and workload.
```
marcoderama: Might want to update the comment too, although it seems kinda redundant anyway.

jsravn: Indeed! Don't want to burden a PR just for this, so I'll fix it next time.

PR #207, opened 2018-11-29
Tweak smoke_ingress-tls in the same was as smoke_ingress (102e601) to red…

marcoderama:
…uce/eliminate transient failures.

Nothing fancy here.  This stunt seems to have worked to reduce the flakiness of the `smoke_ingress` test.


marcoderama:
I'll take a half-frown.  Better than last time...  :-)

PR #212, opened 2018-12-05
Various lbc.py improvements/fixes

mitkus:
For https://github.com/lightbend/es-backend/issues/476, https://github.com/lightbend/es-backend/issues/473, https://github.com/lightbend/es-backend/issues/474, https://github.com/lightbend/es-backend/issues/472

This PR makes following changes to `lbc.py`:
- Better `--set` value parsing with helm syntax compatibility
- Include PV info in debug dumps
- Try to reuse clusterrole and PV resources by default
- Option to skip checking alertmanager when doing `verify` for use with external alertmanagers
- Always verify install when `--wait` flag is set

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -291,7 +291,7 @@ def are_cluster_roles_created():
 # Takes helm style "key1=value1,key2=value2" string and returns a list of (key, value)
 # pairs. Supports quoting, escaped or non-escaped commas and values with commas inside, eg.:
 #  parse_set_string('am=amg01:9093,amg02:9093') -> [('am', 'amg01:9093,amg02:9093')]
-#  parse_set_string('am="am01,am02",es=NodePort') -> [('alertManagers', 'am01,am02'), ('es', 'NodePort')]
+#  parse_set_string('am="am01,am02",es=NodePort') -> [('am', 'am01,am02'), ('es', 'NodePort')]
 def parse_set_string(s):
     # Keyval pair with commas allowed
     keyval_pair_re = re.compile(r'(\w+)=([\w\-\+\*\:\\,]+)')
@@ -327,7 +327,7 @@ def install(creds_file):
     version_arg = ('--version ' + args.version) if args.version != None else '--devel'
 
     # Helm args are separated from lbc.py args by double dash, filter it out
-    helm_args = ' '.join([arg for arg in args.helm if arg != '--'])
+    helm_args = ' '.join([arg for arg in args.helm if arg != '--']) + ' '
 
     # Add '--set' arguments to helm_args
     if args.set != None:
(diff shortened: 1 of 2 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -288,16 +290,54 @@ def are_cluster_roles_created():
         fail_msg='Found some cluster roles from previous console install, but not all: {}. Please clean them up manually.'
     )
 
+# Takes helm style "key1=value1,key2=value2" string and returns a list of (key, value)
```
srikanthops: markus, did you consider passing command line arg as is to helm.....

marcoderama: It already has that feature.  Anything after an arg of `--` gets passed on to helm.  (You might find the discussion in https://github.com/lightbend/es-backend/issues/474 of interest.)

enterprise-suite/scripts/lbc.py
```diff
@@ -288,16 +290,54 @@ def are_cluster_roles_created():
...
+            mq = keyval_pair_quot_re.match(left)
+            mn = keyval_pair_nc_re.match(left)
+            m = mq or mn
+            if m != None:
+                matchlen = len(m.group(0))
+                result.append((m.group(1), m.group(2).replace('\\,', ',')))
+                if matchlen == len(left) or left[matchlen] == ',':
+                    left = left[matchlen+1:]
+                else:
+                    raise ValueError('unexpected characted "{}"'.format(left[matchlen]))
```
marcoderama: unexpected characteR

enterprise-suite/scripts/lbc.py
```diff
@@ -558,8 +611,12 @@ def setup_args(argv):
                         action='store_true')
     install.add_argument('--export-yaml', help='export resource yaml to stdout',
                         choices=['creds', 'console'])
-    install.add_argument('--reuse-resources', help='try to reuse PVCs and/or cluster roles from a previous install',
-                        action='store_true')
+
+    install.add_argument('--reuse-resources', dest='reuse_resources',
+                         help='try to reuse PVCs and/or cluster roles from a previous install', action='store_true')
+    install.add_argument('--no-reuse-resources', dest='reuse_resources', help=argparse.SUPPRESS, action='store_false')
```
marcoderama: Question:  How does `help=argparse.SUPPRESS` work?

mitkus: It doesn't print a help line for that argument. We have `--reuse-resources` documented and now it says that default is `true`. To turn such flag off, standard is to have the same argument with prefix `no`.

enterprise-suite/scripts/lbc.py
mitkus: Fixed

enterprise-suite/scripts/lbc.py
marcoderama: Got it.  Tnx.

PR #213, opened 2018-12-07
Prometheus- and alertmanager-generated links work

marcoderama:
If you provide a value for the new `esConsoleURL` variable in `values.yaml`, it will be used by prometheus and alertmanager when they generate links back to themselves or each other.  This would be the case
* in an alerts page generated by alertmanager that includes a link to prometheus
* in alert emails/pages/slacks that link to alertmanager


marcoderama:
I've tested this in my dev env (cf. https://github.com/lightbend/es-backend/issues/435#issuecomment-445368376), but I think automated testing will be non-trivial.  We'd have to:
* fire up a proxy and/or configure things to expose the console (to a known address)
* tweak things to generate alerts
* scrape an alertmanager alert page to check the links (actually tests prometheus change)
* scrape an actual alert somehow (would test alertmanager change)

I think that's worthy of a couple new issues.  Will create.

jsravn:
@marcoderama you could just verify the correct URLs are passed to alertmanager / prom, right?

marcoderama:
I guess so...  Do you mean in the configs or by querying them somehow?  Btw, it wasn't enough to just use `--web.external-url`.  I had to explicitly specify `--web.route-prefix=/` even though that's the supposed default.

jsravn:
@marcoderama yeah I agree with that. Same with testing helm stuff in general. We should wait for https://github.com/lightbend/es-backend/issues/504 as a first step.

jsravn:
@marcoderama yeah I agree with that. Same with testing helm stuff in general. We should wait for https://github.com/lightbend/es-backend/issues/504 as a first step.

PR #218, opened 2018-12-11
Go testing foundations

mitkus:
This is an initial look at how e2e tests in go would work. Uses new go modules, so it doesn't need to live in $GOPATH! That means it requires go 1.11 though.

It still exhibits flakyness, but overtime it could be easier to handle than in the bash tests. View this as a foundation for all of us to work on.

Tests start assuming kube cluster is running & helm is installed. It can do it for you too with `--start-minikube` flag - I find that convenient for a clean run on my local machine.


dbrinegar:
yay!  I wonder about driving minikube

1. I think Travis may be doing its own minikube management ahead of the build
2. we want to aim at other kubernetes clusters

and the install itself as we may want some variation control (install with this script and that option) ahead of the same set of tests.

Would it be reasonable to have the tests start with some Kubernetes and Console context, already installed? Require that be set up first?

jsravn:
I agree w/ @dbrinegar. The boundary should assume helm and kubectl are already set up. That way we can run the tests on many combinations of setups (openshift, GKE, minikube) without the tests having to worry about it.

jsravn:
Started reviewing this... looks really great. Can you a quick pointer on how to setup your environment to run the tests @mitkus?

jsravn:
This happened when I tried to run it locally:

```
$ ginkgo -- --start-minikube                                                           (console-test)
go: downloading k8s.io/apimachinery v0.0.0-20181215012845-4d029f033399
go: downloading k8s.io/client-go v10.0.0+incompatible
go: downloading k8s.io/api v0.0.0-20181130031204-d04500c8c3dd
go: downloading github.com/onsi/gomega v1.4.3
go: downloading golang.org/x/net v0.0.0-20180906233101-161cd47e91fd
go: downloading gopkg.in/yaml.v2 v2.2.1
go: downloading golang.org/x/text v0.3.0
go: downloading github.com/google/gofuzz v0.0.0-20170612174753-24818f796faf
go: downloading k8s.io/klog v0.1.0
go: downloading github.com/gogo/protobuf v1.2.0
go: downloading github.com/googleapis/gnostic v0.2.0
go: downloading github.com/gregjones/httpcache v0.0.0-20181110185634-c63ab54fda8f
go: downloading github.com/modern-go/reflect2 v1.0.1
go: downloading golang.org/x/crypto v0.0.0-20181203042331-505ab145d0a9
go: downloading github.com/peterbourgon/diskv v2.0.1+incompatible
go: downloading sigs.k8s.io/yaml v1.1.0
go: downloading github.com/spf13/pflag v1.0.3
go: downloading github.com/imdario/mergo v0.3.6
go: downloading github.com/json-iterator/go v1.1.5
go: downloading github.com/golang/protobuf v1.2.0
go: downloading golang.org/x/oauth2 v0.0.0-20181203162652-d668ce993890
go: downloading github.com/google/btree v0.0.0-20180813153112-4030bb1f1f0c
go: downloading golang.org/x/time v0.0.0-20181108054448-85acf8d2951c
go: downloading gopkg.in/inf.v0 v0.9.1
go: downloading github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd

Running Suite: Console Suite
============================
Random Seed: 1547228491
Will run 4 of 5 specs

Failure [0.593 seconds]
[BeforeSuite] BeforeSuite 
/home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:27

  Expected
      <bool>: true
  not to be true

  /home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:30
------------------------------
Unable to delete console installation - no release named enterprise-suite found
Failure [0.283 seconds]
[AfterSuite] AfterSuite 
/home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:65

  Expected success, but got an error:
      <*errors.errorString | 0xc0003b0030>: {
          s: "'../scripts/lbc.py uninstall' return status 1",
      }
      '../scripts/lbc.py uninstall' return status 1

  /home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:66
------------------------------

Ran 4 of 0 Specs in 0.877 seconds
FAIL! -- 0 Passed | 4 Failed | 0 Pending | 0 Skipped
--- FAIL: TestConsole (0.88s)
FAIL

Ginkgo ran 1 suite in 15.42882668s
Test Suite Failed
```

mitkus:
@jsravn It seems you already had minikube running. Try running without `--start-minikube` in such case.

jsravn:
Okay, I still got a failure:

```
------------------------------
• Failure [1.701 seconds]
all:verify
/home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:81
  Console
  /home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:88
    is verified [It]
    /home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:89

    Expected success, but got an error:
        <*errors.errorString | 0xc0002088f0>: {
            s: "'../scripts/lbc.py verify --namespace lightbend-test' return status 1",
        }
        '../scripts/lbc.py verify --namespace lightbend-test' return status 1

    /home/james/lightbend/console-charts/enterprise-suite/console_test/console_test.go:90
```

With 3 passes.

In this case - when the verify step fails - can we make sure we are printing it to stderr so ginkgo will log the output? Otherwise it's hard to know why it failed.

jsravn:
^ Note I think it fails when console is already installed.

jsravn:
Seems some flakiness - got this now:

```
P [PENDING]
minikube:ingress
/home/james/lightbend/console-charts/enterprise-suite/console_test/smoke_ingress_test.go:28
  responds to requests
  /home/james/lightbend/console-charts/enterprise-suite/console_test/smoke_ingress_test.go:29
------------------------------
••
------------------------------
• Failure [18.025 seconds]
all:prometheus
/home/james/lightbend/console-charts/enterprise-suite/console_test/prometheus_test.go:20
  works [It]
  /home/james/lightbend/console-charts/enterprise-suite/console_test/prometheus_test.go:94

  Expected
      <bool>: false
  to be true

  /home/james/lightbend/console-charts/enterprise-suite/console_test/prometheus_test.go:121
------------------------------
•

info: deleting previous console installation enterprise-suite with status 'deployed'
helm delete --purge enterprise-suite
release "enterprise-suite" deleted

warning: helm delete does not wait for resources to be removed- if the script fails on install, please re-run it.
Summarizing 1 Failure:

[Fail] all:prometheus [It] works 
/home/james/lightbend/console-charts/enterprise-suite/console_test/prometheus_test.go:121

Ran 4 of 5 Specs in 47.524 seconds
FAIL! -- 3 Passed | 1 Failed | 1 Pending | 0 Skipped
--- FAIL: TestConsole (47.52s)
FAIL
```

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+const DefaultTimeout = 3 * time.Second
+
+type CmdBuilder struct {
+	name                 string
+	args                 []string
+	timeout              time.Duration
+	expectZeroExitStatus bool
+	panicOnError         bool
+	printStdout          bool
+	printStderr          bool
```
jsravn: Shouldn't we default these to `true`? As I believe ginkgo will hide stdout/stderr unless the test fails, so no harm in logging all output.

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+
+	var exitcode int
+	var cmderr error
+
+	// Setup stdout/sterr pipes
+	cmdstderr, err := cmd.StderrPipe()
+	if err != nil {
+		exitcode, cmderr = -100, err
+	}
+	cmdstdout, err := cmd.StdoutPipe()
```
jsravn: `StdoutPipe`/`StderrPipe` docs say it's an error to call `Wait` before consuming the pipe. But it looks like we only consume these when we have nonempty `stdoutWriters`/`stderrWriters`. So we shouldn't create these unless we are going to consume them (or always consume below - I think it will be a noop if you do `io.MultiWriter(nil)`).

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+			if cb.printStdout {
+				stdoutWriters = append(stdoutWriters, os.Stdout)
+			}
+			if cb.printStderr {
+				stderrWriters = append(stderrWriters, os.Stderr)
+			}
+
+			// Hook up command stdout/stderr to potential destinations
+			if len(stdoutWriters) > 0 {
+				io.Copy(io.MultiWriter(stdoutWriters...), cmdstdout)
```
jsravn: Check `error` result.

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+			if len(stderrWriters) > 0 {
+				io.Copy(io.MultiWriter(stderrWriters...), cmdstderr)
```
jsravn: Check `error` result.

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+const maxRepeats = 100
+const failureSleep = 500 * time.Millisecond
+
+// Repeatedly runs a function, sleeping for a bit after each time, until it returns true or reaches maxRepeats
+func WaitUntilTrue(f func() bool) error {
+	for i := 0; i < maxRepeats; i++ {
+		if f() == true {
+			return nil
+		}
+		time.Sleep(failureSleep)
```
jsravn: Exponential back off with a smaller starting wait time would be a bit nicer for the general case.

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+	"time"
+)
+
+const DefaultTimeout = 3 * time.Second
+
+type CmdBuilder struct {
```
jsravn: I like this a lot - makes the tests nice and readable. :+1:

enterprise-suite/console_test/README
```diff
@@ -0,0 +1,12 @@
...
+Depends on having ginkgo installed:
+> go get github.com/onsi/ginkgo/ginkgo
+
+Running tests locally with no minikube or helm running:
+> ginkgo -- --start-minikube
+
+Running on openshift:
+> ginkgo --skip=*minikube*
+
+Test names should include one of prefixes "all:", "minikube:", "openshift:" to describe valid platforms.
```
jsravn: Maybe show an example here, so it's a bit clearer what you mean.

enterprise-suite/console_test/console_test.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// The following variables are used in this package to access kubernetes
+// and make requests on Console components.
+var k8sClient *kubernetes.Clientset
+var consoleAddr string
+var prometheusAddr string
+var monitorAPIAddr string
+
+var _ = BeforeSuite(func() {
```
jsravn: I think instead of having a global BeforeSuite, it's better to extract this into a function, and call it from the BeforeSuite of the various `Describe`. That lets us customize as needed in those tests. It's also a little more explicit (it's not obvious the beforesuite is executing from another file).

enterprise-suite/console_test/prometheus_test.go
```diff
@@ -0,0 +1,264 @@
...
+		// Specific test for regression of es-backend/issues/430.
+		Expect(prom.HasData("{job=\"kubernetes-cadvisor\",es_monitor_type=\"es-test\"}")).To(BeTrue())
+
+		// Succeeds if all data for the workload es-test  has a matching es_monitor_type
+		// Note we're currently ignoring health metrics because 'bad' data can stick around for 15m given their time window.
+		err = util.WaitUntilTrue(func() bool {
+			return prom.HasData("{es_workload=\"es-test\", es_monitor_type!=\"es-test\", __name__!=\"health\"}")
+		})
+		Expect(err).To(Succeed())
+	})
```
jsravn: It would be a lot nicer if this was broken up into a few asserts per "It" block. It's much harder to see what we're testing with a giant block like this. It also makes the test failures/messages less obvious, since "it works" "failed" isn't very specific. I know it's just copied from what we have, but a little organization will go a long way in the future I think. (e.g. have a separate "it" or "describe" block for testing Pod service discovery, one for Service, and so on).

jsravn: There are also a lot of opportunities for using data tables here - see https://godoc.org/github.com/onsi/ginkgo/extensions/table. That will reduce the duplication - we can create a list of prometheus queries and their expected outcome for example.

enterprise-suite/console_test/prometheus_test.go
```diff
@@ -0,0 +1,264 @@
...
+			return prom.HasData(appInstancesQuery)
+		})
+		Expect(err).To(Succeed())
+
+		// Pod discovery - automatic metrics should gain an es_monitor_type label when a custom monitor is created
+		Expect(esMonitor.Make("es-test/my_custom_monitor", "up")).To(Succeed())
+		err = util.WaitUntilTrue(func() bool {
+			return prom.HasModel("my_custom_monitor")
+		})
+		Expect(err).To(Succeed())
```
jsravn: Is there a way to make nicer error messages when these tests fail? It would be nice to see the query for example, and what it returned (or didn't return). Maybe we need another utility method which returns a more detailed error. I think it will be quite difficult to figure out why these failed on a Travis build without that detail.

enterprise-suite/console_test/smoke_ingress_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Console NodePort service
+const consoleServiceName = "expose-es-console"
+
+// Name of Ingress that this test creates
+const testIngressName = "console-test-ingress"
+
+// TODO(mitkus): Make this test work. On my machine both this and bash e2e smoke_ingress
+// stopped working after minikube update.
+var _ = XDescribe("minikube:ingress", func() {
```
jsravn: I guess this is WIP then?

enterprise-suite/console_test/console_test.go
```diff
@@ -0,0 +1,98 @@
...
+		panic(err.Error())
+	}
+	k8sClient, err = kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
+
+	// If we started minikube, install helm as well
+	if args.StartMinikube {
+		Expect(helm.Install(k8sClient)).To(Succeed())
```
jsravn: Maybe installing helm should be a separate flag? Although I'm not sure we should be doing this in the tests. I wonder if we should just rely on helm being setup already. :conflicted:

enterprise-suite/console_test/prometheus_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/lightbend/console_test/util/prometheus"
+
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const promTestNamespace = "smoketest-prometheus"
```
jsravn: This can't be hardcoded - because it will vary based on environment we're deploying, such as with openshift.

I'm thinking we shouldn't specify it at all. Instead we should rely on the kubectl context being set-up to point to the correct namespace (which it does currently in travis).

enterprise-suite/console_test/util/helm/helm.go
```diff
@@ -0,0 +1,65 @@
...
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/lightbend/console_test/util"
+)
+
+const ServiceAccountName = "tiller"
+const TillerNamespace = "kube-system"
```
jsravn: I think this should be the same namespace as wherever we run the tests. E.g. we don't want to install tiller in kube-system on a shared cluster.

Again though - we could also just rip this out and let the upper layer scripts sort out helm, since it largely depends on environment.

enterprise-suite/console_test/util/util.go
mitkus: Done

enterprise-suite/console_test/util/util.go
mitkus: Right, restructured this part.

enterprise-suite/console_test/util/util.go
mitkus: Done

enterprise-suite/console_test/util/util.go
mitkus: I tried it out, ginkgo prints out stdout/stderr by default on my machine. Maybe it's only in BeforeSuite/AfterSuite? Anyway, it gets quite noisy, I would vote for keeping the default `false`.

enterprise-suite/console_test/README
```diff
@@ -0,0 +1,12 @@
+Lightbend Console e2e tests
```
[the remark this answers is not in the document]
mitkus: Done

enterprise-suite/console_test/util/util.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// Setup stdout/sterr pipes
+	cmdstderr, err := cmd.StderrPipe()
+	if err != nil {
+		exitcode, cmderr = -100, err
+	}
+	cmdstdout, err := cmd.StdoutPipe()
+	if err != nil {
+		exitcode, cmderr = -200, err
+	}
```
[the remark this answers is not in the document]
mitkus: Yup, changed most non-recoverable errors to panics in setup/teardown stages.

enterprise-suite/console_test/util/util.go
mitkus: Implemented exponential backoff here

enterprise-suite/console_test/console_test.go
mitkus: Right, this whole restructuring into packages needed that as well. Now environment initialization is in `testenv` package.

enterprise-suite/console_test/prometheus_test.go
mitkus: Should be better now!

enterprise-suite/console_test/prometheus_test.go
mitkus: Improved it a bit with additional message param in `WaitUntilTrue`

enterprise-suite/console_test/smoke_ingress_test.go
mitkus: I'll tackle this next after merging what's currently working

enterprise-suite/console_test/console_test.go
mitkus: Kept this all under `--start-minikube` for now, but I'm open to changes. I find some value in being able to fire and forget a test suite without any manual env setup, so I like that `console_test` can start minikube & helm by itself. Other workflows may wary of course!

enterprise-suite/console_test/prometheus_test.go
mitkus: Removed this, now all testing happens in `lightbend-test` which is also overridable with `--namespace` flag.

enterprise-suite/console_test/util/helm/helm.go
mitkus: Reverted to using namespace provided by commandline flag with `lightbend-test` default.

enterprise-suite/console_test/tests/prometheus/prometheus_test.go
```diff
@@ -0,0 +1,251 @@
...
+			}, "some data of es-test workload doesn't have matching es_monitor_type")
+			Expect(err).To(Succeed())
+		})
+	})
+})
+
+func TestPrometheus(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Prometheus Suite")
+}
```
marcoderama: Newline on end please.

enterprise-suite/console_test/util/monitor/monitor.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// Connection to es-monitor-api server
+type Connection struct {
+	url string
+}
+
+// Make adds a simple threshold metric with non-configurable parameters.
+// TODO: Make this more configurable!
+func (m *Connection) Make(name string, metric string) error {
```
marcoderama: The function name confused me.  Would it make sense to be "MakeMonitor" or something like that?

mitkus: Good suggestion, renamed to MakeMonitor

enterprise-suite/console_test/tests/prometheus/prometheus_test.go
mitkus: Fixed

enterprise-suite/console_test/resources/app_service_with_only_endpoints.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: es-test-service-with-only-endpoints
+  annotations:
+    "prometheus.io/scrape": "true"
```
jsravn: Do you mind changing these to match the current tests? I think it's `console-backend-e2e.io/scrape`. This is so we can run the tests in a shared kubernetes cluster (e.g. the openshift tests).

enterprise-suite/console_test/resources/app_with_multiple_ports.yaml
```diff
@@ -0,0 +1,32 @@
...
+        # source: https://github.com/lightbend/k8s-explore/tree/master/query/nan
+        - name: es-test-1
+          image: lightbend-docker-registry.bintray.io/enterprise-suite/es-test:v1
+          imagePullPolicy: IfNotPresent
+          ports:
+            - containerPort: 8080
+              name: metrics
+
+        - name: es-test-2
+          image: lightbend-docker-registry.bintray.io/enterprise-suite/es-test:v1
```
jsravn: Wish I knew where es-test came from... :thinking:

enterprise-suite/console_test/setup-minikube.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+minikube start --cpus=3 --memory=6000
+kubectl create serviceaccount --namespace kube-system tiller
+kubectl create clusterrolebinding kube-system:tiller --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+helm init --wait --service-account tiller --tiller-namespace=kube-system
```
jsravn: eol

jsravn: Is this script used anywhere? I think just stick it in the Makefile if you want.

enterprise-suite/console_test/args/args.go
```diff
@@ -0,0 +1,20 @@
+package args
```
jsravn: I don't think we need a separate package for this one file. What about sticking it in the `tests` package?

enterprise-suite/console_test/testenv/testenv.go
```diff
@@ -0,0 +1,75 @@
+package testenv
```
jsravn: I don't think we need a separate package for this one file. What about sticking it in the `tests` package?

enterprise-suite/console_test/resources/app_with_multiple_ports.yaml
mitkus: A mystery for all of us!

enterprise-suite/console_test/setup-minikube.sh
mitkus: It's just something I started using once `install-helm` target disappeared. Yeah, I'll put it in the Makefile.

enterprise-suite/console_test/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Running only minikube or openshift:
+`ginkgo -r --skip=*minikube*`
+`ginkgo -r --skip=*openshift*`
+
+Running a single test suite:
+`ginkgo tests/prometheus`
+
+Test names should include one of prefixes `all:`, `minikube:`, `openshift:` to describe valid platforms.
+For example `all:prometheus` should be testing in all environment, while `minikube:ingress` only in minikube environment.
```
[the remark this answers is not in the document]
mitkus: I'm a noob with go 1.11 modules, but it seems you don't have to do anything to add a dependency. Simply write an import statement in the code and it works. I'll add something about updating.

enterprise-suite/console_test/README.md
```diff
@@ -0,0 +1,25 @@
...
+Running all tests locally:
+`ginkgo -r`
+
+Running only minikube or openshift:
+`ginkgo -r --skip=*minikube*`
+`ginkgo -r --skip=*openshift*`
+
+Running a single test suite:
+`ginkgo tests/prometheus`
+
```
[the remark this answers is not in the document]
mitkus: No way to --set helm values from the ginkgo commandline right now. My idea was that any tests that need custom values would set it up in `BeforeSuite` environment setup functions, possibly even using `lbc` module to do a custom install.

enterprise-suite/console_test/args/args.go
mitkus: I find it more readable to have commandline args in their own package, because any code that uses them can say `args.Namespace` to reference it. Can still change it if you feel strongly towards having less packages.

enterprise-suite/console_test/testenv/testenv.go
mitkus: If we keep `args` in a separate package, this new `tests` package would only have the current contents of `testenv`, so maybe better to keep the less general name?

mitkus:
Renamed folder to `go-test`, addressed most of the comments. I left `setup-minikube.sh` script for now, because helm install in the makefile uses `lightbend` namespace and that works with `lbc.py` only with `TILLER_NAMESPACE` env variable. Still pondering what to do about it...

enterprise-suite/console_test/resources/app_service_with_only_endpoints.yaml
mitkus: Updated to keep insync with bash tests

enterprise-suite/console_test/setup-minikube.sh
mitkus: On the other hand... Makefile sets up helm in `lightbend` namespace, which is slightly annoying for testing. Is there a reason why installing Tiller in the good old `kube-system` should be avoided in our test/development clusters?

enterprise-suite/console_test/README.md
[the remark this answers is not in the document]
jsravn: I imagine it's creating a lockfile still and so on.

enterprise-suite/console_test/args/args.go
jsravn: Okay I'll let you decide since I don't feel that strongly. It just felt a little out of place at the top level to me. But I think it's fine and easy to move if we change our minds.

PR #219, opened 2018-12-11
[wip] Some cleanup - remove makefile conditional logic

jsravn:
Better to use explicit targets than conditional.


enterprise-suite/Makefile
```diff
@@ -35,26 +34,20 @@ script-tests:
 	$(CHART_DIR)/tests/lib/bats/bats $(CHART_DIR)/tests
 	python $(CHART_DIR)/scripts/lbc_test.py
 
-minikube-tests: build install-local
-	@echo Installed Console from helm charts: $(CHART)
-	@echo Running tests
-	( cd $(CHART_DIR)/tests ; ./smoketest ; )
+minikube-backend-tests: build
+	@echo Running minikube tests
+	( cd $(CHART_DIR)/tests ; ./smoketest_minishift.sh ; )
```
mitkus: Shouldn't the file be named `smoketest_minikube.sh`?

enterprise-suite/templates/es-console-expose.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: expose-es-console
+  name: expose-console-server
```
mitkus: This needs to be updated in the docs too.

enterprise-suite/tests/smoketest_minishift.sh
```diff
@@ -0,0 +1,49 @@
...
+
+source ./smoketest_ci.sh
+
+function test_context() {
+    echo minikube
+}
+
+function diagnostics() {
+    # diagnostics
+    minishift status
```
mitkus: Should this be `minikube status`?

mitkus:
Oh, I got it. Travis runs one test suite on minikube, and another on openshift targeted at our Central Park cluster. All the references to minishift confused me :)

jsravn:
> All the references to minishift confused me :)

My mistake, I will fix all those.

jsravn:
Unfortunately I've gotten stuck on the frontend tests failing. Without the artifact bucket working, I can't see what's happening.

jsravn:
There also seems to be quite a lot of flakiness in smoke_prometheus when running on the openshift cluster. So I think there is still quite a bit of work here remaining.

enterprise-suite/Makefile
```diff
@@ -90,53 +83,13 @@ lint-allyaml: allyaml  ## Check that all.yaml is valid (Need Minikube running)
...
 allyaml: $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)/$(ALL_YAML)  ## Build all.yaml file
 
 $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)/$(ALL_YAML): $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz
 	$(call banner)
 	mkdir -p $(HELM_CHARTS_DIR)/docs/$(RELEASE_NAME)
 	helm --namespace=$(NAMESPACE) template $< $(EXPOSE_SERVICES) > $@
 
-TILLER_NAMESPACE := kube-system
-
-install-helm:  ## Install required helm components into cluster
```
marcoderama: I still use this target.  Why remove it?

enterprise-suite/Makefile
```diff
@@ -90,53 +83,13 @@ lint-allyaml: allyaml  ## Check that all.yaml is valid (Need Minikube running)
...
 
-
-## These targets used for dev work
-#
-
-# Expose services using NodePorts.  Normally disabled but often used
-# with minikube.  We enable for debugging purposes.
-# (If desired, you can override with something like "make <target> EXPOSE_METHOD=LoadBalancer"
-# or "make <target> EXPOSE_METHOD=" to get the default unexposed behavior.)
-EXPOSE_SERVICES=--set exposeServices=NodePort
```
marcoderama: It looks like EXPOSE_SERVICES is undefined now, so what of the reference on line 115/91 below?  That either needs to be removed (in which case we can only use the default), or we need some value defined.

enterprise-suite/Makefile
```diff
@@ -35,26 +34,20 @@ script-tests:
...
-minikube-tests: build install-local
-	@echo Installed Console from helm charts: $(CHART)
-	@echo Running tests
-	( cd $(CHART_DIR)/tests ; ./smoketest ; )
+minikube-backend-tests: build
+	@echo Running minikube tests
+	( cd $(CHART_DIR)/tests ; ./smoketest_minikube.sh ; )
 
-backend-tests: script-tests minikube-tests
+openshift-backend-tests: build
```
marcoderama: These new targets should be added to the `.PHONY` line at the bottom.  (And the old ones removed.)

I've been considering changing to the style that adds the `.PHONY` bit with each such target.  I think this would reduce the probability of these getting overlooked.  e.g.
```
[...]
.PHONY: openshift-backend-tests
openshift-backend-tests: build
[...]
```

.travis.yml
```diff
@@ -12,23 +12,30 @@ install:
...
     if: type in (cron)
     script:
-    - scripts/setup-minikube-for-linux.sh
-    - make test CHARTS=enterprise-suite-latest
+    - scripts/setup-minikube-for-linux.sh || travis_terminate 1
+    - make -C enterprise-suite-latest minikube-backend-tests
 
-  - name: ES Frontend E2E
+  - name: Frontend E2E Tests - Minikube
+    if: false
```
marcoderama: This here for testing?

enterprise-suite/tests/resources/ingress-test.tls.yaml
```diff
@@ -2,7 +2,6 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: ingress-test-tls
-  namespace: lightbend
```
marcoderama: This still works without this?

enterprise-suite/tests/smokecommon
```diff
@@ -108,13 +108,13 @@ test_summary() {
     exit $TEST_FAILURES
 }
 
-# Get URL for NodePort assignment
+# Get URL for accessing console externally
 # nodeport <name>
 nodeport() {
-  if [[ $CURRENT_CONTEXT == "minishift" ]]; then
```
marcoderama: Doesn't need to work on minishift anymore or does openshift cover that too?

enterprise-suite/tests/smoketest
```diff
@@ -21,27 +21,37 @@ function print_debug() {
...
+        kubectl -n $NAMESPACE logs --all-containers "$pod"
         echo
     done
     echo
     echo "========================================"
     echo
-    echo -e "\ntravis_fold:end:debug${1}.logs\r"
 }
 
-print_debug "pre"
```
marcoderama: oops

enterprise-suite/tests/smoketest_minikube.sh
```diff
@@ -0,0 +1,49 @@
...
+function cleanup() {
+    # optimization - no need for cleanup on travis, as VM will be torn down
+    if [[ "${TRAVIS:-}" == "true" ]]; then return; fi
+
+    helm del --purge enterprise-suite
+    kubectl delete namespace ${NAMESPACE}
+}
+
+# Only run main if not sourced.
+(return 2>/dev/null) || main
```
marcoderama: Interesting stunt.

scripts/setup-openshift.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -e
+
+curl -OL https://github.com/openshift/origin/releases/download/v3.10.0/openshift-origin-client-tools-v3.10.0-dd10d17-linux-64bit.tar.gz
```
marcoderama: I see `v3.10.0` in here several times.  Worth putting that in var to make upgrades less error prone?

.travis.yml
jsravn: Yep, need to remove these when it's ready.

enterprise-suite/Makefile
jsravn: Good point, I missed that. But I'm not sure we should be exposing nodeport by default in all.yml.

enterprise-suite/Makefile
jsravn: These were all moved into tests/smoketest_minikube.sh or tests/smoketest_openshift.sh, because the commands are different based on where you are installing it. In fact you can just run tests/smoketest_minikube.sh now, no need to use the make target. Is that okay? Otherwise I can try to extract the helm setup for minikube into its own shell script, if you want to just invoke that.

enterprise-suite/Makefile
```diff
@@ -90,53 +83,13 @@ lint-allyaml: allyaml  ## Check that all.yaml is valid (Need Minikube running)
...
-HELM_VALUES_FLAG=$(if $(HELM_VALUES_FILE),--values $(HELM_VALUES_FILE))
-
-# Used for minikube tests.
-install-local: $(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz install-helm  ## Install local tarball
-	# We set a podUID here for test purposes to ensure everything works as non-root.
-	# We usePersistentVolumes to ensure PVCs work as expected.
-	-$(CHART_DIR)/scripts/lbc.py install --local-chart=$(HELM_CHARTS_DIR)/docs/$(RELEASE).tgz \
-		--wait $(EXPOSE_SERVICES) --set podUID=10001,usePersistentVolumes=true -- $(HELM_VALUES_FLAG)
-
-install-dev: install-helm ## Install content of enterprise-suite/ folder
```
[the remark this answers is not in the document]
jsravn: As said above, I moved all the setup code into shell scripts, where I can also share common logic between openshift and minikube. We could keep these but I never use them myself - always just invoke the lbc.py script directly. I think that's cleaner than putting another layer of logic in the Makefile that we need to maintain.

enterprise-suite/tests/smokecommon
jsravn: We don't test on minishift, so I have no idea if this code even works. It didn't work out of the box on openshift. We should avoid keeping code that we never run, it just rots.

enterprise-suite/tests/smoketest
jsravn: I removed this on purpose to try and avoid nuking Travis - as our builds seem to die during this step. I'm not sure we need it anyways, as we log at the end.

enterprise-suite/tests/resources/ingress-test.tls.yaml
jsravn: Yes because we specify the namespace when applying.

enterprise-suite/Makefile
jsravn: Other benefit is to break cyclic dependencies between makefile and our scripts (the makefile calling a script which calls the makefile).

enterprise-suite/Makefile
```diff
@@ -35,26 +34,20 @@ script-tests:
 	$(CHART_DIR)/tests/lib/bats/bats $(CHART_DIR)/tests
 	python $(CHART_DIR)/scripts/lbc_test.py
 
-minikube-tests: build install-local
-	@echo Installed Console from helm charts: $(CHART)
-	@echo Running tests
-	( cd $(CHART_DIR)/tests ; ./smoketest ; )
+minikube-backend-tests: build
```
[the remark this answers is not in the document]
jsravn: Hmm not sure if we even need the build thing anymore. I think that is a holdover from when we built against the tarball rather than the local dir.

enterprise-suite/Makefile
jsravn: It's a matter of style I guess. Some people might find it annoying having .PHONY everywhere.

jsravn: I'll move it to one per target, because right now it's a mess of non-existent targets.

enterprise-suite/Makefile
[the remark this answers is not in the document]
jsravn: Can you let me know what would be the most useful? I don't want to keep all the targets just for personal use, and these are no longer used by the tests. Would a single target which installs helm and the local chart suffice?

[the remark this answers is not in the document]
jsravn: Maybe:

```make
install-dev:
  helm init
  scripts/lbc.py install --local-chart=. --wait --set exposeServices=NodePort
```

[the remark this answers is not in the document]
jsravn: Okay, I've added that. Let me know if it is sufficient.

enterprise-suite/tests/smoke_prometheus
```diff
@@ -9,8 +9,7 @@ source smokecommon
 #
 
 set -e
-NAMESPACE="smoketest-prometheus"
-kubectl create namespace $NAMESPACE
+NAMESPACE=${NAMESPACE:=lightbend}
```
marcoderama: I assume this change is fine.  I just figured there was _some_ reason for using the other namespace originally.  :-)  Didn't have the nerve to change it...

enterprise-suite/tests/smokecommon
marcoderama: We don't test on minishift from Travis, but I wonder if @srikanthops was using this...

enterprise-suite/tests/smokecommon
```diff
@@ -108,13 +108,13 @@ test_summary() {
     exit $TEST_FAILURES
 }
 
-# Get URL for NodePort assignment
+# Get URL for accessing console externally
 # nodeport <name>
```
marcoderama: Should remove this comment since it seems we don't use any supplied arg now.

enterprise-suite/tests/smoketest
marcoderama: Heh.  I figured that was some cruft I accidentally left behind.  :-)

enterprise-suite/Makefile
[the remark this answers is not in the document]
marcoderama: Having the help text means it shows up in the output of `make help` is all.  I try to add it for any of the targets people are at all likely to use.

enterprise-suite/Makefile
jsravn: Yeah I didn't want to change it. But we should just remove all.yaml as I keep saying :).

enterprise-suite/tests/smokecommon
jsravn: Setting the context to "openshift" should have the same effect.

enterprise-suite/Makefile
jsravn: Yes, it's a good point. Personally I don't use anything except lbc.py. I set up helm just from the command line, with helm init. It's hard to cater for everyone, so we should draw the line somewhere of what we keep in here. For me it's customer facing stuff like lbc.py, the rest is just infrastructure for running our CI.

PR #224, opened 2018-12-12
Content-type for /version response is application/json

marcoderama:
Fix for https://github.com/lightbend/es-backend/issues/530


marcoderama:
It's not clear why `add_header` was additive, but this change fixes the problem.
```
HTTP/1.1 200 OK
Server: nginx/1.15.7
Date: Wed, 12 Dec 2018 23:12:23 GMT
Content-Type: application/json
Content-Length: 2274
Connection: keep-alive
```

As suggested by second answer to https://stackoverflow.com/questions/19629930/force-nginx-to-send-specific-content-type/33327564

PR #226, opened 2018-12-13
Remove obsolete cluster role check from lbc.py

jsravn:
The cluster roles are namespaced now, so this check doesn't make sense.

Also, it doesn't work if the installer has limited permissions.

Related to https://github.com/lightbend/es-backend/issues/420

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -50,11 +50,6 @@ CONSOLE_PVCS = [
     'prometheus-storage'
 ]
 
-CONSOLE_CLUSTER_ROLES = [
-    'kube-state-metrics',
-    'prometheus-server'
-]
-
 DEFAULT_TIMEOUT=3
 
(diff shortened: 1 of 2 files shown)


jsravn:
Need to fix the test - but out of time today, will do it tomorrow.

enterprise-suite/scripts/lbc.py
```diff
@@ -282,14 +277,6 @@ def are_pvcs_created(namespace):
         fail_msg='Found some PVCs ({}) from previous console install, but not all expected: {}.\nTo avoid data loss, please remove them manually'
     )
 
-# Checks for console cluster roles
-def are_cluster_roles_created():
```
marcoderama: Could we still do this test
- if we include namespace in the check query, and
- only do it on envs where it's expected to work?

I guess I'm just asking if there's still some value here.

marcoderama: Got it.  Thanks.

PR #227, opened 2018-12-14
Don't add createPersistentVolumes=false if PVCs exist

jsravn:
Otherwise, it will actually delete the existing PVCs on an upgrade.

Instead, simply warn the user and let them make the correct choice.

The new output looks like:

```
$ scripts/lbc.py install --namespace=jsravn-console --local-chart=. --set usePersistentVolumes=true --set defaultStorageClass=gp2
info: Found existing PVCs from a previous console installation.
info: Please remove them with `kubectl delete pvc`, or pass --set createPersistentVolumes=false.
info: Otherwise, the install may fail.
helm install . --name enterprise-suite --namespace jsravn-console --devel --values /tmp/tmpEqcPas --set usePersistentVolumes=true --set defaultStorageClass=gp2 
Error: release enterprise-suite failed: persistentvolumeclaims "alertmanager-storage" already exists


Command 'helm install . --name enterprise-suite --namespace jsravn-console --devel --values /tmp/tmpEqcPas --set usePersistentVolumes=true --set defaultStorageClass=gp2 ' failed!
```

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -396,7 +396,8 @@ def install(creds_file):
                 .format(args.helm_name, chart_ref, version_arg,
                         creds_arg, helm_args))
         else:
-            if are_pvcs_created(args.namespace):
+            createPVs = len(filter(lambda x: 'createPersistentVolumes=false' in x, sys.argv)) == 0
+            if createPVs and are_pvcs_created(args.namespace):
                 printerr('info: Found existing PVCs from a previous console installation.')
                 printerr('info: Please remove them with `kubectl delete pvc`, or pass --set createPersistentVolumes=false.')
                 printerr('info: Otherwise, the install may fail.')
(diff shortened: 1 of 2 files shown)


marcoderama:
A bit late in the game but...

> $ scripts/lbc.py install --namespace=jsravn-console --local-chart=. --set usePersistentVolumes=true --set defaultStorageClass=gp2
info: Found existing PVCs from a previous console installation.
info: Please remove them with `kubectl delete pvc`, or pass --set createPersistentVolumes=false.
info: Otherwise, the install may fail.
helm install . --name enterprise-suite --namespace jsravn-console --devel --values /tmp/tmpEqcPas --set usePersistentVolumes=true --set defaultStorageClass=gp2 
Error: release enterprise-suite failed: persistentvolumeclaims "alertmanager-storage" already exists

I'm not sure how to interpret
> info: Please remove them with `kubectl delete pvc`, or pass --set createPersistentVolumes=false.

I assume that `kubectl delete pvc` will delete them but what about `--set createPersistentVolumes=false`?  Will that preserve them?  Basically, we should make it clearer in the sense of:
- Actually want to keep them?  Do this...
- Actually want to delete them?  Do this...

jsravn:
Sure, we can clarify. Next time feel free to open a PR with the change, no need to discuss it beforehand.

PR #232, opened 2018-12-19
Add default monitor for bad monitor count 

srikanthops:
lightbend/es-backend/issues/509
**Haven't unit tested these changes yet**

* Add scraping for es-monitor-api
* Add new default monitor es_monitor_api_monitor_errors 
* Add new rule


enterprise-suite/es-monitor-api/default-monitors.json
```diff
@@ -104,6 +104,24 @@
           "description": "{{$labels.instance}} has anomalous scrape_duration_seconds"
         }
       },
+      "es_monitor_api_errors": {
```
dbrinegar: "lightbend" instead of "es" ?  what do you think?

enterprise-suite/es-monitor-api/default-monitors.json
```diff
@@ -104,6 +104,24 @@
...
+          "metric": "es_monitor_api_last_error",
+          "window": "1m",
+          "confidence": "1",
+          "severity": {
+            "warning": {
+              "comparator": ">",
+              "threshold": "0"
+            }
+          },
+          "summary": "es console monitor errors",
```
dbrinegar: "lightbend" instead of "es" seems more clear to me. Imagine this showing up in PagerDuty at 3am

enterprise-suite/es-monitor-api/default-monitors.json
```diff
@@ -104,6 +104,24 @@
...
+          "description": "one or more monitors are erroneous"
```
dbrinegar: "one or more monitors are erroneous" seems an odd phrasing and possibly not correct if the issue is something not in the parsing of the monitor. I think we want to direct attention to logs for specifics, so I would maybe put the instance information in here

"Lightbend monitor api on {{$labels.instance}} has errors"

enterprise-suite/es-monitor-api/default-monitors.json
srikanthops: how about 
lb_console_monitor_api_errors

enterprise-suite/es-monitor-api/default-monitors.json
srikanthops: agree will change

PR #234, opened 2019-01-04
change selector to reflect new structure

pcraddick:
Another one of these trailing changes to reflect the structure of code that will soon be merged to master.


dbrinegar:
I think we need a release candidate to see the tests pass. Some bump here:

https://github.com/lightbend/console-charts/blob/master/enterprise-suite/values.yaml#L5

dbrinegar:
Force merging this to break the diamond dependency in es-monitor-api tests. Master will be broken for a bit.

PR #235, opened 2019-01-05
bump es-console-spa version

pcraddick:


dbrinegar:
I can't tell why the smoketest is failing, maybe travis.org is super underwater

jsravn:
The smoketests are just mega flaky. It's slowed me down a lot in the openshift ticket as well (where they seem to fail even more often randomly). Really hoping @mitkus rewrite of the tests will help us nail some of these problems down. Otherwise we just need to roll up our sleeves and figure out the underlying problems.

dbrinegar:
doing this in line with the test change over in #236

PR #236, opened 2019-01-07
name change

dbrinegar:
just trying to get master green again

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,11 +29,11 @@ jobs:
     - make test CHARTS=enterprise-suite-latest
 
   - name: ES Frontend E2E
-    addons:
-      apt:
-        update: true
-        packages:
-        - libgconf2-4
+    # addons:
+    #   apt:
+    #     update: true
+    #     packages:
+    #     - libgconf2-4
     cache:
       directories:
         - ~/.npm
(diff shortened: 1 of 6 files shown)


jsravn:
I fixed the apt build failure at https://github.com/lightbend/console-charts/pull/239. So master should be green now.

.travis.yml
```diff
@@ -48,11 +48,11 @@ jobs:
...
-    addons:
-      apt:
-        update: true
-        packages:
-        - libgconf2-4
+    # addons:
+    #   apt:
+    #     update: true
+    #     packages:
+    #     - libgconf2-4
```
jsravn: I don't think this is necessary anymore since I fixed the apt issue.

enterprise-suite/tests/smokecommon
```diff
@@ -143,7 +143,7 @@ test_es_console_responding() {
   CONSOLE_URL=$1
   shift
   es_console_responding() {
-      curl_query $CONSOLE_URL "" "$@" | fgrep -q '<title>Enterprise Suite Reactive Console</title>'
+      curl_query $CONSOLE_URL "" "$@" | fgrep -q '<title>Enterprise Suite Lightbend Console</title>'
```
jsravn: Hmm interesting. Why do the ingress tests pass though if this is completely wrong? Is there something messed up in the test detecting failures?

jsravn: "Enterprise Suite Reactive Console" is the correct title I think. I deployed master, and that is the title being used.

PR #237, opened 2019-01-07
whitesource image scan

dbrinegar:
this is going to wedge on master being red, fixing that first...

### DIFF
--- a/enterprise-suite/.ws.conf
+++ b/enterprise-suite/.ws.conf
@@ -1,0 +1,10 @@
+productName=Console Docker Images
+projectName=not applicable
+
+docker.scanImages=true
+docker.includes=.*enterprise-suite.* alpine busybox prom/.* .*kube-state-metrics janeczku/go-dnsmasq jimmidyson/configmap-reload
+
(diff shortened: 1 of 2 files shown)


enterprise-suite/.ws.conf
```diff
@@ -0,0 +1,10 @@
+productName=Console Docker Images
+projectName=not applicable
+
+docker.scanImages=true
+docker.includes=.*enterprise-suite.* alpine busybox prom/.* .*kube-state-metrics janeczku/go-dnsmasq jimmidyson/configmap-reload
```
jsravn: Does this need versions?

Also - is it documented how this works somewhere? I'm not sure how this picks up the actual image names or knows the repo.

dbrinegar: docs are https://whitesource.atlassian.net/wiki/spaces/WD/pages/33718339/Unified+Agent

and

https://github.com/whitesource/fs-agent

dbrinegar: it does `docker images` against minikube and unpacks whatever matches the regex

PR #244, opened 2019-01-11
Allow for specification of add'l Grafana configs in values.yaml

marcoderama:
With the sample configs uncommented, we see this after running
`helm template . -x templates/es-grafana-deployment.yaml -f values.yaml`
(Things are unchanged if no values are provided.)

```
[...]
        env:
          [...]
          - name: GF_ANALYTICS_CHECK_FOR_UPDATES
            value: "false"
          - name: GF_SMTP_ENABLED
            value: "true"
          - name: GF_SMTP_HOST
            value: "smtp.gmail.com:465"
          - name: GF_SMTP_PASSWORD
            value: "password"
          - name: GF_SMTP_USER
            value: "username"
        ports:
[...]
```

### DIFF
--- a/enterprise-suite/templates/es-grafana-deployment.yaml
+++ b/enterprise-suite/templates/es-grafana-deployment.yaml
@@ -50,6 +50,10 @@ spec:
             value: "false"
           - name: GF_ANALYTICS_CHECK_FOR_UPDATES
             value: "false"
+          {{- range $key, $val := .Values.esGrafanaEnvVars }}
+          - name: {{ $key }}
+            value: {{ $val | quote }}
+          {{- end }}
         ports:
           - containerPort: 3000
         lifecycle:
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -122,3 +122,16 @@ prometheusVolumeSize: 256Gi
 alertmanagerVolumeSize: 32Gi
 # Grafana volume size - used for storing users, custom dashboards, and plugins.
 esGrafanaVolumeSize: 32Gi
+
+#################################################
+# Grafana environment variables
+#
+# Any additional configuration for Grafana, using environment variables as specified at
+# http://docs.grafana.org/installation/configuration/#using-environment-variables.
+#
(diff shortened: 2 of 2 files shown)


enterprise-suite/templates/es-grafana-deployment.yaml
```diff
@@ -50,6 +50,10 @@ spec:
             value: "false"
           - name: GF_ANALYTICS_CHECK_FOR_UPDATES
             value: "false"
+          {{- range $key, $val := .Values.esGrafanaConfigs }}
+          - name: {{ $key }}
+            value: {{ $val | quote }}
+          {{- end }}
```
jsravn: Are there any values above that should be moved so they are configurable as well?

enterprise-suite/values.yaml
```diff
@@ -122,3 +122,16 @@ prometheusVolumeSize: 256Gi
...
 # Grafana volume size - used for storing users, custom dashboards, and plugins.
 esGrafanaVolumeSize: 32Gi
+
+#################################################
+# Grafana configs
+#
+# Any additional configuration for Grafana, using environment variables as specified at
+# http://docs.grafana.org/installation/configuration/#using-environment-variables.
+#
+esGrafanaConfigs:
```
jsravn: I think it should be called `grafanaEnvVars` or something similar, so it's clear what they are.

enterprise-suite/templates/es-grafana-deployment.yaml
marcoderama: Having this loop at the end allows a user to override any of the settings above.  (Just confirmed with a test.)  I actually wondered if there were any we _didn't_ want to allow the user to modify.

I figure they're free to break things if they want though...

jsravn: Yes, possibly. It's the most flexible approach in case a customer wants to change something.

Unfortunately this is sort of the nature of helm - to make it customizable, you eventually end up putting everything in values.yaml.

PR #247, opened 2019-01-14
whitesource scan docker images after smoke tests

dbrinegar:
Honestly not sure where this should go. We need the images to be available, so I figure after the smoke tests. This seems expedient but perhaps not ideal as it ends up kind of buried in the test job.

There might be a pure-docker way to do this without kubernetes, as all we need is for the docker client in the scanner to be able to unpack the images, but I couldn't sort it out quickly. If we did this in a separate minikube job, then we'd have to write something to wait for the helm install to be done downloading all the images needed.


jsravn:
For https://github.com/lightbend/es-backend/issues/458

PR #249, opened 2019-01-16
Support different frontend image for e2e test

katsutoxin:
Support different frontend image for e2e test. 
For example, run 
```
ES_CONSOLE_VERSION=e2e make frontend-tests
```

The purpose of this PR is to use the same e2e setup cross repo es-console-spa and helm-charts

### DIFF
--- a/enterprise-suite/tests/smoketest_minikube.sh
+++ b/enterprise-suite/tests/smoketest_minikube.sh
@@ -22,20 +22,20 @@ function diagnostics() {
 }
 
 function setup() {
-	kubectl create namespace ${NAMESPACE}
-	kubectl create namespace ${TILLER_NAMESPACE}
-	kubectl create serviceaccount --namespace ${TILLER_NAMESPACE} tiller
-	kubectl create clusterrolebinding ${TILLER_NAMESPACE}:tiller --clusterrole=cluster-admin \
-	    --serviceaccount=${TILLER_NAMESPACE}:tiller
-	helm init --wait --service-account tiller --upgrade --tiller-namespace=${TILLER_NAMESPACE}
+    kubectl create namespace ${NAMESPACE}
+    kubectl create namespace ${TILLER_NAMESPACE}
+    kubectl create serviceaccount --namespace ${TILLER_NAMESPACE} tiller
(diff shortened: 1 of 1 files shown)


jsravn:
@katsutoxin I'm a little confused how this is used? Also how would we prevent it from breaking something in console-spa? There's no way to know how this is used from looking at the change alone.

enterprise-suite/tests/smoketest_minikube.sh
```diff
@@ -22,6 +22,8 @@ function diagnostics() {
 }
 
 function setup() {
+	local es_console_version; es_console_version=''
```
jsravn: Use `local es_console_version=`.

enterprise-suite/tests/smoketest_minikube.sh
```diff
@@ -22,6 +22,8 @@ function diagnostics() {
...
+	[[ -n "${ES_CONSOLE_VERSION:-}" ]] && es_console_version=",esConsoleVersion=${ES_CONSOLE_VERSION}"
```
jsravn: It would be nicer to just do `es_console_version="--set esConsoleVersion=${ES_CONSOLE_VERSION}"`.

katsutoxin:
> @katsutoxin I'm a little confused how this is used? Also how would we prevent a change to it in the future from breaking something in console-spa? There's no way to know how this is used from looking at console-charts alone.

@jsravn Is it to use for es-console-spa prs test.  In es-console-spa prs test, we want to re-use the script smoketest_minikube.sh.  However, currently smoketest_minikube.sh is always using the esConsoleVersion hardcoded in `enterprise-suite/values.yaml`, we want to pass different value.   For example, in es-console-spa prs build, we build a es-console-spa docker image with version "e2e", and we call `ES_CONSOLE_VERSION=e2e make frontend-tests` to do the test.

katsutoxin:
The usage example is here https://github.com/lightbend/es-console-spa/blob/e0c42fdd4188cf226bc677e3670aeee5ee92353c/scripts/run_e2e_travis.sh#L10

I created a es-console-spa travis build based on this  https://travis-ci.com/lightbend/es-console-spa/builds/97626591  (the e2e test).  It works.

PR #251, opened 2019-01-18
Security headers

dbrinegar:
Fixes lightbend/es-backend#552

### DIFF
--- a/enterprise-suite/tests/smoke_nginx
+++ b/enterprise-suite/tests/smoke_nginx
@@ -92,3 +92,19 @@ securitay "${CONSOLE}/service/prometheus/"
 securitay "${CONSOLE}/service/grafana/"
 securitay "${CONSOLE}/service/alertmanager/"
 securitay "${CONSOLE}/service/es-monitor-api/"
+
+#
+# xss guard
+# this tries to inject a quote to break html string injected into base href
+# the guard works if base href is a normal string with two quotes
+# like: "/cluster%22/%3E%3Cscript%3Ealert(789)%3C/"
+# and fails if something with extra quote marks
+# like: "/cluster"/><script>alert(789)</script>/"
+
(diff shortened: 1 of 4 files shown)


enterprise-suite/tests/smoke_nginx
```diff
@@ -48,3 +48,63 @@ test_no_cache() {
...
+    return 1
+  fi
+  if [[ $val != "$3" ]]; then
+    echo "expecting $2: $3, got \"$val\""
+    return 1
+  fi
+  return 0
+}
+
+securitay() {
```
marcoderama: Any particular reason for the name?  Would `security` make more sense?

enterprise-suite/tests/smoke_nginx
```diff
@@ -48,3 +48,63 @@ test_no_cache() {
...
+  headers=$( curl_headers "$1" )
+
+  ! echo "$headers" | grep -q "^Server: "
+  T $? "$1" has no Server header
+
+  check_header "$headers" "X-Frame-Options" "DENY"
+  T $? "$1" has X-Frame-Options
```
marcoderama: Should mention "DENY" here?  (And "1" below.)

enterprise-suite/tests/smoke_nginx
dbrinegar: download

dbrinegar: image

jsravn:
@dbrinegar are there any tests we can add to verify this works as expected? Or do you think the existing ones are sufficient..?

enterprise-suite/tests/smoke_nginx
jsravn: I agree w/ @marcoderama on this one :). This repo is open source and the face of our product after all. Let's keep it professional.

PR #252, opened 2019-01-18
update prometheus to v2.7.0-rc.0

dbrinegar:
* fix syntax errors in monitor description templates
* upgrade prometheus to rc with subqueries

subqueries allow us to eliminate promjs for backtesting monitors, now we can do it all in a server query

https://docs.google.com/document/d/1P_G87zN88YvmMr4iwLWygChMTZhai1L7S_c0awu1CAE/edit


jsravn:
Shouldn't we wait for the final release?

PR #253, opened 2019-01-21
Drop go-dnsmasq

jsravn:
As it's unmaintained, and we don't need the ability to re-resolve
service IPs at runtime.


jsravn:
It doesn't work, because we are using variables in proxy_pass which requires an explicit resolver in nginx.

PR #256, opened 2019-01-22
Rename createPersistentVolumes to managePersistentVolumes

jsravn:
So it's meaning is hopefully clearer, to avoid accidental data loss.

Also a few other changes to prevent accidental data loss:
- Remove convoluted logic around whether to create the PVC resource.
Instead, just use `managePersistentVolumes` as the flag. So it's very
clear there is only one way to delete PVCs (set managePersistentVolumes
to false).
- Remove deprecated parameter `useEmptyDirVolumes`. Instead use
`usePersistentVolumes` (which defaults to false).

For https://github.com/lightbend/es-backend/issues/554

### DIFF
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -108,12 +108,11 @@ prometheusMemoryRequest: 250Mi
 #################################################
 # Persistent Volumes
 #
-# Deprecated value - set to false to disable emptyDir and use persistent volumes. Prefer `usePersistentVolumes`.
-useEmptyDirVolumes: true
 # Set to true to use persistent volumes, otherwise emptyDir volumes are used.
 usePersistentVolumes: false
-# If true, the chart creates persistent volumes using PVCs. Set to false to reuse previously created persistent volumes.
-createPersistentVolumes: true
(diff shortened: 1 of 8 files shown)


enterprise-suite/values.yaml
```diff
@@ -108,12 +108,11 @@ prometheusMemoryRequest: 250Mi
...
 # Persistent Volumes
 #
-# Deprecated value - set to false to disable emptyDir and use persistent volumes. Prefer `usePersistentVolumes`.
-useEmptyDirVolumes: true
 # Set to true to use persistent volumes, otherwise emptyDir volumes are used.
 usePersistentVolumes: false
-# If true, the chart creates persistent volumes using PVCs. Set to false to reuse previously created persistent volumes.
-createPersistentVolumes: true
+# If true, the chart manages persistent volumes by creating PVCs. Set to false to not manage PVCs at all.
+# Warning: If you "unmanage" PVCs that helm is managing by setting this to false, helm will delete those PVCs.
```
marcoderama: is -> was?

jsravn: I think "is" is right here, since it's saying helm is currently managing rather than "was" in the past.

PR #258, opened 2019-01-23
Use cluster DNS directly in nginx config

jsravn:
Instead of using go-dnsmasq, just parse /etc/resolv.conf directly and
insert the needed values into nginx.conf.

This avoids the need for an intermediate container, and relies on the
stable interface of /etc/resolv.conf. Note there is currently no way to
query k8s for this information other than checking resolv.conf.

For https://github.com/lightbend/es-backend/issues/541

### DIFF
--- a/enterprise-suite/templates/es-console-configmap.yaml
+++ b/enterprise-suite/templates/es-console-configmap.yaml
@@ -18,12 +18,12 @@ data:
 
       # use external resolver to lookup backends, cache for 30 seconds
 
-      resolver 127.0.0.1:5353 ipv6=off valid=30s;
+      resolver $KUBE_RESOLVER ipv6=off valid=30s;
 
-      set $prometheus "prometheus-server";
-      set $monitorapi "es-monitor-api";
-      set $grafana "grafana-server";
-      set $alertmanager {{ splitList "," .Values.alertManagers | first | quote }};
+      set $prometheus "prometheus-server.$KUBE_DOMAIN";
+      set $monitorapi "es-monitor-api.$KUBE_DOMAIN";
+      set $grafana "grafana-server.$KUBE_DOMAIN";
+      set $alertmanager {{ print (splitList "," .Values.alertManagers | first) `.$KUBE_DOMAIN:9093` | quote }};
 
       # nginx config primer:
       # location ~ (regex.*)(matchers.*) { regex matchers become $1 and $2 in the block }
--- a/enterprise-suite/templates/es-console-configmap.yaml
+++ b/enterprise-suite/templates/es-console-configmap.yaml
@@ -23,7 +23,7 @@ data:
       set $prometheus "prometheus-server.$KUBE_DOMAIN";
       set $monitorapi "es-monitor-api.$KUBE_DOMAIN";
       set $grafana "grafana-server.$KUBE_DOMAIN";
-      set $alertmanager {{ print (splitList "," .Values.alertManagers | first) `.$KUBE_DOMAIN:9093` | quote }};
+      set $alertmanager {{ splitList "," .Values.alertManagers | first | quote }};
 
       # nginx config primer:
       # location ~ (regex.*)(matchers.*) { regex matchers become $1 and $2 in the block }
--- a/enterprise-suite/templates/es-console-configmap.yaml
+++ b/enterprise-suite/templates/es-console-configmap.yaml
@@ -20,9 +20,9 @@ data:
 
       resolver $KUBE_RESOLVER ipv6=off valid=30s;
 
-      set $prometheus "prometheus-server.$KUBE_DOMAIN";
-      set $monitorapi "es-monitor-api.$KUBE_DOMAIN";
-      set $grafana "grafana-server.$KUBE_DOMAIN";
+      set $prometheus "prometheus-server.$KUBE_CONSOLE_DOMAIN";
+      set $monitorapi "es-monitor-api.$KUBE_CONSOLE_DOMAIN";
+      set $grafana "grafana-server.$KUBE_CONSOLE_DOMAIN";
       set $alertmanager {{ splitList "," .Values.alertManagers | first | quote }};
 
       # nginx config primer:
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -71,7 +71,7 @@ imageCredentials:
 createAlertManager: true
 
 # Comma separated list of alertmanager addresses (can be k8s local DNS service names)
-alertManagers: alertmanager.$KUBE_DOMAIN:9093
+alertManagers: alertmanager.$KUBE_CONSOLE_DOMAIN:9093
 
 # Alertmanager ConfigMap. Set to the name of a ConfigMap, with a file `alertmanager.yml`.
 alertManagerConfigMap: alertmanager-default
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -15,10 +15,6 @@ alertManagerVersion: v0.15.1
 configMapReloadVersion: v0.2.2
 # gcr.io/google_containers/kube-state-metrics
 kubeStateMetricsVersion: v1.2.0
-# prom/node-exporter
-nodeExporterVersion: v0.15.2
-# janeczku/go-dnsmasq
-goDnsmasqVersion: release-1.0.7
 # alpine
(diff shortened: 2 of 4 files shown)


enterprise-suite/templates/es-console-configmap.yaml
```diff
@@ -18,12 +18,12 @@ data:
...
+      resolver $KUBE_RESOLVER ipv6=off valid=30s;
 
-      set $prometheus "prometheus-server";
-      set $monitorapi "es-monitor-api";
-      set $grafana "grafana-server";
-      set $alertmanager {{ splitList "," .Values.alertManagers | first | quote }};
+      set $prometheus "prometheus-server.$KUBE_DOMAIN";
+      set $monitorapi "es-monitor-api.$KUBE_DOMAIN";
+      set $grafana "grafana-server.$KUBE_DOMAIN";
+      set $alertmanager {{ print (splitList "," .Values.alertManagers | first) `.$KUBE_DOMAIN:9093` | quote }};
```
marcoderama: Any way to parameterize `9093`?

marcoderama: I mean what if the list is user-supplied and the first entry happens to run on a different port?  Do we maybe have to somehow do something like:
```
if value contains a port
   use port
else
   use default of 9093
```

enterprise-suite/values.yaml
```diff
@@ -79,7 +71,7 @@ imageCredentials:
 createAlertManager: true
 
 # Comma separated list of alertmanager addresses (can be k8s local DNS service names)
-alertManagers: alertmanager:9093
+alertManagers: alertmanager
```
mitkus: Removing port number here needs to also be reflected in the docs: https://github.com/lightbend/es-docs/blob/master/src/main/paradox/installation/alertmanager.md#installing-with-existing-alertmanager

enterprise-suite/templates/es-console-configmap.yaml
jsravn: I was avoiding the problem, because the other pieces also don't specify port :). But since we do support external AM, I should fix this. I'll have a think.

enterprise-suite/values.yaml
jsravn: Fixed it to not use port.

enterprise-suite/templates/es-console-configmap.yaml
jsravn: Okay, I changed this by moving KUBE_DOMAIN into values.yaml. That makes more sense anyways - if a user wants to provide their own alertmanagers, it won't make sense to append KUBE_DOMAIN.

enterprise-suite/values.yaml
```diff
@@ -79,7 +71,7 @@ imageCredentials:
 createAlertManager: true
 
 # Comma separated list of alertmanager addresses (can be k8s local DNS service names)
-alertManagers: alertmanager:9093
+alertManagers: alertmanager.$KUBE_CONSOLE_DOMAIN:9093
```
marcoderama: How does `$KUBE_CONSOLE_DOMAIN` get set when we kick off console locally?  Does this need to be set in the Makefile?

jsravn: No it's not evaluated here - it gets copied directly into `default.conf` as `$KUBE_CONSOLE_DOMAIN`. Which then gets replaced at start up.

marcoderama: Right.  D'oh!

dbrinegar: I think the problem is ```            - --alertmanagers={{ .Values.alertManagers }}
``` over in prometheus-deployment

**Comment by dbrinegar** [comment]
Alright I think I better understand the intricacies of this change and I'm more sceptical that this is a general purpose solution, as we had before. I feel like we're going from a full featured name resolver to hoping that /etc/resolv.conf is set up a certain way.

But

1. There could be multiple resolvers. Old way would balance between them, back off down resolvers, and cache. New way picks first and sticks to it, no cache. This could cause problems with lookup load as well as inability to handle failure when that first dns node is unreachable.

2. There could be performance concerns. Old way caches each proxy domain so only one remote lookup per domain is done per cache tick. New way does a remote dns lookup for every proxied request.  The old way handles arbitrary number of UI clients with no extra dns load, while new way compounds dns load with each client.

3. There could be up to six search domains. Old way would walk through them, not needing fully qualified domains, until it matches. New way picks the first domain and if it doesn't work, we're sunk. We cannot take partial domain names like "alertmanager.eastcoast." I think we must then be very careful to qualify what we accept in those variables so they have no dots, and then require /etc/resolv.conf to follow a convention where the first domain is fully qualified for a no-dot name. This seems to me one of those things that will go unnoticed as long as it works, then a giant pain when it fails to work.

3. other /etc/resolv.conf options are ignored. Old way would obey options, ndots, timeout, attempts, rotate, etc.

As a simple example from minikube:

root@prometheus-server-84b7ff9fd6-2pcdq:/# cat /proc/1/root/etc/resolv.conf 
```
nameserver 10.96.0.10
search lightbend.svc.cluster.local svc.cluster.local cluster.local
options ndots:5
```

if we ask for nginx to proxy  "alertmanager.lightbend" then the init script will incorrectly change that to "alertmanager.lightbend.lightbend.svc.cluster.local." and fail, while such a setting would work fine the old way and elsewhere as a perfectly valid internal domain name.

As another example, this silly example from the docs on how one might customize dns is just dead in the water for our new method, as there are two peer searches that must be done (two fully qualified domains).

https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/service/networking/custom-dns.yaml



I think it might be better to have a full featured dns client.

jsravn:
As a side note, I'm not sure why we specify `valid` since nginx already respects TTL: http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver.  May be worth removing that so we rely on the DNS record.

dbrinegar:
Since we've alighted on the same outcome we don't need to spend tons of time in the weeds, but just for the sake of clarifying my understanding:

"Mutliple resolvers" means multiple hosts serving dns, peers not cascading. Normally a prod environment would have multiple dns hosts, all serving the same thing, right? For load balancing, failure resilience. Making sense? A client that picks first one is ignoring the others, can't round robin, etc.

On cache I was thinking about the sidecar's cache, good point that nginx caches these itself. That's good, whew!

On search domains and the rest, I think you may be thinking of the default dns setup that one gets out of the box, but Kubernetes allows you to "roll your own dns" and do literally anything you want in /etc/resolv.conf.  I think that can cause all kinds of trouble. I'm just thinking back to how T-Mobile and other customers set things up there are quite often odd joins between different networks where some things are one way, the rest another way.  The Kubernetes network laws make this a lot better, but I do think DNS could get weird as a way for customers to integrate a Kubernetes cluster with other networks, for example.

dbrinegar:
Oh and we should keep looking to ditch the sidecar if we can. I think one path is to re-write the config using openresty's LUA modules, which appear to have much nicer facilities for the things we want to do as well as a normal dns client built in.

Later

jsravn:
Users can't just do whatever they want with DNS in k8s. They can certainly customize it of course, to some extent. But there are well defined behaviours around k8s DNS resolution that are expected in any functioning cluster.

This is not well documented or spec'd anywhere, unfortunately. I will try to summarize my own understanding from building production k8s clusters and working on the kubelet code.

We use the `ClusterFirst` policy, which requires that search domains are handled correctly. The only way that can happen is for the expected search domains to be present on resolv.conf, and for queries to be forwarded to something which performs DNS routing (to handle split DNS if needed). It is not possible to do split DNS in resolv.conf, so in working setups resolv.conf either points to the cluster DNS service IP, or a local forwarder (like dnsmasq). The former approach being the most common, as it's the easiest to setup and supports split DNS. As the IP in resolv.conf is managed statically (it's a kubelet flag), it needs to be stable - so a service IP or a fixed IP on the localhost. I can't think of any working setup where this could ever have more than a single IP.

kubelet also manages resolv.conf directly on pods, so users are also limited in what they can change in it anyways (they can't just stick whatever they want in resolv.conf). I think, for instance, the search domains are not alterable. I'd need to double check this though.

This doesn't apply to the other DNS policies (like `None`), where you can use the node's resolv.conf which is fully managed by the user. But we're not using that in our case, and we're relying on the search domain behavior.

PR #261, opened 2019-01-25
Fix alertmanager / prometheus integration

jsravn:
This fixes the integration in the direct-DNS solution being used in nginx at the moment.

For https://github.com/lightbend/es-backend/issues/541


jsravn:
Gonna just restore the old sidecar instead.

PR #263, opened 2019-01-28
Add smoke_alert to test generation and issuing of alerts

marcoderama:
The general idea here is to:

- Fire up a pod that's running a netcat server
- Create a ConfigMap for AlertManager that will route all alerts to netcat server
- Tweak AlertManager to use the new ConfigMap
- Create an alerting monitor
- Look for the alerts showing up at the netcat server
- Put things back

This borrows on the existing `smoke_prometheus` test.  The common bits have been moved to `prometheus_common`.

### DIFF
--- a/enterprise-suite/tests/prometheus_common
+++ b/enterprise-suite/tests/prometheus_common
@@ -1,0 +1,85 @@
+# functions common to smoke_prometheus and smoke_alert
+
+deployments_up() {
+  results=$( kubectl get pods -n $NAMESPACE --no-headers | awk '{print $3;}' | grep -v Running | wc -l )
+  [[ results -eq 0 ]]
+}
+
+#
+# prometheus fetchers
+#
+
+ES_CONSOLE=$( busy_wait nodeport )
+PROM="$ES_CONSOLE/service/prometheus"
+ES_MONITOR_API="$ES_CONSOLE/service/es-monitor-api"
+
+prom_query() {
+  curl -fsG "$PROM/api/v1/query" --data-urlencode "query=$*"
+}
+
+prom_has_data() {
+  test_prom_has_data () {
+    results=$( prom_query "$@" | jq '.data.result | length' )
+    ! [[ results -eq 0 ]]
+  }
+  timeout test_prom_has_data "$@"
+  T $? timeseries: "$*"
+}
+
+prom_has_no_data() {
+  test_prom_has_data () {
+    results=$( prom_query "$@" | jq '.data.result | length' )
+    [[ results -eq 0 ]]
+  }
+  timeout test_prom_has_data "$@"
+  T $? timeseries: "$*"
+}
+
+model_exists() {
+  results=$( prom_query "model{name=\"$1\"}" | jq '.data.result | length' )
+  ! [[ results -eq 0 ]]
+}
+
+#
+# log fetchers
+#
+
+kube_state_metrics_logs() {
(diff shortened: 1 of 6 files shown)


enterprise-suite/tests/prometheus_common
```diff
@@ -0,0 +1,85 @@
+# functions common to smoke_prometheus and smoke_alert
+
+deployments_up() {
+  results=$( kubectl get pods -n $NAMESPACE --no-headers | awk '{print $3;}' | grep -v Running | wc -l )
```
marcoderama: NOTE:  This didn't have the `-n $NAMESPACE` bit before.  I think this may have explained some of the flakiness with the original `smoke_prometheus`.  I was having all sorts of flaky behavior with `smoke_alert` until I fixed that.

dbrinegar: good find!!

enterprise-suite/tests/smoke_alert
```diff
@@ -0,0 +1,120 @@
...
+source smokecommon
+source prometheus_common
+
+# 1. Create new configmap for alertmanager deployment
+# 2. Cache existing configmap setting in deployment
+# 3. Update alertmanager deployment configs to use new configmap
+# 4. run tests
+# 5. Put original configmap back in place
+
+export alertmanager_smoke_config=$(mktemp -d -t configmap.XXX)
```
dbrinegar: `alertmanager_smoke_config` --> `alertmanager_smoke_config_dir` would read more clearly for me. looks like things writing the same config file otherwise

marcoderama: Changed.

marcoderama:
Fwiw, either `smoke_prometheus` or `smoke_alert` run successfully in the minikube backend e2e tests.  They just don't work if both are included...  Fixing.  (Thus the change to WIP.)

marcoderama:
Travis tests work.  Open for review again...

dbrinegar:
✅

enterprise-suite/tests/resources/alert-app.yaml
```diff
@@ -0,0 +1,23 @@
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: es-alert-test
+spec:
+  replicas: 2
```
dbrinegar: could be 1 replica?

enterprise-suite/tests/smoke_alert
```diff
@@ -0,0 +1,122 @@
...
+
+busy_wait prom_three_scrapes kube_pod_info
+T $? Prometheus kube-state-metrics has at least three scrapes
+
+PROM_SCRAPES=$( prom_scrapes )
+[[ PROM_SCRAPES -ge 4 ]]
+T $? "Prometheus scraping $PROM_SCRAPES targets"
+
+# app tests
+app_instances='count( count by (instance) (ohai{es_workload="es-alert-test", namespace="'$NAMESPACE'"}) ) == 2'
```
dbrinegar: does the 2 replicas thing here signify anything special in the alert tests? just thinking about pruning down resources, no biggie either way

enterprise-suite/tests/resources/alert-app.yaml
marcoderama: Sure.  Will change.  I just blindly copied `app.yaml`.

enterprise-suite/tests/smoke_alert
marcoderama: Nope.  Changed to one replica.

PR #264, opened 2019-01-29
WIP: kubelet metrics

dbrinegar:
part of lightbend/es-backend#308

This adds a little north of 100 metrics for each kubelet, cardinality ~1000. They wouldn't ever appear in our app's metric picker. Is enough load to consider filtering the collection down to the things we want, especially if you consider that a 100 node cluster may load on the order of 100,000 timeseries from this data source, when all we want are volume stats.

But I see one other issue, which is that `health` is computed for `prometheus_target_down` and `scrape_time` so I think to avoid confusion we'd want to add some filtering to those rules to avoid anything without `es_workload` labels, as these could alert but then not be shown in the UI.

### DIFF
--- a/enterprise-suite/es-monitor-api/prometheus.yml
+++ b/enterprise-suite/es-monitor-api/prometheus.yml
@@ -92,6 +92,34 @@ scrape_configs:
 
 {{ .MonitorTypeRules | indent 6 }}
 
+  # kubelet metrics (not for monitoring, just grafana data)
+  - job_name: 'kubelet'
+    scheme: https
+    tls_config:
+      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+      insecure_skip_verify: true
+    bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+
+    kubernetes_sd_configs:
+      - role: node
+
+    honor_labels: true
+
+    relabel_configs:
+      - action: labelmap
+        regex: __meta_kubernetes_node_label_(.+)
+      - target_label: __address__
+        replacement: kubernetes.default.svc:443
+      - source_labels: [__meta_kubernetes_node_name]
+        regex: (.+)
+        target_label: __metrics_path__
+        replacement: /api/v1/nodes/${1}/proxy/metrics
+      - source_labels: [__meta_kubernetes_node_name]
+        target_label: node_name
+      - source_labels: [__meta_kubernetes_node_address_InternalIP]
+        target_label: node_ip
+
+
   # kube-state-metrics
 
   - job_name: 'kube-state-metrics'


enterprise-suite/es-monitor-api/prometheus.yml
```diff
@@ -92,6 +92,34 @@ scrape_configs:
...
+      - target_label: __address__
+        replacement: kubernetes.default.svc:443
+      - source_labels: [__meta_kubernetes_node_name]
+        regex: (.+)
+        target_label: __metrics_path__
+        replacement: /api/v1/nodes/${1}/proxy/metrics
+      - source_labels: [__meta_kubernetes_node_name]
+        target_label: node_name
+      - source_labels: [__meta_kubernetes_node_address_InternalIP]
+        target_label: node_ip
```
jsravn: Is this the normal way to scrape kubelet? I'm sorta concerned about pulling large amounts of metrics over a single point via the apiserver. If it's a large cluster it could cause performance problems on the apiserver. It would be much better to scrape kubelet directly.

dbrinegar: This is just like cadvisor, as well, so we're already scraping each kubelet through api for those.

I don't know of a way to reduce that point of load. We can drop metrics _after_ we've transfered them through the scrape, but I don't think the scrape endpoint can be parameterized to only print selected metrics.

I'll look into scraping the kubelets directly, but based on quick read this has been tricky across vendors / different security setups. Prometheus Operator and Kubelet code around this has gone around and around on the question, so is a good question but might not have settled answers.

dbrinegar: ref: https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/jsonnet/kube-prometheus/kube-prometheus-insecure-kubelet.libsonnet

dbrinegar: ref: https://github.com/coreos/prometheus-operator/issues/1623

dbrinegar: ref: https://github.com/kubernetes/kubernetes/issues/68522

jsravn:
You'll need to push a commit (e.g. just amend commit and push), to re-trigger on travis-ci.com.

enterprise-suite/es-monitor-api/prometheus.yml
jsravn: Neither of those links are using the apiserver as a proxy. The prometheus-operator issue agrees it is a bad idea. I don't think we should provide this as the out of the box behaviour for customers - if it nukes their apiserver 1) it will be hard for them to debug and 2) it will look really bad on us. If we are doing this for cadvisor, I'd consider that a bug.

If needed, we'll have to ask customers to do extra work as part of the installation setup. If kubelet is being run in a daemonset / service (in some set ups), then it is pretty easy for us to automate.

jsravn: Digging into it, it looks like the coreos prometheus operator actually manages a kubelet `Endpoints` object to work around this. Then prometheus can scrape the list of kubelets directly. https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/cluster-monitoring.md#preparing-kubernetes-components

@longmuir good example of why using core prometheus would be a good idea for providing an out of box operationally ready system.

jsravn: I made https://github.com/lightbend/es-backend/issues/586 for the cadvisor scraping. I would consider this a blocker for operational readiness.

dbrinegar: > Neither of those links are using/suggesting the apiserver as a proxy afaict.

Indeed, and that was not the intention either. Those links are in reference to "I'll look into scraping the kubelets directly"

I've only looked into it enough to realize that there isn't a simple answer. The problem lay in needing permissions to talk to the kubelets directly. Going through the API is the only "out-of-the-box" way to do that kind of thing, as far as I can tell, with the tradeoff being that we're loading the API servers with those proxy requests. The prometheus-operator went through the API initially, so we're essentially trying to "follow them" if that makes sense.

Some new exporter proxy as a daemonset that operates like node-exporter might be workable, but it will incur all the overhead of installing that node-exporter brought, as well as being a new binary we need to build and test. Something kinda like this https://github.com/cofyc/kubelet-exporter.

The kubelet-service idea from prometheus-operator is also interesting, but seems to have been quite fiddly and tricky to get right. But it has been their way of doing this for a few months so definitely worth looking into.

So I think next steps are to figure out the design we want.

dbrinegar: Our current config comes from prometheus docs

https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L53-L58

> # Rather than connecting directly to the node, the scrape is proxied though the
> # Kubernetes apiserver.  This means it will work if Prometheus is running out of
> # cluster, or can't connect to nodes for some other reason (e.g. because of
> # firewalling).

dbrinegar: the array of options one must handle for establishing a direct connection:

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-authentication-authorization/

* anonymous auth
* x509 client cert
* api bearer tokens

as well as various possible auth restrictions, which appear to mirror but not required to be identical to apiserver auth (what we do now).

----------
GKE for example would appear to require a client cert that works against some per-kubelet self-signed cert which I have yet to see anyone describe as workable for Prometheus. But I think it would look something like

```
cat <<EOF | cfssl genkey - | cfssljson -bare server
{
  "hosts": [
    "every.kubelet.here.cluster.local",
    "and-or-ip-addrs"
  ],
  "CN": "my-pod.my-namespace.pod.cluster.local",
  "key": {
    "algo": "ecdsa",
    "size": 256
  }
}
EOF
```
which you then submit to the cluster CertificateSigningRequest, which is then approved... later? possibly manually? Then you download the signed cert and use it.

ref https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/

dbrinegar: This looks promising

https://github.com/kubernetes-incubator/metrics-server

jsravn: Yeah, this is a painful thing to solve. The coreOS approach seems the most reasonable way to me so far. Maybe we can even borrow that. Using the apiserver proxy is good for demo/example purposes, but I wouldn't be comfortable if a customer used it in production. We could merge this PR now, as we're already doing it for cadvisor, but I think we need to fix it pretty soon for the next release (1.1).

PR #266, opened 2019-01-30
Make sure smoke_alert and smoke_prometheus wait for their pods to com…

marcoderama:
…e up.

Wait for the test-specific pods to be up and running before proceeding with the tests.


marcoderama:
Weird.  Reran and all was good.  Included failures unrelated to these changes.

PR #269, opened 2019-01-31
Format everything, add gotests Makefile

jsravn:
- Run gofmt/lint/vet, do this automatically in Travis as well.
- Created a gotests specific Makefile so we can just work in that
directory. As a result I had to rejig the other Makefile/travis.yml.

### DIFF
--- a/enterprise-suite/gotests/Makefile
+++ b/enterprise-suite/gotests/Makefile
@@ -26,7 +26,7 @@ lint:
 
 .PHONY: go-tests
 run-tests:
-	ginkgo -r -- --namespace=$(TILLER_NAMESPACE) --tiller-namespace=$(TILLER_NAMESPACE)
+	ginkgo -r -- --namespace=$(NAMESPACE) --tiller-namespace=$(NAMESPACE)
 
 .PHONY: format
 format:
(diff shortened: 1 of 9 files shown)


enterprise-suite/gotests/Makefile
```diff
@@ -0,0 +1,33 @@
...
+vet:
+	go vet ./...
+
+.PHONY: lint
+lint:
+	golint -set_exit_status $(go list ./...)
+
+.PHONY: go-tests
+run-tests:
+	ginkgo -r -- --namespace=$(TILLER_NAMESPACE) --tiller-namespace=$(TILLER_NAMESPACE)
```
dbrinegar: just double checking that TILLER_NAMESPACE is intended to be used for two namespace things here..

jsravn: It's a little confusing - it should install in the same namespace both things. I'll change it...

enterprise-suite/gotests/Makefile
```diff
@@ -0,0 +1,39 @@
...
+.PHONY: go-tests
+run-tests:
+	ginkgo -r -- --namespace=$(NAMESPACE) --tiller-namespace=$(NAMESPACE)
+
+.PHONY: format
+format:
+	goimports -w $(files)
+
+.PHONY: setup-tools
+setup-tools:
```
srikanthops: can we call setup-tools as part of check

jsravn: setup-tools is quite slow - so I didn't want to do this every time.

PR #274, opened 2019-02-01
WIP - Parallelize specs

jsravn:


marcoderama:
Seems the change is to randomize but the title of this PR says parallelize.  Just a step on the way?

PR #283, opened 2019-02-02
readiness and metrics

dbrinegar:
* use /api/health for readiness (lighter payload, made for kubernetes, doesn't stuff the log with confusing messages)

* scrape metrics

fixes lightbend/es-backend#587

### DIFF
--- a/enterprise-suite/templates/es-grafana-deployment.yaml
+++ b/enterprise-suite/templates/es-grafana-deployment.yaml
@@ -19,6 +19,9 @@ spec:
       labels:
         app: grafana
         component: server
+      annotations:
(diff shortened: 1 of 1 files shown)


enterprise-suite/templates/es-grafana-deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
       labels:
         app: grafana
         component: server
+      annotations:
```
marcoderama: Do we need any new label_configs for these?

dbrinegar: regular pod rules seem to be workin fine

PR #285, opened 2019-02-04
Improve error reporting in failing tests

jsravn:
Also:

* Fix missing test coverage in some places that wasn't ported over.
* Return errors everywhere rather than bool, to ensure tests are
reporting errors.

### DIFF
--- a/enterprise-suite/gotests/util/util.go
+++ b/enterprise-suite/gotests/util/util.go
@@ -193,8 +193,8 @@ const maxRepeats = 30
 const firstSleepMs = 20
 const maxSleepMs = 10000
 
-// Repeatedly runs a function, sleeping for a bit after each time, until it returns true or reaches maxRepeats.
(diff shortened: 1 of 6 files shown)


enterprise-suite/gotests/util/util.go
```diff
@@ -195,11 +194,11 @@ const firstSleepMs = 20
 const maxSleepMs = 10000
 
 // Repeatedly runs a function, sleeping for a bit after each time, until it returns true or reaches maxRepeats.
-// failMsg will be printed in the result error in case of timeout, can be empty.
-func WaitUntilTrue(f func() bool, failMsg string) error {
+func WaitUntilTrue(f func() error) error {
```
mitkus: I think this should be named WaitUntilSuccess now. Also, the comment needs updating.

jsravn: K, will do.

PR #291, opened 2019-02-07
prom AnyData tests

dbrinegar:
AnyData tests if any prometheus data exists in past 10 minutes, to avoid staleness glitches

Uses a new style (prom 2.7) subquery expression to generically wrap any promql


jsravn:
Nice.

PR #292, opened 2019-02-07
Reduce rate periods in static-rules

jsravn:
* Having very large periods can extrapolate single errors for a long
time. This leads to a poor experience when it seems to take a long time
for a monitor to go ready. An example is the akka_http_server_5xx
monitor, which will go orange for 5m+ with only a few 5xx at the start
up of the cluster.
* We should rely on our monitor models for handling spikey data via its
window/confidence mechanism, rather than smoothing things out in static
rules.

It's also not great we have these rules - they are hidden, if a user is
just looking at the UI. We should think of how we can surface this
better. I think they should just be dynamically generated recording
rules, ideally. At least, we should have docs for them.

### DIFF
--- a/enterprise-suite/es-monitor-api/static-rules.yml
+++ b/enterprise-suite/es-monitor-api/static-rules.yml
@@ -30,16 +30,16 @@
   expr: 100 * prometheus_target_sync_length_seconds{quantile="0.99"} / on (job,instance) group_left(interval) prometheus_target_interval_length_seconds{quantile="0.01"}
 
 - record: prometheus_notifications_dropped_rate
-  expr: rate(prometheus_notifications_dropped_total[10m])
+  expr: rate(prometheus_notifications_dropped_total[1m])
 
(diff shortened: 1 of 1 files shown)


craigupson:
which monitor type? isn't the time window the one shown in the monitor edit
form? Or is there an additional time window hidden deep in the model?

dbrinegar:
we could also put a warmup guard on akka_http_5xx rate as it seems we always have initial errors there that should be ignored.  The warmup guard could also be applied to the monitor more generally, but these seems a specific issue with this metric so maybe guard it here.

```
  expr: irate(akka_http_http_server_responses_5xx[1m]) and akka_http_http_server_responses_5xx offset 1m
```

(using 5m for irate to be resilient in the case that user slows down sampling. We want a window big enough to get 2 points, even if say sampling is 1m)

jsravn:
Changed it to irate, please have another look @dbrinegar.

PR #295, opened 2019-02-08
Improve logic around PVCs, PVs, and StorageClass

marcoderama:
Changes for https://github.com/lightbend/es-backend/issues/572.

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -167,7 +167,7 @@ def check_kubectl(minishift=False):
     require_version('kubectl version --client=true --short=true', REQ_VER_KUBECTL)
 
     # Check if kubectl is connected to a cluster. If not connected, version query will timeout.
-    stdout, returncode = run('kubectl version', DEFAULT_TIMEOUT)
+    _, returncode = run('kubectl version', DEFAULT_TIMEOUT)
     if returncode != 0:
         msg = 'Cannot reach cluster with kubectl'
         if minishift:
@@ -264,7 +264,7 @@ def preinstall_check(creds, minikube=False, minishift=False):
         require_version('oc version', REQ_VER_OC)
 
     # Check if helm is set up inside a cluster
-    stdout, returncode = run('helm version', DEFAULT_TIMEOUT)
+    _, returncode = run('helm version', DEFAULT_TIMEOUT)
     if returncode != 0:
         fail('Cannot get helm status. Did you set up helm inside your cluster?')
 
@@ -595,7 +595,7 @@ def check_install(external_alertmanager=False):
         if returncode == 0:
             # Skip first (deployment name) and last (running time) items
             cols = [int(col) for col in stdout.replace('/', ' ').split()[1:-1]]
-            desired, current, up_to_date, available = cols[0], cols[1], cols[2], cols[3]
+            desired, _, _, available = cols[0], cols[1], cols[2], cols[3]
             if desired <= 0:
                 printinfo('failed')
                 printerr('Deployment {} status check: expected to see 1 or more desired replicas, found 0'
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -389,7 +389,7 @@ def check_pv_usage(aboutToUninstall=False, namespace=None):
 
         return
     elif returncode != 0:
-        printerr("Unable to retrieve Persistent Volume info.  Proceed with caution.")
+        printerr("Unable to retrieve get release information.  Proceed with caution.")
         return
 
     hasPVs = 'usePersistentVolumes: true' in stdout
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -335,7 +335,7 @@ def pvsRetainedOrNot():
             # pvc-accccc27-2ef8-11e9-a408-080027c68b7b   32Gi       RWO            Delete           Bound    ariano-console/alertmanager-storage   standard                24h
             claim = args.namespace + '/' + pvc
 
-            match = re.search(r'^([\w-]+)\s+\w+\s+\w+\s+(\w+)\s+Bound\s+('+claim+')\s', stdout, re.MULTILINE)
+            match = re.search(r'^([\w-]+)\s+\w+\s+\w+\s+(\w+)\s+Bound\s+('+claim+r')\s', stdout, re.MULTILINE)
 
             if match:
                 if match.group(2) == 'Retain':
@@ -346,7 +346,7 @@ def pvsRetainedOrNot():
 
     return (retained, notRetained)
 
-def checkPVthings():
+def checkPVthings(aboutToUninstall=False):
     stdout, returncode = run('helm get ' + args.helm_name,
                             DEFAULT_TIMEOUT, show_stderr=False)
     # Don't fail here?...
@@ -358,25 +358,23 @@ def checkPVthings():
     # This assumes the default is true.  What happens if they modify this in values.yaml?  We'll miss that I think.
     wantsPVs = len(filter(lambda x: 'usePersistentVolumes=false' in x, sys.argv)) == 0
 
-    printerr("hasPVs: {}, wantsPVs: {}'".format(hasPVs, wantsPVs))
+    # debug.  remove
+    printerr("hasPVs: {}, wantsPVs: {}".format(hasPVs, wantsPVs))
 
-    # Need to set this dynamically
-    aboutToDelete = False
-
-    if ((not hasPVs and wantsPVs) or (not hasPVs and aboutToDelete)):
+    if ((not hasPVs and wantsPVs) or (not hasPVs and aboutToUninstall)):
         # This case would be typical for a dev/demo.  Chances are we're okay losing the data.
         #    warn user that they will lose their Console data
         #    (Not sure how useful this is really.  Don't think they can (easily) grab the data.)
-        printerr("Given the current and desired configs, continued installation will result in the loss of Console data.")
-        printerr("Stopping.  Invoke again with '--force' to proceed anyway, but save your data first if so desired")
-    elif ((hasPVs and not wantsPVs) or (hasPVs and aboutToDelete)):
+        printerr("WARNING: usePersistentVolumes was false, now true. Continued installation will result in the loss of Console data.")
+        fail("Stopping.  Invoke again with '--nowarn' to proceed anyway, but save your data first if so desired")
+    elif ((hasPVs and not wantsPVs) or (hasPVs and aboutToUninstall)):
         retainedPVs, notRetainedPVs = pvsRetainedOrNot()
 
         if len(notRetainedPVs) > 0:
             printerr("WARNING: Given the current and desired configs, continued installation will result in the loss of Console data.")
             for pv in notRetainedPVs:
                 printerr("info: Reclaim policy for PV {} for claim {} is not 'Retain'".format(pv[0], pv[1]))
-            printerr("Stopping.  Invoke again with '--force' to proceed anyway, but save your data first if so desired")
+            fail("Stopping.  Invoke again with '--nowarn' to proceed anyway, but save your data first if so desired")
 
         if len(retainedPVs) > 0:
             printerr("WARNING: Given the current and desired configs, continued installation will orphan existing Console data.")
@@ -384,7 +382,7 @@ def checkPVthings():
             printerr("See associated documentation at blahdiblah.")
             for pv in retainedPVs:
                 printerr("info: Reclaim policy for PV {} for claim {} is 'Retain'".format(pv[0], pv[1]))
-            printerr("Stopping.  Invoke again with '--force' to proceed anyway.")
+            fail("Stopping.  Invoke again with '--nowarn' to proceed anyway.")
 
 # This is the behavior we'll have to implement to be in line with the thinking of
 # https://github.com/lightbend/es-backend/issues/572
@@ -522,7 +520,6 @@ def install(creds_file):
         status = install_status(args.helm_name)
         if status == 'deployed':
             if args.force_install:
-                # calling uninstall here.  Should warn about PV deletion stuff.  Here or in uninstall?
                 uninstall(status=status)
             else:
                 should_upgrade = True
@@ -539,27 +536,18 @@ def install(creds_file):
         if args.wait:
             helm_args += ' --wait'
 
-        checkPVthings()
+        if not args.nowarn:
+            checkPVthings()
 
         if should_upgrade:
             execute('helm upgrade {} {} {} {} {}'
                 .format(args.helm_name, chart_ref, version_arg,
                         creds_arg, helm_args))
         else:
-            # Marco has commented this out for testing.  We'll have to rewrite this bit.
-            # See new comments above.
-
-            # createPVs = len(filter(lambda x: 'managePersistentVolumes=false' in x, sys.argv)) == 0
-            # if createPVs and are_pvcs_created(args.namespace):
-            #     printerr('info: Found existing PVCs from a previous console installation.')
-            #     printerr('info: Please remove them with `kubectl delete pvc`, or pass --set managePersistentVolumes=false.')
-            #     printerr('info: Otherwise, the install may fail.')
-
             execute('helm install {} --name {} --namespace {} {} {} {}'
                 .format(chart_ref, args.helm_name, args.namespace,
                         version_arg, creds_arg, helm_args))
 
-# add a check for --force here
 def uninstall(status=None):
     if status == None:
         status = install_status(args.helm_name)
@@ -569,6 +557,9 @@ def uninstall(status=None):
     elif status == 'deleting':
         fail('Unable to delete console installation {} - it is already being deleted'.format(args.helm_name))
     else:
+        if not args.nowarn:
+            checkPVthings(aboutToUninstall=True)
+
         printerr("info: deleting previous console installation {} with status '{}'".format(args.helm_name, status))
         execute('helm delete --purge ' + args.helm_name)
         printerr(('warning: helm delete does not wait for resources to be removed'
@@ -743,7 +734,6 @@ def setup_args(argv):
                         action='store_true')
 
     # Install arguments
-    # maybe change (add) this to --force and use in uninstall too.  See common args stuff below
     install.add_argument('--force-install', help='set to true to delete an existing install first, instead of upgrading',
                         action='store_true')
     install.add_argument('--export-yaml', help='export resource yaml to stdout',
@@ -768,6 +758,8 @@ def setup_args(argv):
 
     # Common arguments for install and uninstall
     for subparser in [install, uninstall]:
+        subparser.add_argument('--nowarn', help='ignore warnings about PVs and proceed anyway. CAUTION!',
+                            action='store_true')
         subparser.add_argument('--dry-run', help='only print out the commands that will be executed',
                                action='store_true')
         subparser.add_argument('--helm-name', help='helm release name', default='enterprise-suite')
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -375,7 +375,7 @@ def check_pv_usage(aboutToUninstall=False, namespace=None):
         # This case would be typical for a dev/demo.  Chances are we're okay losing the data.
         #    (Not sure how useful this is really.  Don't think they can (easily) grab the data.)
         printerr("WARNING: usePersistentVolumes was false, now true. Continued (un)installation will result in the loss of Console data.")
-        fail("Stopping.  Invoke again with '--nowarn' to proceed anyway, but save your data first if so desired")
+        fail("Stopping.  Invoke again with '--unbind' to proceed anyway, but save your data first if so desired")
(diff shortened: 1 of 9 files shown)


marcoderama:
Definitely a WIP at this point.  I'm not sure I have the python chops to make the changes I've suggested there...

enterprise-suite/templates/alertmanager-pvc.yaml
```diff
@@ -1,12 +1,12 @@
...
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: alertmanager-storage
-  annotations:
-    "helm.sh/resource-policy": keep
 spec:
+  {{ if or .Values.defaultStorageClass .Values.esAlertmanagerStorageClass }}
   storageClassName: {{ default .Values.defaultStorageClass .Values.esAlertmanagerStorageClass }}
+  {{ end }}
```
jsravn: Do we need to say 
```
{{ else }}
 storageClassName: ""
```

enterprise-suite/templates/es-grafana-pvc.yaml
```diff
@@ -1,12 +1,12 @@
...
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: es-grafana-storage
-  annotations:
-    "helm.sh/resource-policy": keep
 spec:
+  {{ if or .Values.defaultStorageClass .Values.esGrafanaStorageClass }}
   storageClassName: {{ default .Values.defaultStorageClass .Values.esGrafanaStorageClass }}
+  {{ end }}
```
jsravn: Same comment as above.

enterprise-suite/templates/prometheus-pvc.yaml
```diff
@@ -1,12 +1,12 @@
...
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: prometheus-storage
-  annotations:
-    "helm.sh/resource-policy": keep
 spec:
+  {{ if or .Values.defaultStorageClass .Values.prometheusStorageClass }}
   storageClassName: {{ default .Values.defaultStorageClass .Values.prometheusStorageClass }}
+  {{ end }}
```
jsravn: Same comment as above.

enterprise-suite/values.yaml
```diff
@@ -106,12 +104,15 @@ prometheusMemoryRequest: 250Mi
...
-# Default StorageClass to use for persistent volumes
-defaultStorageClass: standard
+usePersistentVolumes: true
+# Default StorageClass to use for persistent volumes.  If unspecified, use the cluster's DefaultStorageClass.
+#defaultStorageClass: standard
+# Application-specific storage classes. If unspecified (and usePersistentVolumes is true), the application-specific
+# PVCs will use the defaultStorageClass setting.
+#prometheusStorageClass:
+#esAlertmanagerStorageClass:
+#esGrafanaStorageClass:
```
jsravn: This is all WIP?

enterprise-suite/scripts/lbc.py
```diff
@@ -284,6 +284,36 @@ def check_resource_list(cmd, expected, fail_msg):
...
+#    or
+#       we're about to "helm delete" and usePersistentVolumes was true
+# then
+#    if PVC reclaim policy is not Retain
+#       warn user that they will lose their Console data
+#    else
+#       warn that their data will not be deleted but they'll have to do some manual steps
+#       - if they want to reuse it, or
+#       - if they actually want to delete it
+#
```
jsravn: I think this pseudo code matches the steps I listed already, right? https://github.com/lightbend/es-backend/issues/572#issuecomment-460725610

enterprise-suite/templates/alertmanager-pvc.yaml
```diff
@@ -1,12 +1,12 @@
-{{ if (and .Values.createAlertManager .Values.managePersistentVolumes) }}
+{{ if (and .Values.createAlertManager .Values.usePersistentVolumes) }}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
   name: alertmanager-storage
-  annotations:
-    "helm.sh/resource-policy": keep
```
jsravn: This makes me pretty uneasy still - this is our only failsafe.

enterprise-suite/templates/alertmanager-pvc.yaml
marcoderama: I think we _don't_ want that.  If we specify `storageClassName: ""`, then we won't bind with the default storage class if such exists.  (https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1)

enterprise-suite/templates/es-grafana-pvc.yaml
marcoderama: Ditto

enterprise-suite/templates/prometheus-pvc.yaml
marcoderama: Ditto

enterprise-suite/values.yaml
marcoderama: Nope.  Everything commented out now.  That will use defaults.  User needs to tweak to use their own storage class(es).

enterprise-suite/scripts/lbc.py
marcoderama: I think to so too.  The only possible caveat is that it is careful to consider the current state as well, not just the desired state.

enterprise-suite/templates/alertmanager-pvc.yaml
marcoderama: Not 100% sure on this one.  Keeping the PVC's around is what made things confusing for us developers.  I think we just need to ensure that `lbc.py` makes the choices clear to normal users.  If they specify `--force-install` then that's on them...  Or not...

jsravn:
@marcoderama I think it'd make sense to split the PRs. Use this PR just for chart changes, then a separate one for the python changes.

marcoderama:
Most things are working pretty well at this point.  I definitely need to automate testing for these changes though...  I haven't finished testing manually so a few more changes are very likely but the fundamentals of the changes aren't likely to change.

marcoderama:
I started testing my stuff with a little shell script (that eventually turned into a bigger shell script).  It exercises various installation scenarios and spits out relevant info as it goes, particularly with respect to PVs and PVCs.  I've uploaded the script and output from a run to https://gist.github.com/marcoderama/3776ecadd94ceb5a37956377cdb09984

It runs through a multistep scenario starting from a dev (re)installing with `emptyDir`, then adding PVs, trying to (re)(un)install on top of that.  The various warnings generated are there to see.  Finally it changes the PVs to have a reclaim policy of Retain and shows how Console can be deleted and then later rewired to those PVs.

The tests/output demonstrate how I think this should work (and is working).

And yeah, I need to put the same kinds of tests in `lbc_test.py`.  I started with what I knew...  'Nuther PR for that.

marcoderama:
Still a couple of tweaks required but it's worthy of removing the WIP now...

enterprise-suite/scripts/lbc_test.py
```diff
@@ -131,15 +131,13 @@ def test_install(self):
         expect_cmd(r'helm repo add es-repo https://repo.lightbend.com/helm-charts')
         expect_cmd(r'helm repo update')
         expect_cmd(r'helm status enterprise-suite', returncode=-1)
-        expect_cmd(r'kubectl get pvc --namespace=lightbend --no-headers')
         expect_cmd(r'helm install es-repo/enterprise-suite --name enterprise-suite --namespace lightbend --devel --values \S+')
-        lbc.main(['install', '--skip-checks', '--creds='+self.creds_file])
+        lbc.main(['install', '--skip-checks', '--nowarn', '--creds='+self.creds_file])
```
marcoderama: This all needs refinement. I just sprinkled in `--nowarn` to allow things to work as is.

mitkus: Maybe a simple improvement could be to extract all the repeating params into a list, since there's a few now:
```
lbc.main(install_common +  ['--export-yaml=console'])
```

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+                    list = retained
+                else:
+                    list = notRetained
+                list.append((match.group(1), match.group(3)))
+
+    return (retained, notRetained)
+
+# This is the behavior as described at
+# https://github.com/lightbend/es-backend/issues/572
+def checkPVthings(aboutToUninstall=False, namespace=None):
```
mitkus: Could use a better name. `guard_pvs`?

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+                            DEFAULT_TIMEOUT, show_stderr=True)
+
+    if (returncode != 0) and not ('Error: release: "{}" not found'.format(args.helm_name) in stderr):
+        printerr("Unable to retrieve PV info.  Proceed with caution.")
+        return
+
+    hasPVs = 'usePersistentVolumes: true' in stdout
+
+    # This assumes the default is true.  What happens if they modify this in values.yaml?  We'll miss that I think.
+    wantsPVs = len(filter(lambda x: 'usePersistentVolumes=false' in x, sys.argv)) == 0
```
mitkus: This pattern is coming up in a few places, let's extract it into a function `get_args_helm_value(name)` or similar. Then we can also later improve it to work better with computed or values.yaml stuff if needed.

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+
+    # debug.  remove
+    printerr("hasPVs: {}, wantsPVs: {}".format(hasPVs, wantsPVs))
+
+    if ((not hasPVs and wantsPVs) or (not hasPVs and aboutToUninstall)):
+        # This case would be typical for a dev/demo.  Chances are we're okay losing the data.
+        #    (Not sure how useful this is really.  Don't think they can (easily) grab the data.)
+        printerr("WARNING: usePersistentVolumes was false, now true. Continued (un)installation will result in the loss of Console data.")
```
mitkus: This message is confusing in case `hasPVs=false` and `aboutToUninstall=true`, because `usePersistentVolumes` is still not true as far as I can tell?

marcoderama: When `hasPVs` is false, it means the data has been stored on non-permanent volumes.  (e.g. `emptyDir`)  If the user does a delete, then they'll lose that data.  Chances are they don't care about it, but I chose not to ignore that case.

mitkus: Sure, that needs a warning message. The part that's confusing me is `usePersistentVolumes was false, now true`, because in case of uninstall (`aboutToUninstall=true`) it is not true.

enterprise-suite/scripts/lbc.py
```diff
@@ -98,15 +98,15 @@ def run(cmd, timeout=None, stdin=None, show_stderr=True):
     finally:
         if timer != None:
             timer.cancel()
-        return stdout, returncode
+        return returncode, stdout, stderr
```
jsravn: Would it be simpler to combine stdout and stderr?

jsravn: Another option to make this less cumbersome is to return a dictionary.

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+
+    if returncode != 0:
+        printerr("Unable to retrieve PV info.  Proceed with caution.")
+    else:
+        for pvc in CONSOLE_PVCS:
+            # We might want to do this by parsing yaml instead...
+            # pvc-accccc27-2ef8-11e9-a408-080027c68b7b   32Gi       RWO            Delete           Bound    lightbend/alertmanager-storage   standard                24h
+            claim = namespace + '/' + pvc
+
+            match = re.search(r'^([\w-]+)\s+\w+\s+\w+\s+(\w+)\s+Bound\s+('+claim+r')\s', stdout, re.MULTILINE)
```
jsravn: Instead of parsing the human friendly output with regex, I think it would be much better to use kubectl's templating system. E.g. you can do something like

```
kubectl get pv -o go-template='{{ range .items }}{{ printf "%s\n" .metadata.name }}{{ end }}'
```

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+        printerr("Unable to retrieve PV info.  Proceed with caution.")
+        return
+
+    hasPVs = 'usePersistentVolumes: true' in stdout
+
+    # This assumes the default is true.  What happens if they modify this in values.yaml?  We'll miss that I think.
+    wantsPVs = len(filter(lambda x: 'usePersistentVolumes=false' in x, sys.argv)) == 0
+
+    # debug.  remove
+    printerr("hasPVs: {}, wantsPVs: {}".format(hasPVs, wantsPVs))
```
jsravn: Remove?

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+
+    if ((not hasPVs and wantsPVs) or (not hasPVs and aboutToUninstall)):
```
jsravn: I'm confused by this - why do we care if `not hasPVs`? In that case there are no PVs, right?

enterprise-suite/scripts/lbc.py
jsravn: This doesn't seem like a bomb proof approach either. It would be much better if we can check the actual parsed arg value - I thought we already use getopts?

enterprise-suite/scripts/lbc.py
jsravn: I don't agree with this. We don't need to warn the user if they are using emptyDir. Their data is already ephemeral! If the pod is bounced for some reason, they will lose their data as well. There is no way for them to save their data, so warnings are pointless.

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+            namespace = args.namespace
+        retainedPVs, notRetainedPVs = pvsRetainedOrNot(namespace)
+
+        stop = False
+
+        if len(notRetainedPVs) > 0:
+            printerr("WARNING: Given the current and desired configs, continued (un)installation will result in the loss of Console data.")
+            for pv in notRetainedPVs:
+                printerr("   info: Reclaim policy for PV {} for claim {} is not 'Retain'".format(pv[0], pv[1]))
+            printerr("Invoke again with '--nowarn' to proceed anyway, but save your data first if so desired")
```
jsravn: I think `nowarn` is the wrong name for this arg - it's too generic. Can we call it `deletedata` or similar instead?

enterprise-suite/scripts/lbc.py
```diff
@@ -284,13 +293,91 @@ def check_resource_list(cmd, expected, fail_msg):
...
+            stop = True
+
+        if len(retainedPVs) > 0:
```
jsravn: I don't think this warning is necessary. It's also bad form to abort the install if it won't actually result in data loss. We only want to abort install as a last resort I think (e.g. if data loss will definitely happen).

enterprise-suite/scripts/lbc_test.py
marcoderama: Good idea.  I'll look at that when I figure out how to add more tests for these changes (in a separate PR).  (A bit of a mystery to me atm.)

dbrinegar:
There's an abundance of feedback here so I hesitate to add more, but it'd be nice to see the test grid in some summary form (the option, the result we want) even if the tests themselves are a separate PR. I found the test and output most helpful. Understanding all the optionality and knock ons of each combination of choices is a little hard, and maybe a small map of those would help shed light on the remaining semantic questions.

enterprise-suite/scripts/lbc.py
marcoderama: Hmmm.  The existing `lambda` code does check the parsed arg value as I understand it (unless you meant matching on `false`, `False`, `0` etc.).  My comment was that this will miss the case where the user changes the value (e.g. `usePersistentVolumes`) in the `values.yaml` (or any such values file) referred to in the helm-specific values passed to `lbc.py`.  For example, instead of specifying

`./lbc.py install --local-chart=.. -- --set usePersistentVolumes=true`
the user does something like
`./lbc.py install --local-chart=.. -- -f setsUsePersistentVolume.yaml`   or the very unlikely (imho)
`./lbc.py install --local-chart=.. -- --set-file usePersistentVolumes=file/with/true`
To be completely general, we'd also have to deal with
`./lbc.py install --local-chart=.. -- --set-string someStringVarWeCareAbout=57`

And now that I think about it, I wonder if we deal properly with the "rightmost value wins" semantics used by helm.

In any case, I just went with the flow for now...

enterprise-suite/scripts/lbc.py
marcoderama: @mitkus: I understand what you're saying now.  Good point.  I'll fix if this whole case (warning on loss of ephemeral data) doesn't disappear...

@jsravn: I chose to call out the case of deleting ephemeral data because it might still catch someone off-guard.  (Say you've been running the console all day while working on some prometheus bug.)  I can be convinced otherwise on that though.

enterprise-suite/scripts/lbc.py
marcoderama: See comment below.  (https://github.com/lightbend/console-charts/pull/295/files#r257423196)

enterprise-suite/scripts/lbc.py
marcoderama: Yeah...  I agree it needs changing but to what is the question?  Even if the user passes this flag, the data isn't necessarily deleted (e.g. with PVs that have a reclaim policy of Retain it's just unbound).  How about... `--unbind`?

I'll go with that for now but am open to suggestions.  (`--imperildata`?)

enterprise-suite/scripts/lbc.py
marcoderama: I'll change this to only spit out the warning and not abort the install.  I think it's important to let the user know about the orphaned data so they can decide what to do about it.

enterprise-suite/scripts/lbc.py
marcoderama: I'd rather not combine them.  Gonna leave alone for now but I agree it could be improved.

enterprise-suite/scripts/lbc.py
marcoderama: Yeah.  Figured there was a better way but my Go template foo is yellow-belt at best.  Changed.

marcoderama:
I've updated the gist files showing the test run against the most recent changes:
https://gist.github.com/marcoderama/3776ecadd94ceb5a37956377cdb09984

enterprise-suite/scripts/lbc.py
jsravn: This is going to be a huge pain though - even in dev mode with emptyDir we'll need to pass in `nowarn`? I think it's taking things too far. We only need to worry about users actually wanting to persist their data with `--usePersistentVolumes`.

enterprise-suite/scripts/lbc.py
jsravn: I think `unbind` is similarly vague. I'm not even sure what that means as a developer - unbind what? I think it should be related somehow to the data being deleted. `--delete-pvc` or something that is very explicit.

enterprise-suite/scripts/lbc.py
jsravn: sgtm

enterprise-suite/scripts/lbc.py
marcoderama: I chose `unbind` cuz all the serious warnings are related to unbinding the PVC and PV.  That said, I'm not super-crazy about the name either.  `unbindPVs`?

marcoderama:
@dbrinegar et al.:  I've moved some of the discussion of the data states back to the issue.  (
https://github.com/lightbend/console-backend/issues/572#issuecomment-464966786)  I think that addresses your desire to better understand the combination of choices.

@jsravn et al.:  Afaict, the only significant difference of opinion at this point is whether to warn for transitions away from ephemeral data.  And not that significant as I'm okay if we don't warn in those cases (but also think devs would be fine to add `--unbind` when invoking `lbc.py`).  Let's resolve in the issue and then I'll make the code here agree.

enterprise-suite/scripts/lbc.py
mitkus: `--risk-data` ?

jsravn: What's wrong with `--delete-pvcs` - that describes what we're doing exactly.

marcoderama:
- Changed `unbind` to `delete-pvcs`
- Added warnings around orphaning data (transitions 16 and 18 in diagram)

marcoderama:
I've also updated the gist with the test script and output using the latest code.
https://gist.github.com/marcoderama/3776ecadd94ceb5a37956377cdb09984

PR #296, opened 2019-02-08
Add checksum annotations to guarantee pod restart when configmap changes

srikanthops:
lightbend/es-backend/issues/564
Verified pods restarts by changing the following 
es-grafana-configmap-datasource.yaml
es-grafana-configmap-datasource.yaml
es-grafana/akka-exporter.json
alertmanager/alertmanager.yaml 
es-monitor-api/default-monitors.json

### DIFF
--- a/enterprise-suite/templates/prometheus-deployment.yaml
+++ b/enterprise-suite/templates/prometheus-deployment.yaml
@@ -19,7 +19,6 @@ spec:
       app: prometheus
       component: server
   template:
-
     metadata:
       labels:
         app: prometheus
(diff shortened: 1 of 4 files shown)


dbrinegar:
Can you add a rule in prometheus to drop these, so we don't have checksum hashes in all the data? The idea is: we should filter out annotations that are not human readable. We already have template hash distinction tags, so don't need these too.

enterprise-suite/templates/prometheus-deployment.yaml
```diff
@@ -19,12 +19,15 @@ spec:
       app: prometheus
       component: server
   template:
+
```
dbrinegar: please remove this extraneous whitespace

PR #300, opened 2019-02-12
Fix Makefile to work with enterprise-suite-latest

marcoderama:
Also removes tiller install from `install-dev` step so useable in non-minkube context.


jsravn:
Can we remove latest?

marcoderama:
Yeah, I wondered if we're ready to remove latest now but I have no idea who might still be depending on it...  I just thought I'd run `make install-dev` on it to get a local install with the latest packages.  It's easy enough to do that from the `enterprise-suite` directory though.  But I 'm not sure what other people are doing with it, if anything.

jsravn:
@marcoderama I've brought it up a few times in Slack, and there were no objections. Could ask again in standup though.

PR #301, opened 2019-02-12
[WIP] Use container with all prerequisites to speed up travis tests

mitkus:


.travis.yml
```diff
@@ -28,23 +26,34 @@ jobs:
...
   - name: Backend E2E Tests - Openshift
     script:
+    - travis_retry scripts/setup-tools-for-ubuntu.sh
     - travis_retry scripts/setup-openshift.sh ${OC_TOKEN}
     - make -C enterprise-suite openshift-backend-tests
 
   - name: Backend E2E Go Tests - Minikube
+    before_script:
+    - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin ${DOCKER_REGISTRY}
+    - travis_retry docker pull ${DOCKER_REGISTRY}/${GOTESTS_CONTAINER}
```
jsravn: Can we call it BUILD_CONTAINER (idea being we can use it for the frontend tests as well, it's not gotests specific).

.travis.yml
```diff
@@ -28,23 +26,34 @@ jobs:
 
   - name: Backend E2E Tests - Openshift
     script:
+    - travis_retry scripts/setup-tools-for-ubuntu.sh
     - travis_retry scripts/setup-openshift.sh ${OC_TOKEN}
     - make -C enterprise-suite openshift-backend-tests
 
   - name: Backend E2E Go Tests - Minikube
+    before_script:
+    - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin ${DOCKER_REGISTRY}
```
jsravn: Why not just put the build container in the public bintray repo? Does it really need to be private? I think the build container repo could be public too, like console-charts is.

scripts/console-test-minikube.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
```
jsravn: Please use `set -eu`.

scripts/console-test-openshift.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
jsravn: Please use `set -eu`.

scripts/console-test-openshift.sh
```diff
@@ -0,0 +1,5 @@
...
+oc login https://centralpark.lightbend.com --token=$1
+oc status && oc project && oc get pods
+oc GO111MODULE=on make -C enterprise-suite gotests-openshift NAMESPACE=console-backend-go-tests
```
jsravn: eol

scripts/console-test-minikube.sh
```diff
@@ -0,0 +1,6 @@
...
+cd console-charts
+minikube start --vm-driver=none
+minikube addons enable ingress
+GO111MODULE=on make -C enterprise-suite install-helm gotests-minikube
```
jsravn: eol

.travis.yml
mitkus: I did make the repo public initially but then got an email from @JustinPihony about that. The policy is to make everything private unless there's a good reason for it to be public.

scripts/console-test-minikube.sh
mitkus: Done

scripts/console-test-openshift.sh
mitkus: Done

scripts/console-test-openshift.sh
mitkus: Done

scripts/console-test-minikube.sh
mitkus: Done

mitkus:
I'm calling this experiment over with a negative result, dockerized tests are not faster at all:

Dockerized with image cache: 9 min 26 sec
Dockerized with pull everytime: 8 min 7 sec
Old way: 8 min 12 sec

Part of the problem is that image with all the dependencies is 1.5 GB and still has to be pulled. Smarter caching could improve this marginally, but Travis still downloads the cache from somewhere.

PR #303, opened 2019-02-13
Don't cleanup by default, add StartAsync/StopAsync

jsravn:
Preliminaries for the smoke_proxy test conversion.

### DIFF
--- a/enterprise-suite/gotests/testenv/testenv.go
+++ b/enterprise-suite/gotests/testenv/testenv.go
@@ -61,14 +61,8 @@ func InitEnv() {
 		Expect(err).To(Succeed(), "new k8sclient")
 	}
 
-	// In case helm release or PVCs were left from the previous run - clean them up now
-	anyPVCFound := false
-	for _, pvc := range consolePVCs {
-		if kube.PvcExists(K8sClient, args.ConsoleNamespace, pvc) {
-			anyPVCFound = true
-		}
-	}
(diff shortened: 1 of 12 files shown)


enterprise-suite/gotests/testenv/testenv.go
```diff
@@ -154,12 +155,11 @@ func cleanup(allowFailures bool) {
...
-	if foundStorageClass {
-		// Delete PVCs which are left after helm uninstall
-		for _, pvc := range consolePVCs {
-			if err := kube.DeletePvc(K8sClient, args.ConsoleNamespace, pvc); err != nil && !allowFailures {
-				Expect(err).To(Succeed(), "delete PVCs")
-			}
+	// Delete PVCs which are left after helm uninstall
+	for _, pvc := range consolePVCs {
+		if err := kube.DeletePvc(K8sClient, args.ConsoleNamespace, pvc); err != nil && !allowFailures {
+			Expect(err).To(Succeed(), "delete PVCs")
```
mitkus: Not sure this will work on local minishift. It doesn't have any storage classes, so tests will install `emptyDir` version of Console. Then here it will try to remove PVCs which were not created.

jsravn: It seemed to work without any PVCs in place. But I'll check again - it could also check it exists before deletion.

jsravn: Addressed it.

PR #304, opened 2019-02-14
Install script out of date/srikanth

srikanthops:
Compare the content of the current installer with latest installer.
Warn if contents don't match

lightbend/console-backend/issues/399


enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
...
+
+url_http_timeout=3 # 3 seconds
+installer_url="https://raw.githubusercontent.com/lightbend/console-charts/master/enterprise-suite/scripts/lbc.py"
+warning_display_timeout=2 # 2 seconds
+
+# compare the contents of current running installer and remote installer.
+# if they are different print warning a new installer is present
+def check_new_install_script():
+    try:
+        response=url.urlopen(installer_url, timeout=url_http_timeout)
```
mitkus: We had problems with Python TLS versions on Mac OS when using urllib. It can be a part of separate PR, but it's probably worth to extract http request function that tries to use curl and then urllib2 as a fallback, like what check_credentials does.

enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
...
+
+    # read the contents of the current installer
+    with open(os.path.abspath(__file__)) as f:
+        current_installer_contents = f.read()
+
+    if remote_installer_contents != current_installer_contents:
+        printinfo("\nThe installer you are running is out of date...")
+        printinfo("New installer is available, use the following command to download it")
+        printinfo (" curl -O " + installer_url + "\n")
+        time.sleep(warning_display_timeout)
```
mitkus: What does this sleep accomplish here?

enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
...
     finally:
-        return success 
+        return success
 
-def preinstall_check(creds, minikube=False, minishift=False):
+
+
+url_http_timeout=3 # 3 seconds
+installer_url="https://raw.githubusercontent.com/lightbend/console-charts/master/enterprise-suite/scripts/lbc.py"
+warning_display_timeout=2 # 2 seconds
```
mitkus: Global constants like these should be in uppercase, just to be consistent with existing code. However, since they are only used in check_new_install_script() and not something that is expected to be changed, maybe they don't have to be global and can be limited to the function scope.

enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
...
+    with open(os.path.abspath(__file__)) as f:
+        current_installer_contents = f.read()
+
+    if remote_installer_contents != current_installer_contents:
+        printinfo("\nThe installer you are running is out of date...")
+        printinfo("New installer is available, use the following command to download it")
+        printinfo (" curl -O " + installer_url + "\n")
+        time.sleep(warning_display_timeout)
+
+def preinstall_checkpreinstall_check(creds, minikube=False, minishift=False):
```
mitkus: This name `preinstall_checkpreinstall_check` looks like an editor mishap!

enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
...
     check_helm()
     check_kubectl()
+    check_new_install_script()
```
mitkus: No need to check it here if it's also at the start of main()

enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
         else:
             printerr('error: check_credentials failed: {}'.format(err))
     finally:
-        return success 
+        return success
 
-def preinstall_check(creds, minikube=False, minishift=False):
+
+
+url_http_timeout=3 # 3 seconds
```
mitkus: Probably good idea to have http request timeout, but then it should also be applied in check_credentials().

enterprise-suite/scripts/lbc.py
```diff
@@ -653,6 +683,8 @@ def main(argv):
     global args
     args = setup_args(argv)
 
+    check_new_install_script()
```
mitkus: This should be conditional on `skip_checks` flag, or it will brake lbc_tests.py otherwise.

enterprise-suite/scripts/lbc.py
```diff
@@ -214,11 +215,40 @@ def check_credentials(creds):
...
+        # if we cannot connect to remote server, ignore for now...
+        return
+
+    # read the contents of the current installer
+    with open(os.path.abspath(__file__)) as f:
+        current_installer_contents = f.read()
+
+    if remote_installer_contents != current_installer_contents:
+        printinfo("\nThe installer you are running is out of date...")
+        printinfo("New installer is available, use the following command to download it")
```
marcoderama: How about
A new installer is available. Use the following command to download it.

PR #305, opened 2019-02-15
Port forwarding test in go

mitkus:
For https://github.com/lightbend/console-backend/issues/556

This PR also turns on Travis caching of go modules and build cache, it saves ~1-2 minutes of build time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,10 @@ jobs:
     - make -C enterprise-suite openshift-backend-tests
 
   - name: Backend E2E Go Tests - Minikube
+    cache:
+      directories:
+        - $HOME/.cache/go-build
+        - $HOME/gopath/pkg/mod
     script:
     - travis_retry scripts/setup-minikube-for-linux.sh
     - GO111MODULE=on make -C enterprise-suite install-helm gotests-minikube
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -32,13 +32,21 @@ jobs:
     - make -C enterprise-suite openshift-backend-tests
 
   - name: Backend E2E Go Tests - Minikube
+    cache:
+      directories:
+        - $HOME/.cache/go-build
+        - $HOME/gopath/pkg/mod
```
jsravn: What does this accomplish? If it happens in parallel, I don't think anything is shared.

enterprise-suite/gotests/tests/portforward/portforward_test.go
```diff
@@ -0,0 +1,71 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+const consoleRemotePort = 8080
+const consoleDeployment = "deployment/es-console"
+
+// Hardcode the local port because with random port setup it's currently
+// tricky to get cmd.StartAsync stdout without stopping it first.
+const localPort = 58363
```
jsravn: Can you copy the logic at https://github.com/sky-uk/merlin/blob/master/e2e/e2e.go#L257?

We can stick it in util somewhere. It provides the ability to discover an open port on the local machine.

.travis.yml
mitkus: These get cached between consecutive builds. Before running the script the contents in these directories will be restored to the state that was after a previous run. So all the go dependency modules don't have to be downloaded & compiled again.

enterprise-suite/gotests/tests/portforward/portforward_test.go
mitkus: Ok, added utility function `FindFreePort`

PR #306, opened 2019-02-18
Add apiserver proxy tests

jsravn:
With some additional util improvements:
- Log if the install is taking a while, so it's clear what's going on.
- Make WaitUntilSuccess more configurable, so we don't wait excessively
on failures. Also reduce the max wait to 70s for long waits, from 4
minutes.
- Remove error from FindFreePort - just panic if it fails (which it shouldn't ever).

### DIFF
--- a/enterprise-suite/gotests/testenv/testenv.go
+++ b/enterprise-suite/gotests/testenv/testenv.go
@@ -97,7 +97,7 @@ func InitEnv() {
 		for {
 			select {
 			case <-ticker.C:
-				if _, err := fmt.Fprintf(os.Stderr, "lbc.py installation is still ongoing [%fs]...", time.Now().Sub(start).Seconds()); err != nil {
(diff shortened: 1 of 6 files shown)


jsravn:
Build broken atm because of https://github.com/lightbend/console-charts/pull/307.

enterprise-suite/gotests/testenv/testenv.go
```diff
@@ -87,6 +89,22 @@ func InitEnv() {
 		}
 	}
 
+	// Create an async goroutine to report when install is taking a while.
+	ticker := time.NewTicker(15 * time.Second)
```
mitkus: Not very important, but maybe this should be longer than 15 seconds? On my underpowered mac laptop minikube lbc install always takes a minute or so, so this would print repeated messages every time.

PR #309, opened 2019-02-20
Non-released version includes '-next' suffix

marcoderama:
These changes are a possible interim solution for https://github.com/lightbend/console-backend/issues/594.

These changes leave the existing release process alone but make non-release versions have a version with suffix `-next`.  e.g. `1.0.1-next`.

Tested release scripts locally.  This resulted in a version of `1.0.1`.  Also installed HEAD locally.  That had a version of `1.0.1-next`.

### DIFF
--- a/scripts/export-chart-version.sh
+++ b/scripts/export-chart-version.sh
@@ -9,5 +9,4 @@ rel_dir=$( dirname $script_dir )
 chart_dir=$rel_dir/$1
 
 cd $chart_dir
-# Strip off any "-next" suffix
-yq r Chart.yaml version | sed 's/\(.*\)-next/\1/'
+yq r Chart.yaml version
--- a/scripts/export-chart-version.sh
+++ b/scripts/export-chart-version.sh
@@ -9,4 +9,5 @@ rel_dir=$( dirname $script_dir )
 chart_dir=$rel_dir/$1
 
 cd $chart_dir
-yq r Chart.yaml version
+# Strip off any "-next" suffix
+yq r Chart.yaml version | sed 's/\(.*\)-next/\1/'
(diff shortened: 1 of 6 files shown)


scripts/export-chart-version.sh
```diff
@@ -9,4 +9,5 @@ rel_dir=$( dirname $script_dir )
 chart_dir=$rel_dir/$1
 
 cd $chart_dir
-yq r Chart.yaml version
+# Strip off any "-next" suffix
+yq r Chart.yaml version | sed 's/\(.*\)-next/\1/'
```
jsravn: I don't think this should be stripping next - this script should export the version as-is. Since the stripping is for the release, we should do it in make-release.sh.

scripts/make-release.sh
```diff
@@ -63,12 +64,22 @@ git tag -a $git_tag -m "Release $git_tag"
...
+# Increment last number-only component and add '-next' if not already there.
+# e.g. 1.2.3            -->  1.2.4-next
+#      1.2.3-rc.4       -->  1.2.3-rc.5-next
+#      1.2.3-rc.4-next  -->  1.2.3-rc.5-next
 cd $chart_dir
-prefix=${version%.*}
-last_component=${version##*.}
-((last_component++))
-next_version="$prefix.$last_component"
+next_version=$( echo $version | sed -E -e 's/(.*\.)([0-9]+)(\.([[:alnum:]]*[[:alpha:]-]+[[:alnum:]-]*))?/\1 \2 \3/' \
```
jsravn: Is there a simpler way to do this? If not... it needs a lot of comments to explain it :).

scripts/export-chart-version.sh
marcoderama: I really didn't like that change either.  Fixed.  Other things thus simplified.

scripts/make-release.sh
marcoderama: Probably, but documented...

PR #312, opened 2019-02-21
Add redirect to trailing slash

jsravn:
This change checks for the presence of a single level path without a
trailing slash, and then redirects it.

This should work in all cases, because we already require that all
application routes have specific location blocks to handle them.

Also some other small changes:
- Added anchors to all the regex matches, as recommended by nginx best
practices. This just makes sure we don't match things we shouldn't.
- Dropped 'index index.html' in the final catch-all location block. I
couldn't figure out the point of this, and it didn't seem to have any
effect removing it.

For https://github.com/lightbend/console-backend/issues/598

### DIFF
--- a/enterprise-suite/templates/es-console-configmap.yaml
+++ b/enterprise-suite/templates/es-console-configmap.yaml
@@ -44,12 +44,18 @@ data:
 
       # proxy_pass backends with dynamic base url
 
+      # Location matching strategy:
+      # - Explicit location matchers for each route.
+      # - Anything not matched by the explicit location matchers is treated as a path prefix (base href), added by
+      #   e.g. ingress or apiserver proxy.
+      # - Catch all location matcher is last, which captures the path prefix up to the penultimate slash.
+
       sub_filter_once off;
 
       set $nocache 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
 
       # prometheus UI and api endpoints
-      location ~ (.*/service/prometheus)(/.*) {
+      location ~ ^(.*/service/prometheus)(/.*)$ {
         # encode base path
         array_split '/' $1 to=$encoded_base;
         array_map_op set_escape_uri $encoded_base;
@@ -63,13 +69,13 @@ data:
       }
 
       # es-monitor-api
-      location ~ (.*/service/es-monitor-api)(/.*) {
+      location ~ ^(.*/service/es-monitor-api)(/.*)$ {
         proxy_pass http://$monitorapi$2$is_args$args;
         proxy_redirect '/' ' $1/';
       }
 
       # grafana plugin
-      location ~ (.*/service/grafana)(/dashboard/script/exporter-async.js) {
+      location ~ ^(.*/service/grafana)(/dashboard/script/exporter-async.js)$ {
         # encode base path
         array_split '/' $1 to=$encoded_base;
         array_map_op set_escape_uri $encoded_base;
@@ -90,7 +96,7 @@ data:
       }
 
       # grafana ui
-      location ~ (.*/service/grafana)(/.*) {
+      location ~ ^(.*/service/grafana)(/.*)$ {
         # encode base path
         array_split '/' $1 to=$encoded_base;
         array_map_op set_escape_uri $encoded_base;
@@ -105,7 +111,7 @@ data:
       }
 
       # alertmanager ui
-      location ~ (.*/service/alertmanager)(/.*) {
+      location ~ ^(.*/service/alertmanager)(/.*)$ {
         proxy_pass http://$alertmanager$2$is_args$args;
         proxy_redirect '/' ' $1/';
       }
@@ -114,13 +120,13 @@ data:
       # redirect /service/grafana -> /service/grafana/
       # redirect /service/alertmanager -> /service/alertmanager/
 
-      location ~ (.*/service/(prometheus|grafana|alertmanager))$ {
+      location ~ ^(.*/service/(prometheus|grafana|alertmanager))$ {
         return 301 ' $1/';
       }
 
       # version
 
-      location ~ (.*)/version$ {
+      location ~ ^(.*)/version$ {
         return 200 '{
           "description": {{ .Chart.Description | quote }},
           "version": {{ .Chart.Version | quote }},
@@ -142,7 +148,7 @@ data:
 
       # console entry point
 
-      location ~ (.*/)index.html {
+      location ~ ^(.*/)index.html$ {
         # Console CSP
         add_header Content-Security-Policy $app_csp;
         # encode base path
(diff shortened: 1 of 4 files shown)


dbrinegar:
I think with this design you'll have to encode all of the App routes so they aren't redirected by mistake.

For example  "/cluster" should not go to "/cluster/"

enterprise-suite/templates/es-console-configmap.yaml
```diff
@@ -206,16 +212,22 @@ data:
...
+      # Specific handler for assets.
+      location ~ ^(.*)(/assets/.*)$ {
         alias $base$2;
       }
 
-      location ~ (.*)(/.*) {
+      # If there is a single level path with no trailing slash, redirect to trailing slash.
+      # E.g. "http://consoleurl/monitoring" -> 301 to "http://consoleurl/monitoring/".
+      location ~ ^/([^/]+)$ {
+        return 301 $scheme://$host$request_uri/;
```
dbrinegar: I don't think $host or $scheme always works and recommend just using the scheme and host from the regex to be consistent with all the others.

To test this I think you want to load through https --> http and if remembering the other one is a localhost proxy.

The idea is: we want to use the URI "as-is" and not try to rebuild it from the limited point of view of our inner nginx.

dbrinegar:
I like the idea here, am going to poke a bit at it to see if I can show the issues above more clearly

dbrinegar:
test:

1. install in minikube
2. visit http://192.168.99.100:30080/cluster

drops "port" and redirects to "/cluster/"

image

dbrinegar:
test:

1. install in minikube
2. run `kubectl proxy`
3. visit http://127.0.0.1:8001/api/v1/namespaces/lightbend/services/console-server:80/proxy/

loads index.html but everything else is routed through wrong path

image


4. visit http://127.0.0.1:8001/api/v1/namespaces/lightbend/services/console-server:80/proxy/cluster

redirects to  127.0.0.1/cluster

image

dbrinegar:
I suppose we should add some smoke tests for these examples :-|

dbrinegar:
em... just a note on Chrome at least:  once you've got the above unworkable links in cache, they're hard to shed. Going from this branch back to master is a little sticky, have to clear Chrome cache to see that the above links do indeed work in master.

enterprise-suite/templates/es-console-configmap.yaml
jsravn: The regex doesn't know the scheme though. How do you propose figuring it out? Good point about reverse proxies, perhaps we should just do an internal rewrite here.

jsravn:
Given `/cluster` is the only thing we need to be able to handle, I don't see a big deal in adding it. I don't expect the UI to be adding loads of new routes - if they do, it's a simple matter to add a new location.

jsravn:
I'll take a look at the other failures.

jsravn:
Okay @dbrinegar, have another look. I believe I've fixed all the scenarios you posted above.

1. Changed redirect to a rewrite, so we don't need to worry about schema, etc. in the face of reverse proxies
2. Added an explicit `location` handler for the SPA root assets `/(cluster|.*\.js|.*\.css|.*\.woff|.*\.woff2|.*\.ttf)`.

A tricky aspect is the SPA doesn't have a predictable location for its js, css, and font files - these all get loaded at the root base href. A nice to have is if the SPA would load all these things from a subpath, like `/assets/` - then we could easily create a generic `location` that handles anything loaded from that directory, versus having to look for specific file patterns as I did here. WDYT @pcraddick?

jsravn:
Note all the >1 level routes are already explicitly handled (e.g. `/namespaces`) - so I think cluster is the only thing to take care of.

enterprise-suite/templates/es-console-configmap.yaml
dbrinegar: `return 301 /$1/;` would work I think?

dbrinegar: oh if there's an nginx behavior that auto fills a relative url, might have to do the leading space config trick which we use elsewhere. nginx is chockablock full of implicit behavior, but the idea would be to return a relative path 301 that just adds the backslash to whatever it was in the first place

leading space trick to avoid auto-fill is like line 124 (look for the other 301)

`return 301 " /$1/";

dbrinegar:
I'll have to load this up and poke around at things. Will try to add some quick tests as suitable.


Thinking longer term (not in this PR, but as a consequence of this design to handle all the app routes):

It would be quite nice to organize the boundary with App Routes so it is simpler and following some more bullet proof pattern.

The other thing we can do to reduce copy paste for each new Route is to rewrite this in lua, which would give us things like a list of routes and a single general purpose handler

enterprise-suite/templates/es-console-configmap.yaml
```diff
@@ -206,16 +212,28 @@ data:
...
       }
 
+      # Specific handler for assets.
+      location ~ ^(.*)(/assets/.*)$ {
+        alias $base$2;
+      }
 
-      # static files
+      # Handler for application routes and assets not inside of /assets.
+      location ~ ^(.*)(/(cluster|.*\.js|.*\.css|.*\.woff|.*\.woff2|.*\.ttf))$ {
```
dbrinegar: the "/cluster" route needs to be handled like the other application routes or else it won't have the right headers, cache control, security handlers, etc. Unfortunately I think we need another copy / paste section of the chunky section copy pasted in this config...

jsravn:
We really should have tests for all these things. It's bad that we can break it without realizing it, esp. given the complexity of the nginx config it seems easy to make a mistake.

enterprise-suite/templates/es-console-configmap.yaml
jsravn: I don't think any of this has changed - do you mean it's already broken? This should rewrite `/cluster` to the `index.html` location block which will provide the correct headers.

jsravn: It looks okay to me - please clarify:

```
$ curl -sD - http://192.168.99.116:30080/cluster
HTTP/1.1 200 OK
Date: Mon, 25 Feb 2019 09:55:04 GMT
Content-Type: text/html
Transfer-Encoding: chunked
Connection: keep-alive
X-Frame-Options: DENY
X-XSS-Protection: 1
Content-Security-Policy: default-src 'self' 'unsafe-eval' 'unsafe-inline';
Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0
```

jsravn:
@dbrinegar can you have another look? The security headers look fine to me at `/cluster`.

jsravn:
I've added some test coverage now as well.

enterprise-suite/templates/es-console-configmap.yaml
dbrinegar: Ah right this will flow into the index.html handler. Now I'm wondering if we can do the same for the copy/paste sections (next time)

jsravn:
ping @dbrinegar

jsravn:
Going to merge this since it's been waiting for re-review for a week now, and I really want to get it into 1.0.1. We can tackle any follow up issues in a separate PR.

PR #313, opened 2019-02-22
Print debug info in gotests and add k8s annotations

jsravn:
* Adds the standard kubernetes annotations, useful for the tests to select the correct pods.
* Print debug info when lbc.py fails to install in gotests.
* Reduce install timeout to something more reasonable (2 mins from 6 mins).

### DIFF
--- a/enterprise-suite/templates/alertmanager-deployment.yaml
+++ b/enterprise-suite/templates/alertmanager-deployment.yaml
@@ -19,11 +19,18 @@ spec:
     matchLabels:
       app: prometheus
       component: alertmanager
+
   template:
     metadata:
       annotations:
         checksum/alertmanager-config: {{ (.Files.Glob "alertmanager/*").AsConfig | sha256sum }}
       labels:
+        app.kubernetes.io/name: {{ template "name" . }}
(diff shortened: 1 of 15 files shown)


jsravn:
Should merge this since the commits are separate.

enterprise-suite/templates/alertmanager-deployment.yaml
```diff
@@ -19,11 +19,18 @@ spec:
...
       app: prometheus
       component: alertmanager
+
   template:
     metadata:
       annotations:
         checksum/alertmanager-config: {{ (.Files.Glob "alertmanager/*").AsConfig | sha256sum }}
       labels:
+        app.kubernetes.io/name: {{ template "name" . }}
+        helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
```
marcoderama: Just curious why the need for the `replace` bit?

jsravn: This is copied right from the helm docs on what to use. But it's because you can't use `+` in label values I believe, and `+` is a valid semver character.

PR #314, opened 2019-02-22
Add img-src in CSP rule

katsutoxin:
connect to https://github.com/lightbend/console-frontend/issues/572

Background:
after we set up ingress, we found that the down-arrow image is missing. 
The fix is to bundle down-arrow image in js file
However, CSP complains
screen shot 2019-02-22 at 11 33 46 am

This PR is to fix the CSP. 
 down-arrow works with the fix and ingress url (check screenshot)

screen shot 2019-02-22 at 1 46 10 pm

### DIFF
--- a/enterprise-suite/templates/es-console-configmap.yaml
+++ b/enterprise-suite/templates/es-console-configmap.yaml
@@ -15,7 +15,7 @@ data:
       more_set_headers
         "X-Frame-Options: DENY"
         "X-XSS-Protection: 1";
-      set $app_csp "default-src 'self' 'unsafe-eval' 'unsafe-inline';";
+      set $app_csp "img-src 'self' data:; default-src 'self' 'unsafe-eval' 'unsafe-inline';";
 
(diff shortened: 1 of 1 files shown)


dbrinegar:
Why is this the only image we're delivering this way, and why can't it be delivered like all the other images? I'd rather we did things consistently, one way or the other.

Thanks for explaining the rationale and plan forward.

katsutoxin:
> Why is this the only image we're delivering this way, and why can't it be delivered like all the other images? I'd rather we did things consistently, one way or the other.
> 
> Thanks for explaining the rationale and plan forward.

Because arrow-down.png is linked in `.css` file (NOTE: we use arrow-down as the drop down symbol and it only can set in css, it is a "style" setting for drop down selection).  ..   all other images are linked in `.html` file, they are html component and not the style.

Angular use webpack to link/load css and html files differently.

dbrinegar:
Ah, so Webpack is making this a src ref and bundling in data: uri, while before it was an absolute uri so wasn't transformed this way. Got it

PR #315, opened 2019-02-25
Only update whitesource reports on master

jsravn:
So we don't overwrite the reports on a PR.

part of https://github.com/lightbend/console-backend/issues/569

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,6 @@ jobs:
     script:
     - travis_retry scripts/setup-minikube-for-linux.sh
     - make -C enterprise-suite minikube-backend-tests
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -93,6 +90,16 @@ jobs:
     script:
     - make test CHARTS=reactive-sandbox
 
+  - name: Console Whitesource Report
+    # Only deploy for pushes to master (and if manually invoked)
+    #if: type IN (push, api)
```
dbrinegar: uncomment?

jsravn: Yes, will do. Wanted to test that it worked on the PR.

PR #316, opened 2019-02-26
Port the rest of bash smoketests

mitkus:
For https://github.com/lightbend/console-backend/issues/555

Bash `service` and `nginx` tests here are merged into a single `nginx` suite since they are mostly testing the same things.

One thing missing compared to bash tests is checking prefix of alert's backlink to alertmanager. Couldn't figure out how to do it without deploying listener pod and looking at its logs using kubectl. Maybe we can make a separate ticket for it?


mitkus:
@marcoderama yes, this doesn't test alertmanager backlink prefix. That's the only thing I couldn't figure out without using nc pod. Maybe I'll make an issue for it to be worked on outside of this PR?

mitkus:
Made an issue for missing test case: https://github.com/lightbend/console-backend/issues/623

PR #318, opened 2019-02-26
Remove kubernetes namespace label/srikanth

srikanthops:
Remove adding kubernetes_namespace label in  prometheus config.
add gotest to make sure none of the metrics have kubernetes_namespace label

lightbend/console-backend/issues/543


enterprise-suite/gotests/tests/prometheus/prometheus_test.go
```diff
@@ -126,6 +126,8 @@ var _ = Describe("all:prometheus", func() {
 		// All targets should be reachable
 		Expect(prom.HasData("up{kubernetes_name != \"es-test-service-with-only-endpoints\"} == 1")).To(Succeed())
 		Expect(prom.HasData("up{kubernetes_name != \"es-test-service-with-only-endpoints\"} == 0")).ToNot(Succeed())
+		// None of the metrics should have kubernetes_namespace label
+		// Expect(prom.HasData("{kubernetes_namespace!=\"\"}")).ToNot(Succeed())
```
marcoderama: Did you mean to leave that second line commented out?

srikanthops: Oops fixed

srikanthops:
I am not sure why the build is failing

srikanthops:
After increasing the timeout to 45 seconds, the tests on openShift passed @jsravn @mitkus please review and let me know your comments

jsravn:
@srikanthops the change you made makes no difference to the failure. You're seeing a flake that I hoped would be fixed by swapping to subqueries, but still seems to happen occasionally. I'm not sure the cause. All you need to do is rebuild.

enterprise-suite/gotests/util/prometheus/prometheus.go
```diff
@@ -32,8 +33,12 @@ type Connection struct {
...
 func (p *Connection) Query(query string) (*PromResponse, error) {
 	addr := fmt.Sprintf("%v/api/v1/query?query=%v", p.url, url.QueryEscape(query))
-
-	resp, err := http.Get(addr)
+	timeout := time.Duration(45 * time.Second)
+	client := http.Client{
+		Timeout: timeout,
+	}
+	// client.Get(url)
+	resp, err := client.Get(addr)
```
jsravn: I think it's good to have a timeout here (I believe the default is no timeout). But this won't affect the flake you saw.

enterprise-suite/gotests/util/prometheus/prometheus.go
```diff
@@ -32,8 +33,14 @@ type Connection struct {
 
 func (p *Connection) Query(query string) (*PromResponse, error) {
 	addr := fmt.Sprintf("%v/api/v1/query?query=%v", p.url, url.QueryEscape(query))
-
-	resp, err := http.Get(addr)
+	// Some of the tests with openshift clusters timedout in 20 seconds, so adding a timeout for
+	// 45 seconds. TODO: Remove this time out once sub-queries are available.
```
jsravn: This comment isn't correct. Can you remove it or change it? We're already using subqueries.

PR #319, opened 2019-02-28
Fix double namespace/srikanth

srikanthops:
lightbend/console-backend/issues/596

namespace can be provided as helm arg as --namespace=foobar or --namespace foobar 
Remove  from helm args and populate args.namespace
When I made this changes in setup_args tests failed, so moved the install 

Added two new test cases

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -528,26 +528,17 @@ def parse_set_string(s):
 def install(creds_file):
     creds_arg = '--values ' + creds_file
     version_arg = ('--version ' + args.version) if args.version != None else '--devel'
+    namespace_arg = "--namespace {}".format(args.namespace)
 
     helm_args = ''
     if len(args.helm) > 0:
-        # namespace can be passed to helm directly after -- and can take format -namespace <value> or --namespace=<value>
-        # remove namespace from helm args and assign it args.namespace
-        # handle --namespace <val>
-        if  "--namespace" in args.helm and args.helm.index("--namespace") < len(args.helm):
-            ns_index=args.helm.index("--namespace")
-            args.namespace=args.helm[ns_index+1]
-            del args.helm[ns_index:ns_index+2]
-
-        # handle --namespace=<val>
-        r=re.compile("--namespace=.*")
-        if filter(r.match, args.helm):
-            ns_val=filter(r.match, args.helm)[0]
-            args.namespace=ns_val.split("=")[1]
-            del(args.helm[args.helm.index(ns_val)])
-
-        # Helm args are separated from lbc.py args by double dash, filter it out
-        helm_args += ' '.join([arg for arg in args.helm if arg != '--']) + ' '
+        hparser = argparse.ArgumentParser()
+        hparser.add_argument('--namespace')
+        real_helm_args = args.helm[1:]
+        ns_args, unknown = hparser.parse_known_args(real_helm_args)
+        if ns_args.namespace != None:
+            namespace_arg=""
+        helm_args += ' '.join([arg for arg in real_helm_args]) + ' '
 
     # Add '--set' arguments to helm_args
     if args.set:
@@ -589,9 +580,10 @@ def install(creds_file):
                     fail('cannot access fetched chartfile at {}, ES_CHART={}'
                          .format(chartfile_glob, args.chart))
                 chartfile = chartfile[0]
-            execute('helm template --name {} --namespace {} {} {} {}'
-                    .format(args.helm_name, args.namespace, helm_args,
-                            creds_exec_arg, chartfile), print_to_stdout=True)
+
+            execute('helm template --name {} {} {} {} {}'
+                .format(args.helm_name, namespace_arg, helm_args,
+                creds_exec_arg, chartfile), print_to_stdout=True)
         finally:
             if tempdir:
                 shutil.rmtree(tempdir)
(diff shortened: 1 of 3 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -464,6 +464,21 @@ def install(creds_file):
 
     helm_args = ''
     if len(args.helm) > 0:
+        # namespace can be passed to helm directly after -- and can take format -namespace <value> or --namespace=<value>
+        # remove namespace from helm args and assign it args.namespace
+        # handle --namespace <val>
+        if  "--namespace" in args.helm and args.helm.index("--namespace") < len(args.helm):
```
jsravn: Is the `args.helm.index` necessary here? Not sure why it needs that check if `"--namespace" in args.helm` succeeds.

enterprise-suite/scripts/lbc.py
```diff
@@ -464,6 +464,21 @@ def install(creds_file):
...
+            ns_index=args.helm.index("--namespace")
+            args.namespace=args.helm[ns_index+1]
+            del args.helm[ns_index:ns_index+2]
+
+        # handle --namspace=<val>
```
jsravn: "namespace"

enterprise-suite/scripts/lbc.py
```diff
@@ -464,6 +464,21 @@ def install(creds_file):
...
+        r=re.compile("--namespace=.*")
+        if filter(r.match, args.helm):
+            ns_val=filter(r.match, args.helm)[0]
+            args.namespace=ns_val.split("=")[1]
+            del(args.helm[args.helm.index(ns_val)])
```
jsravn: Is this to handle the situation where a user does `--namespace=a -- --namespace=b`? If so, I think it would be better to log an error and exit.

jsravn:
@mitkus can you have a look as well. Seems reasonable to me with a couple small things.

enterprise-suite/scripts/lbc.py
srikanthops: This check is to make sure there at least one more variable provided as value for namespace

enterprise-suite/scripts/lbc.py
```diff
@@ -464,6 +464,21 @@ def install(creds_file):
...
+        # remove namespace from helm args and assign it args.namespace
+        # handle --namespace <val>
+        if  "--namespace" in args.helm and args.helm.index("--namespace") < len(args.helm):
+            ns_index=args.helm.index("--namespace")
+            args.namespace=args.helm[ns_index+1]
+            del args.helm[ns_index:ns_index+2]
+
+        # handle --namespace=<val>
+        r=re.compile("--namespace=.*")
+        if filter(r.match, args.helm):
```
marcoderama: Should this be in an `else`?  Is there a point looking for the `--namespace=...` version if we've already found a bare `--namespace` version?

enterprise-suite/scripts/lbc.py
```diff
@@ -464,6 +464,21 @@ def install(creds_file):
 
     helm_args = ''
     if len(args.helm) > 0:
+        # namespace can be passed to helm directly after -- and can take format -namespace <value> or --namespace=<value>
+        # remove namespace from helm args and assign it args.namespace
+        # handle --namespace <val>
+        if  "--namespace" in args.helm and args.helm.index("--namespace") < len(args.helm):
+            ns_index=args.helm.index("--namespace")
+            args.namespace=args.helm[ns_index+1]
```
marcoderama: I think we should check to see if there's already an `args.namespace` value that's different.  If so, I'd suggest we bail with a message.

Maybe we shouldn't default the `lbc.py` `namespace` value in the `sub parser.add_argument` call.  That way, `args.namespace` would only have a value at this point if it was explicitly set by the user.  We could then set it to our default (`lightbend`) if need be after we've looked for a `namespace` in the helm args.

enterprise-suite/scripts/lbc_test.py
```diff
@@ -206,6 +206,20 @@ def test_helm_args(self):
...
         lbc.main(['install', '--skip-checks', '--creds='+self.creds_file, '--delete-pvcs', '--', '--set', 'minikube=true', '--fakearg'])
 
+    def test_helm_args_namespace(self):
+        expect_cmd(r'helm repo add es-repo https://repo.lightbend.com/helm-charts')
+        expect_cmd(r'helm repo update')
+        expect_cmd(r'helm status enterprise-suite', returncode=-1)
+        expect_cmd(r'helm install es-repo/enterprise-suite --name enterprise-suite --namespace foobar --devel --values \S+ --set minikube=true --fakearg')
+        lbc.main(['install', '--skip-checks', '--creds='+self.creds_file, '--delete-pvcs', '--', '--set', 'minikube=true', '--fakearg', '--namespace=foobar'])
+
+    def test_helm_args_namespace_val(self):
```
marcoderama: I'd also like to see a test with no `--namespace` value specified in the `lbc.py` args.  And maybe one where the `lbc.py` one and the helm one differ.

srikanthops:
Based on the review, I made some changes.. 
use argparse instead of my own re epxression.. 

if there is namespace arg in helm_args in any of the following ways
--namespace=foo
--namespace foo 
--namespace=foo --namespace=bar --namespace space // the last one will be picked by argparse just like helm

I will ignore passing --namespace <args.namespace> to helm command

enterprise-suite/scripts/lbc.py
srikanthops: marco I changed the logic, this comment is not relevant with the latest changes, please check

enterprise-suite/scripts/lbc.py
srikanthops: the new logic, will check if there is namespace value is helm-args, if there is we will suppress the default one.. 
note: the last namespace value wins

enterprise-suite/scripts/lbc_test.py
mitkus: +1

jsravn:
Do you need help w/ the build @srikanthops ?

lightbend-cla-validator:
Hi @srikanthops,

lightbend-cla-validator:
Hi @srikanthops,

lightbend-cla-validator:
Hi @srikanthops,

lightbend-cla-validator:
Hi @srikanthops,

srikanthops:
The build for this is failing.. I think it has to do with my changes..

PR #320, opened 2019-02-28
Rename es-monitor-api to console-api

jsravn:
Preserve /service/es-monitor-api for backwards compatibility with
es-console-spa.

Also preserve the existing values.yml parameters, so we don't break
users that have these values encoded somewhere. That kind of change will
require a breaking release.


jsravn:
Reviewers: need to be careful not to break backwards compatibility here, while trying to avoid too many lingering "es-monitor-api" words. I think I've covered the edge cases.

enterprise-suite/gotests/testenv/testenv.go
```diff
@@ -141,7 +141,7 @@ func InitEnv() {
 	}
 
 	PrometheusAddr = fmt.Sprintf("%v/service/prometheus", ConsoleAddr)
-	MonitorAPIAddr = fmt.Sprintf("%v/service/es-monitor-api", ConsoleAddr)
```
marcoderama: Might there be value in continuing to test the old endpoint as well for now, especially given these changes?

enterprise-suite/tests/prometheus_common
```diff
@@ -20,7 +20,7 @@ pods_up() {
 
 ES_CONSOLE=$( busy_wait nodeport )
 PROM="$ES_CONSOLE/service/prometheus"
-ES_MONITOR_API="$ES_CONSOLE/service/es-monitor-api"
+ES_MONITOR_API="$ES_CONSOLE/service/console-api"
```
marcoderama: You didn't dare change the env var name I see.  :-)

enterprise-suite/tests/smoke_nginx
```diff
@@ -13,7 +13,7 @@ BASEPATH="${CONSOLE}/fee/fie/fou/fum"
 test_es_console_responding "$BASEPATH/"
 test_grafana_responding "$BASEPATH/service/grafana/"
 test_prom_responding "$BASEPATH/service/prometheus/"
-test_es_monitor_API_responding "$BASEPATH/service/es-monitor-api/"
+test_es_monitor_API_responding "$BASEPATH/service/console-api/"
```
marcoderama: This should call `test_console_API_responding` now I think.

enterprise-suite/values.yaml
```diff
@@ -3,8 +3,8 @@
 #
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-console
 esConsoleVersion: v1.0.1
-# lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-monitor-api
-esMonitorVersion: v1.0.9
+# lightbend-docker-commercial-registry.bintray.io/enterprise-suite/console-api
```
marcoderama: Should the version bump go in a separate PR or is that just a function of the name change?

enterprise-suite/values.yaml
```diff
@@ -27,8 +27,8 @@ busyboxVersion: "1.30"
 #
 # es-console image
 esConsoleImage: '{{ .Values.imageCredentials.registry }}/enterprise-suite/es-console'
-# es-monitor-api image
-esMonitorImage: '{{ .Values.imageCredentials.registry }}/enterprise-suite/es-monitor-api'
+# console-api image
+esMonitorImage: '{{ .Values.imageCredentials.registry }}/enterprise-suite/console-api'
```
marcoderama: Do we want to deprecate this and use `consoleImage` here?  (And ditto for `esMonitorVersion` -> `consoleVersion` above?)

enterprise-suite/gotests/testenv/testenv.go
jsravn: Sure - I will add a basic backwards compatibility check.

enterprise-suite/values.yaml
jsravn: It's related to the name change, since that is the version that changes to console-api.

enterprise-suite/values.yaml
jsravn: I think that should be a separate issue - to remove all the `es` prefixes.

PR #321, opened 2019-02-28
Disable go lint for now until it's fixed

jsravn:
This is broken in upstream with go modules.

See https://github.com/golang/lint/issues/436


jsravn:
Well this didn't work at all, so I'm going to just move gotests to `dep`.

PR #322, opened 2019-02-28
Swap gotests away from modules to glide

jsravn:
Swap away from modules due to problems encountered:
- golint randomly started failing
- module cache can get corrupted (should be fixed in 1.12 supposedly)
- client-go not working

Glide "just works" with client-go, so that's what I used. `dep` doesn't work
with client-go either (its SAT solver panics).

Dependencies are vendored so usage of glide isn't needed unless adding a new dependency.

More details at: https://github.com/kubernetes/client-go/blob/master/INSTALL.md#glide


srikanthops:
I am ok with the changes.. I am able to execute them on my laptop but the travis is still not happy

PR #325, opened 2019-03-01
lbc.py: require namespace arg for uninstall

mitkus:
For https://github.com/lightbend/console-backend/issues/622

This should fix the bug where `namespace` variable was being used without initialization. Also, this PR adds `--namespace` argument for `uninstall` subcommand.

I always found it odd that most subcommands (`install`, `verify`, `debug-dump`) had `--namespace` argument, but `uninstall` did not. That comes from `helm delete` which takes release name and doesn't need namespace. Now lbc.py will not have this inconsistency and will check if given namespace is the same as where helm release is installed.

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -553,6 +553,9 @@ def install(creds_file):
 def uninstall(status=None):
     if status == None:
         status, namespace = install_status(args.helm_name)
+        if namespace != args.namespace:
+            fail('Unable to delete console installation - released named {} found in namespace {}, expected in {}'
+                 .format(args.helm_name, namespace, args.namespace))
 
     if status == 'notfound':
         fail('Unable to delete console installation - no release named {} found'.format(args.helm_name))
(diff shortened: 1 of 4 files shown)


jsravn:
@marcoderama made the point that maybe we should eliminate namespace as an argument to lbc.py, and only accept it after `--`. That might simplify it a bit.

But anyways, this change looks good, thanks!

jsravn:
Build is failing when it tries to uninstall for some reason.

enterprise-suite/scripts/lbc.py
```diff
@@ -553,14 +553,17 @@ def install(creds_file):
 def uninstall(status=None):
     if status == None:
         status, namespace = install_status(args.helm_name)
+        if namespace != args.namespace:
+            fail('Unable to delete console installation - released named {} found in namespace {}, expected in {}'
```
marcoderama: released --> release

marcoderama:
I also found it odd that `helm delete` didn't care about namespace.  Is it possible to install two charts with the same name but different namespaces into the same tiller instance?  If not, then that would explain not caring.

I wonder if it's worth testing that before we add `--namespace` to the `uninstall` command?

marcoderama:
Wrt the "uninstall after failed deploy" issue, I have a slightly different proposal for a solution:
https://github.com/lightbend/console-charts/pull/327

Thoughts?

mitkus:
Closing this because the issue got solved in #327

PR #323, opened 2019-03-01
grafana frontend e2e test update

katsutoxin:
connect to https://github.com/lightbend/es-console-spa/pull/443

pass namespace in grafana url in frontend e2e test

PR #326, opened 2019-03-01
Bump es-console-spa to 1.0.3

jsravn:
- Customer logo
- Other improvements

Related also to https://github.com/lightbend/console-backend/issues/621

### DIFF
--- a/enterprise-suite/es-console/config.js
+++ b/enterprise-suite/es-console/config.js
@@ -1,7 +1,7 @@
 {{- /* Concatenates the key/values of the config into a list of JSON key=value, which are finally joined with ','. */ -}}
 {{- $settings := list }}
 {{- range $k, $v := .Values.consoleUIConfig }}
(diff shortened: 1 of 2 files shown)


pcraddick:
Looks like we made a mistake on the frontend side — we're reading a field called 'logo' rather than 'customer' :|

pcraddick:
> Looks like we made a mistake on the frontend side — we're reading a field called 'logo' rather than 'customer' :|

Actually, nm. The real problem is that we deviated from the design proposal in https://waffle.io/lightbend/console-frontend/cards/5c4b6318c050740486ca0f5d
when it came to the PR

The FE is expecting:

```
window.installedConfig = {
  'logo': 'ibm'
};
```

pcraddick:
'Paper' trail:

https://github.com/lightbend/console-frontend/issues/522#issuecomment-468743333

katsutoxin:
@jsravn   I merged the same change in #323    It requires your change to make frontend e2e test passed

jsravn:
Thanks @katsutoxin.

PR #328, opened 2019-03-01
Get PV info from PVCs/StorageClass info if needed (e.g. for OpenShift).

marcoderama:
The `lbc.py` logic to help avert accidental data loss currently depend on getting PV info from k8s.  Turns out that a typical user won't have access to this info on OpenShift.  We need a way to cope with this on OpenShift.

These changes get as much PV information as they can on OpenShift from other sources.

### DIFF
--- a/enterprise-suite/gotests/testenv/testenv.go
+++ b/enterprise-suite/gotests/testenv/testenv.go
@@ -3,6 +3,7 @@ package testenv
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
@@ -64,6 +65,7 @@ func InitEnv() {
 		Expect(err).To(Succeed(), "new k8sclient")
 	}
 
+	additionalLbcArgs := []string{}
 	additionalArgs := []string{"--set esConsoleURL=http://console.test.bogus:30080"}
 	if isMinikube {
 		additionalArgs = append(additionalArgs, "--set exposeServices=NodePort")
@@ -78,8 +80,10 @@ func InitEnv() {
 				break
 			}
 		}
-		if !foundStorageClass {
-			additionalArgs = append(additionalArgs, "--set usePersistentVolumes=false,managePersistentVolumes=false")
+		if foundStorageClass {
+			additionalLbcArgs = append(additionalLbcArgs, "--delete-pvcs")
+		} else {
+			additionalArgs = append(additionalArgs, "--set usePersistentVolumes=false")
 		}
 	}
 
@@ -106,7 +110,7 @@ func InitEnv() {
 	defer ticker.Stop()
 
 	// Install console
-	if err := lbc.Install(args.ConsoleNamespace, additionalArgs...); err != nil {
+	if err := lbc.Install(args.ConsoleNamespace, strings.Join(additionalLbcArgs, " "), additionalArgs...); err != nil {
 		Expect(err).To(Succeed(), "lbc.Install")
 	}
 
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -338,22 +338,57 @@ def pvs_retained_and_not(namespace):
                  '{{ printf "%s %s %s %s\\n" .metadata.name .spec.persistentVolumeReclaimPolicy .spec.claimRef.name .status.phase }}{{ end }}{{ end }}'
                 )
     returncode, stdout, _ = run("kubectl get pv -o go-template='{}'"
-                                .format(go_template))
+                                .format(go_template), show_stderr=False)
+    # stdout contains...
+    # pvc-8149517b-3bae-11e9-83b7-08002755d2dd Delete alertmanager-storage Bound
 
+    pvs = []
     retained = []
     notRetained = []
 
     if returncode != 0:
-        printerr("Unable to retrieve Persistent Volume info.  Proceed with caution.")
-    else:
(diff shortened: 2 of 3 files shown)


marcoderama:
A comment on this issue originally posted in 
https://github.com/lightbend/console-backend/issues/606#issuecomment-468098191

> I've been working on testing this stuff, in particular on OpenShift, and I'm kinda bummed... A lot of the modifications rely on knowing information about the state of the PVs, but it turns out the typical user doesn't have the privileges to do so in OpenShift. With the current changes, this means a "dangerous" uninstall will go ahead unimpeded on OpenShift...

> Thinking about how to improve that...

marcoderama:
A comment on this issue originally posted in
https://github.com/lightbend/console-backend/issues/606#issuecomment-468099511

> Had an idea. Not perfect but better than nothing...

> On OpenShift I _can_ get at the PVC info, and that includes the associated storage class. I can also get at the reclaim policy for a storage class, so I can at least know if it's `Delete` or `Retain`.

> I won't be able to divine the Phase of the PV though. (eg. `Released`, `Available`) That will prevent me from generating certain warnings about the possibility of orphaning data.

> So while this isn't perfect, it should allow us to prevent inadvertent deletion of PVCs associated with a `Delete` storage class. And that's the key issue.

marcoderama:
And that's essentially how these changes work.

jsravn:
@marcoderama why not always use PVCs for the reclaim info?

marcoderama:
@jsravn: The retention policy is tied to the PV, not the PVC, so I considered it preferable to get the info from the PV if possible.  Using the PVC info, we can get the storage class, and are allowed to look at the  _default_ retention policy for the storage class.  It would be possible for the PVC to be bound to a PV of that storage class but with a different retention policy (I believe).  So getting it via the PVC is a good guess but not definitive, like it is when getting it direct from the PV.

jsravn:
Why not always use StorageClass in that case? I know it's not absolute (a user could change it manually on the PV I guess?) but it is a lot simpler!

enterprise-suite/gotests/testenv/testenv.go
```diff
@@ -85,7 +87,9 @@ func InitEnv() {
 				break
 			}
 		}
-		if !foundStorageClass {
+		if foundStorageClass {
+			additionalLbcArgs = append(additionalLbcArgs, "--delete-pvcs")
+		} else {
```
jsravn: Can you remove 'managerPersistentVolumes=false' in the following line?

enterprise-suite/scripts/lbc.py
```diff
@@ -338,22 +338,54 @@ def pvs_retained_and_not(namespace):
                  '{{ printf "%s %s %s %s\\n" .metadata.name .spec.persistentVolumeReclaimPolicy .spec.claimRef.name .status.phase }}{{ end }}{{ end }}'
                 )
     returncode, stdout, _ = run("kubectl get pv -o go-template='{}'"
-                                .format(go_template))
+                                .format(go_template), show_stderr=False)
+    # pvc-8149517b-3bae-11e9-83b7-08002755d2dd Delete alertmanager-storage Bound
```
jsravn: What's this comment?

jsravn:
@mitkus can you have a look as well?

enterprise-suite/gotests/testenv/testenv.go
```diff
@@ -71,6 +72,7 @@ func InitEnv() {
 		cleanup(true)
 	}
 
+	additionalLbcArgs := []string{}
```
mitkus: This should work without `additionalLbcArgs`, just put what you want in `additionalArgs`.

enterprise-suite/scripts/lbc.py
marcoderama: That and the comment at line 356 show what is expected from running the command.  I could remove or add an `# output like` comment...

enterprise-suite/gotests/testenv/testenv.go
marcoderama: Hmmm.  Wouldn't that result in the `--delete-pvcs` bit getting injected after the `--` bit when invoking lbc.py in lbc.go:Install?  Around here...
```
    defaultArgs := []string{"install", "--local-chart", localChartPath,
        "--namespace", namespace,
        "--set prometheusDomain=console-backend-e2e.io",
        "--wait",
        additionalLbcArgs,
        "--", "--timeout 110"}
    fullArgs := append(defaultArgs, additionalArgs...)
    cmd := util.Cmd(lbcPath, fullArgs...)
```

marcoderama:
> Why not always use StorageClass in that case? I know it's not absolute (a user could change it manually on the PV I guess?) but it is a lot simpler!

Yeah, it's a bit simpler but I wanted to get a definitive value if possible.  (Given we're trying to prevent data loss here.)  And a user could bind to an explicit PV no?  In that case there'd be no storage class info to be found on the PVC (I think).

marcoderama:
I haven't been able to figure out why the Go tests are failing the way they are...

marcoderama:
c.f. @mitkus' explorations into why the args to the helm call are getting messed up:  https://github.com/lightbend/console-charts/pull/334

jsravn:
I'll try to have a look as well. It is indeed strange.

marcoderama:
These changes were incorporated into the changes for https://github.com/lightbend/console-charts/pull/342.  That PR also fixed the annoying gotest failures.  Yay!

Closing this one.

PR #330, opened 2019-03-06
re-enable 3 frontend e2e tests

katsutoxin:
connect to https://github.com/lightbend/console-frontend/issues/554


katsutoxin:
> lgtm but what do I know. There does seem to be a lot of sleeping going on. Is all that necessary? How long do these tests take to run in the end?

I disabled several sub-tests to make run time faster.
The new test takes 13 min 10 sec
The original test  takes 9 min 28 sec.
Increase around 3 min 42 sec.   Is it good enough?  

If you think the new test is too long, I need to  separate tests to two sets of tests……
The running result is here https://travis-ci.com/lightbend/console-charts/jobs/182920892

jsravn:
I think it's pushing it. A 10 minute build is a good guideline as a limit to build time. So perhaps we need to look into splitting it up somehow.

katsutoxin:
@jsravn  I split frontend e2e test to two set of tests.  Now each one takes around 10 minutes.  It should be good to go.

However, The real e2e test does not take such a long time.   Most time are spent on spinning up the minikube env and helm install es-console (takes more than 6 minutes).   It will be great if we have a linux image which contains pre-installed minikube/helm env to speed up the total build time.  ( or test it in central park env, we can do helm delete and helm install to do the test to avoid spending time on minikube setup)

PR #331, opened 2019-03-06
Add scripts/ to helmignore so that it doesn't end up in charts tarball

mitkus:
This will remove the following files from our next helm chart releases:

```
enterprise-suite-latest/scripts/install-es.sh
enterprise-suite-latest/scripts/lbc.py
enterprise-suite-latest/scripts/lbc_test.py
enterprise-suite-latest/scripts/pipelines.yaml
enterprise-suite-latest/scripts/run-e2e-tests.sh
enterprise-suite-latest/scripts/validate-promql.sh
```

### DIFF
--- a/enterprise-suite/.helmignore
+++ b/enterprise-suite/.helmignore
@@ -28,5 +28,6 @@ README.md
 \#
 tests/
 gotests/
+scripts/
 .vscode/
 .venv/


marcoderama:
After the fact, but do we not think there's value in shipping with `lbc.py`?

jsravn:
@marcoderama can you explain why? This is the chart tarball helm uses - user will have no access to extra files in it normally.

marcoderama:
I have no good reason to think that way I guess.  Certainly no need for `lbc.py` on the cluster.

PR #334, opened 2019-03-08
[Don't merge] Installer argument parsing investigations

mitkus:
This is just a temporary PR for me to investigate installer on Travis CI.

### DIFF
--- a/enterprise-suite/gotests/testenv/testenv.go
+++ b/enterprise-suite/gotests/testenv/testenv.go
@@ -3,6 +3,7 @@ package testenv
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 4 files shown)


mitkus:
@jsravn To give you an overview where I'm at with this issue:

1) Added additional logging to see what exact python version is used and how arguments are parsed into Namespace object by argparse. For example, `uninstall` prints these two additional lines:
```
lbc.py invoked with arguments Namespace(delete_pvcs=False, dry_run=False, helm=[], helm_name='enterprise-suite', skip_checks=False, subcommand='uninstall')
python version: sys.version_info(major=2, minor=7, micro=12, releaselevel='final', serial=0)
```

2) This revealed that parsing args after `--` using argparse.REMAINDER is broken in Travis and it captured `--set` and `--delete-pvcs` as helm args. I rewrote that part into a simpler manual code that doesn't use argparse.REMAINDER

3) Now it seems argparse cannot parse some `--set` args at all:
```
lbc.py: error: unrecognized arguments: --set prometheusDomain=console-backend-e2e.io
```
I suspect this is what led to all the issues - on some systems argparse can parse it ok, and on some it can't.

jsravn:
This should be fixed now in https://github.com/lightbend/console-charts/pull/342.

PR #335, opened 2019-03-08
Queue travis builds

jsravn:
So only one happens at a time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,13 @@ dist: xenial
 services:
 - docker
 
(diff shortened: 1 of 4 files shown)


marcoderama:
Hmm.  Realize now I should just have said "lgtm" rather than approve at this stage.  Any way to revoke and approval other than by requesting changes?

jsravn:
image

If you click show all reviewers, it gives the option to 'dismiss review'.

jsravn:
Okay, this should be working now!

PR #337, opened 2019-03-12
kafka monitors suitable for pipelines clients

dbrinegar:
fixes https://github.com/lightbend/console-backend/issues/631

Provides pre-aggregated throughput and lag metrics (since Console UI cannot handle aggregation section in a monitor). The metrics all come from vanilla kafka client metrics which only have a `topic` dimension. We include various context tags for the workload kind of scope, and for lag we include an instance tag. Note that throughput is aggregated across partitions and modeled as a grand total, so we are not flagging or catching throughput changes at the partition or instance level. For lag however we have max over window and max over instance is sensible, so we retain the instance distinction.

The Pipelines UI uses `app_kubernetes_io_name` and `topic` to join into other data sets, so inlet and outlet metadata may be added.

Otherwise these are just conservative 15 minute sma and growth monitors, nothing unusual. We may tune them later.

One can see them in action in the zandvoort cluster:

http://console-server-lightbend.zandvoort.lightbend.com/cluster

http://console-server-lightbend.zandvoort.lightbend.com/namespaces/app-call-record-pipeline/workloads/call-record-pipeline-cdr-aggregator-exec

### DIFF
--- a/enterprise-suite/console-api/static-rules.yml
+++ b/enterprise-suite/console-api/static-rules.yml
@@ -64,3 +64,12 @@
 
 - record: akka_http_http_server_responses_5xx_rate
   expr: irate(akka_http_http_server_responses_5xx[5m]) and akka_http_http_server_responses_5xx offset 1m
+
+- record: kafka_consumer_topic_consumed_rate
+  expr: sum by (app_kubernetes_io_component, app_kubernetes_io_managed_by, app_kubernetes_io_name, app_kubernetes_io_part_of, es_monitor_type, es_workload, namespace, topic) (kafka_consumer_consumer_fetch_manager_metrics_records_consumed_rate)
+
+- record: kafka_producer_topic_send_rate
+  expr: sum by (app_kubernetes_io_component, app_kubernetes_io_managed_by, app_kubernetes_io_name, app_kubernetes_io_part_of, es_monitor_type, es_workload, namespace, topic) (kafka_producer_producer_metrics_record_send_rate)
+
+- record: kafka_consumer_topic_lag_max
+  expr: max by (app_kubernetes_io_component, app_kubernetes_io_managed_by, app_kubernetes_io_name, app_kubernetes_io_part_of, es_monitor_type, es_workload, namespace, topic, instance) (kafka_consumer_consumer_fetch_manager_metrics_records_lag_max >= 0)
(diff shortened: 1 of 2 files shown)


enterprise-suite/console-api/static-rules.yml
```diff
@@ -64,3 +64,12 @@
 
 - record: akka_http_http_server_responses_5xx_rate
   expr: irate(akka_http_http_server_responses_5xx[5m]) and akka_http_http_server_responses_5xx offset 1m
+
+- record: kafka_consumer_topic_consumed_rate
+  expr: sum by (app_kubernetes_io_component, app_kubernetes_io_managed_by, app_kubernetes_io_name, app_kubernetes_io_part_of, es_monitor_type, es_workload, namespace, topic) (kafka_consumer_consumer_fetch_manager_metrics_records_consumed_rate)
```
marcoderama: Do you mean to have `kafka_consumer_consumer...` (and `producer_producer` below) in there?

dbrinegar: yep, those are the kafka metrics :-|

jsravn:
@dbrinegar can you also document these in the tech hub docs?

It's not clear to me what these check for - so I think a bit of documentation will go a long way.

I'm also curious where the parameters came from - is that based on testing against a real kafka app?

jsravn:
Do these metrics only work w/ a specific kafka exporter configuration? If so, we need to document that as well. Perhaps just say that they are intended for strimzi operator? (I'm not sure what is setting these metrics up)

dbrinegar:
Ya can add a blurb on the doc site.

These metrics are from pipelines applications, which now have a built in kafka client. They're core kafka client metrics, as configured for pipelines apps, but we've only had them in this form factor for a few days so we're still in learning mode. They appear to be better than the various alternatives we've tried, especially because they export from the app instances so discovery and integration are all easy. We've also found some bad bugs to work around, but so far the tradeoffs are favorable.

The monitor parameters are just default 15m conservative settings, they'll likely need some dialing in as we test pipelines scenarios down the road. These are needed for the Pipelines release, but more for the graphs than the monitors. I think on the monitoring side, the question will again be: can we have one set of monitors for dev mode / debugging, and another for production alerts? But we aren't going there for 1.0, this is essentially a placeholder until we learn more.

jsravn:
ping @dbrinegar - did you add docs?

dbrinegar:
We're still in the mode of trying to figure out the monitor details. Docs will come after.

jsravn:
Since we've shipped these now in the public release, can we add a placeholder doc for them - something to the effect of they are in beta?

If someone sees these in their console all of a sudden, we need some explanation for them. If it's not ready for consumption then we probably should not add them to master yet.

PR #338, opened 2019-03-12
Make frontend e2e test in cron job also splits to 2 set of tests

katsutoxin:
- Make frontend e2e test in cron job also splits to 2 set of tests
- Fix unbound variable if running all front e2e tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ jobs:
       - artifacts upload $(find . -name '*.mp4')
       - artifacts upload $(find . -name '*.png')
 
-  - name: Frontend E2E Tests - Minikube Latest
+  - name: Frontend E2E Tests 1 - Minikube Latest
     # Only build latest in the daily cron - don't block PRs or master releases.
     if: type in (cron)
     cache:
@@ -77,7 +77,23 @@ jobs:
     node_js: 9.4.0
     script:
     - travis_retry scripts/setup-minikube-for-linux.sh
-    - make -C enterprise-suite-latest frontend-tests
+    - make -C enterprise-suite-latest frontend-tests1
+    after_script:
+      # Cypress videos.
+      - artifacts upload $(find . -name '*.mp4')
+      - artifacts upload $(find . -name '*.png')
+
+  - name: Frontend E2E Tests 2 - Minikube Latest
+    # Only build latest in the daily cron - don't block PRs or master releases.
+    if: type in (cron)
+    cache:
+      directories:
+        - ~/.npm
+        - ~/.cache
+    node_js: 9.4.0
+    script:
+    - travis_retry scripts/setup-minikube-for-linux.sh
+    - make -C enterprise-suite-latest frontend-tests2
     after_script:
       # Cypress videos.
       - artifacts upload $(find . -name '*.mp4')
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,9 @@ jobs:
         - ~/.npm
         - ~/.cache
     node_js: 9.4.0
+    install:
+    - travis_retry scripts/setup-tools-for-ubuntu.sh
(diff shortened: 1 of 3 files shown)


enterprise-suite/scripts/run-e2e-tests.sh
```diff
@@ -45,7 +45,7 @@ if [[ "$?" != "0" ]]; then
     exit 1
 fi
 
-if [[ -z "$subset" ]]; then
+if [[ "$subset" == "all" ]]; then
```
marcoderama: Why not make the default to run all of them and require a subset parameter for just 1 or 2?

enterprise-suite/Makefile
```diff
@@ -43,7 +43,7 @@ script-tests: ## Run installer tests
 
 .PHONY: frontend-tests
 frontend-tests: ## Run the frontend tests against minikube
```
marcoderama: The comment here and in the next two targets should be modified.  (e.g. first one runs all tests, second one runs first subset, third one runs second subset.)

enterprise-suite/Makefile
```diff
@@ -43,7 +43,7 @@ script-tests: ## Run installer tests
...
-	$(CHART_DIR)/scripts/run-e2e-tests.sh
+	$(CHART_DIR)/scripts/run-e2e-tests.sh all
```
marcoderama: Don't need this change if you make running all tests the default.

.travis.yml
```diff
@@ -77,7 +77,23 @@ jobs:
...
-    - make -C enterprise-suite-latest frontend-tests
+    - make -C enterprise-suite-latest frontend-tests1
+    after_script:
+      # Cypress videos.
+      - artifacts upload $(find . -name '*.mp4')
+      - artifacts upload $(find . -name '*.png')
+
+  - name: Frontend E2E Tests 2 - Minikube Latest
+    # Only build latest in the daily cron - don't block PRs or master releases.
+    if: type in (cron)
```
marcoderama: I wonder if the common stuff can be included only once somehow.  My Travis-foo isn't good enough to know...

enterprise-suite/scripts/run-e2e-tests.sh
katsutoxin: good point !

enterprise-suite/Makefile
katsutoxin: we still need to pass something,  because we have "set -u" in `run-e2e-tests.sh`
If we don't put any parameter, it will complain ```line 5: $1: unbound variable```

enterprise-suite/scripts/run-e2e-tests.sh
katsutoxin: I made some change, default set "all" as default parameter, but just in case people develop new code and create  3 or 4 as parameter, I still do the parameter check

enterprise-suite/Makefile
marcoderama: I'd still like to see this change but no biggy

enterprise-suite/Makefile
```diff
@@ -42,13 +42,15 @@ script-tests: ## Run installer tests
...
 	$(CHART_DIR)/scripts/run-e2e-tests.sh
 
-frontend-tests1: ## Run the frontend tests against minikube
+.PHONY: frontend-tests1
+frontend-tests1: ## Run subset 1 of frontend tests against minikube
 	$(CHART_DIR)/scripts/run-e2e-tests.sh 1
 
-frontend-tests2: ## Run the frontend tests against minikube
+.PHONY: frontend-tests2
+frontend-tests2: ## Run subset 1 of frontend tests against minikube
```
marcoderama: Heh.  "subset 2" here...  :-)

PR #347, opened 2019-03-16
Spark fix

dbrinegar:
Remove constraint for three digit instances, as spark can get into a
mode where the timestamp stays the same but instances roll on each new
pod, causing the instance number to be arbitrarily large. This seems an
edge case, nevertheless it also seems safe to assume this is a spark
workload if we've seen "timestamp" + "-exec-" + "any number"

Adds "timestamp-driver" workload parsing, which can be seen with spark
jobs. This pattern doesn't happen with Pipelines apps, but does happen
with some of our example spark jobs for FDP.


lightbend-cla-validator:
Hi @dbrinegar,

dbrinegar:
current theory:  the timestamp updates when the driver restarts, the instance updates when the executor restarts

PR #350, opened 2019-03-18
Fix node labels/srikanth

srikanthops:
use node_name instead of node_ip
remove node relabel rules for kube-state-metrics

lightbend/console-backend/issues/602


lightbend-cla-validator:
Hi @srikanthops,

enterprise-suite/console-api/prometheus.yml
```diff
@@ -48,6 +48,8 @@ scrape_configs:
         replacement: /api/v1/nodes/${1}/proxy/metrics/cadvisor
       - source_labels: [__meta_kubernetes_node_name]
         target_label: node_name
+      - source_labels: [__meta_kubernetes_node_name]
+        target_label: node
```
dbrinegar: I notice that you're duplicating a tag here, and leaving `node_name` across the various data sets. Is there a reason to do it this way instead of, say, rename the kube_state_metrics `node` tag to `node_name`?  It looks like that would make everything consistent with no dupes or holes

dbrinegar: Since this is a primary key for joining across data sets, you could add a consistency test. Should be easy, and would catch breakage of this in future.

enterprise-suite/console-api/prometheus.yml
```diff
@@ -140,6 +140,8 @@ scrape_configs:
         action: keep
       - action: labelmap
         regex: __meta_kubernetes_service_label_(.+)
+      - source_labels: [__meta_kubernetes_namespace]
+        target_label: namespace
```
dbrinegar: this is the wrong way to debug the UI test. this is a bad tag, let's figure out what's happening with the UI tests without reverting this fix

enterprise-suite/console-api/prometheus.yml
```diff
@@ -200,6 +192,10 @@ scrape_configs:
       - source_labels: [pod, es_workload]
         regex: (.+);
         target_label: es_workload
+      - source_labels: [node]
+        target_label: node_name
+      - regex: 'node'
```
dbrinegar: this needs to be deprecated slowly, so UI can switch to node_name. After UI switches, we can remove it here

PR #353, opened 2019-03-20
handle both jmx_exporter and cinnamon jvm metrics

dbrinegar:
fixes lightbend/console-backend#646

If jmx_exporter version of jvm metrics is available for a workload,
we'll use those. If not then we'll try the Cinnamon version.

Some of the Cinnamon metrics have a GC or Pool value embedded in the
name, so here we parse those out into a tag to get the same ergonomics
as jmx_exporter. This keeps the jmx_exporter or cinnamon_metric
expression lined up, compresses multiple promql expressions into one,
and handles any gc or pool type found without having to know them ahead
of time.

Note that for GC Rate and GC Time we are using Prometheus 2.7 subquery
features so promtool and in-cluster prometheus must be >= 2.7 for these
to work.

Here's a Pipelines Akka Streamlet with a mix of jmx_exporter and Cinnamon jvm metrics:

image

### DIFF
--- a/enterprise-suite/es-grafana/jvm-exporter.json
+++ b/enterprise-suite/es-grafana/jvm-exporter.json
@@ -3,19 +3,19 @@
     "graphName": "Heap",
     "promQL": [
       {
-        "expr": "sum by (kubernetes_pod_name)(jvm_heap_committed{ContextTags})",
+        "expr": "jvm_memory_bytes_committed{ContextTags,area=\"heap\"} or on (kubernetes_pod_name) jvm_heap_committed{ContextTags}",
         "legendFormat": "committed {{kubernetes_pod_name}}"
       },
       {
-        "expr": "sum by (kubernetes_pod_name)(jvm_heap_init{ContextTags})",
+        "expr": "jvm_memory_bytes_init{ContextTags,area=\"heap\"} or on (kubernetes_pod_name) jvm_heap_init{ContextTags}",
         "legendFormat": "init {{kubernetes_pod_name}}"
       },
       {
-        "expr": "sum by (kubernetes_pod_name)(jvm_heap_max{ContextTags})",
+        "expr": "jvm_memory_bytes_max{ContextTags,area=\"heap\"} or on (kubernetes_pod_name) jvm_heap_max{ContextTags}",
         "legendFormat": "max {{kubernetes_pod_name}}"
       },
       {
-        "expr": "sum by (kubernetes_pod_name)(jvm_heap_used{ContextTags})",
+        "expr": "jvm_memory_bytes_used{ContextTags,area=\"heap\"} or on (kubernetes_pod_name) jvm_heap_used{ContextTags}",
         "legendFormat": "used {{kubernetes_pod_name}}"
       }
     ],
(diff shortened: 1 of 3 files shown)


enterprise-suite/es-grafana/jvm-exporter.json
```diff
@@ -3,19 +3,19 @@
...
     "promQL": [
       {
-        "expr": "sum by (kubernetes_pod_name)(jvm_heap_committed{ContextTags})",
+        "expr": "jvm_memory_bytes_committed{ContextTags,area=\"heap\"} or on (kubernetes_pod_name) jvm_heap_committed{ContextTags}",
         "legendFormat": "committed {{kubernetes_pod_name}}"
       },
       {
-        "expr": "sum by (kubernetes_pod_name)(jvm_heap_init{ContextTags})",
+        "expr": "jvm_memory_bytes_init{ContextTags,area=\"heap\"} or on (kubernetes_pod_name) jvm_heap_init{ContextTags}",
         "legendFormat": "init {{kubernetes_pod_name}}"
```
jsravn: Is there a reason to show "init"?

enterprise-suite/es-grafana/jvm-exporter.json
```diff
@@ -75,37 +75,35 @@
...
     "promQL": [
       {
-        "expr": "sum by (kubernetes_pod_name)(rate(jvm_PS_MarkSweep_count{ContextTags}[5m]))",
-        "legendFormat": "PS_MarkSweep Rate {{kubernetes_pod_name}}"
-      },
-      {
-        "expr": "sum by (kubernetes_pod_name)(rate(jvm_PS_Scavenge_count{ContextTags}[5m]))",
-        "legendFormat": "PS_Scavenge Rate {{kubernetes_pod_name}}"
+        "expr": "rate( (jvm_gc_collection_seconds_count{ContextTags} or on (kubernetes_pod_name, gc) label_replace({ContextTags,__name__=~\"jvm_.*_count\",gc=\"\"}, \"gc\", \"$1\", \"__name__\", \"jvm_(.*)_count\")) [5m:])",
+        "legendFormat": "{{gc}} rate {{kubernetes_pod_name}}"
```
jsravn: Is `rate` really the best thing here? I think it will be quite misleading - you'll see "1 gc per second" which is not accurate. Instead I'd want to see the number of GCs in the last scrape period. I think increase would be better to be honest, or subtract against offset.

At minimum this should be titled "GC Rate (avg per second)".

enterprise-suite/es-grafana/jvm-exporter.json
```diff
@@ -75,37 +75,35 @@
...
     "promQL": [
       {
-        "expr": "max by (kubernetes_pod_name)(jvm_PS_MarkSweep_time{ContextTags})",
-        "legendFormat": "PS_MarkSweep max {{kubernetes_pod_name}}"
-      },
-      {
-        "expr": "max by (kubernetes_pod_name)(jvm_PS_Scavenge_time{ContextTags})",
-        "legendFormat": "PS_Scavenge max {{kubernetes_pod_name}}"
+        "expr": "rate( (jvm_gc_collection_seconds_sum{ContextTags} or on (kubernetes_pod_name, gc) label_replace({ContextTags,__name__=~\"jvm_.*_time\",gc=\"\"}, \"gc\", \"$1\", \"__name__\", \"jvm_(.*)_time\") /1000) [5m:])",
+        "legendFormat": "{{gc}} time {{kubernetes_pod_name}}"
```
jsravn: Same comment as before... using `rate` effectively smooths out the GC time, and will make it look much better than it really is. The important aspect of GC time is seeing if you have very long pause times. This will be lost. At minimum, the title of the graph should reflect this is an "average per second".

enterprise-suite/es-grafana/jvm-exporter.json
dbrinegar: This change was to merge jmx_exporter and cinnamon, and no effort was made to change the graphs otherwise.

I suppose someone tuning a JVM may be interested in init as a baseline, I dunno.

enterprise-suite/es-grafana/jvm-exporter.json
dbrinegar: I'm not sure how top implement subtract against offset, and `increase` is sugar for `rate` so that has the same graph but different y-axis which I don't think you meant.  So I've changed the title. Happy to do some different graph function if it works, but the promql set for counters is limited and I don't know a better way here than rate.

enterprise-suite/es-grafana/jvm-exporter.json
dbrinegar: graph title changed

enterprise-suite/es-grafana/jvm-exporter.json
jsravn: Offset should work, e.g. `jvm_gc_collection_seconds_count - jvm_gc_collection_seconds_count offset 30s` <- GC time spent in the last 30s. I guess it will blow up on a counter reset, but this counter should never reset (if pod is redeployed it will be a new time series I think).

PR #354, opened 2019-03-21
Windows support for lbc.py

mitkus:
For https://github.com/lightbend/console-backend/issues/609

Prerequisites:
- OpenShift client ("oc") in PATH
- kubectl in PATH
- helm in PATH
- Python 2.7 installed

Manually tested scenarios in Central Park:
- Install
- Install with --wait
- Upgrade with --delete-pvcs
- Uninstall with --delete-pvcs
- Verify
- Debug-dump

lbc_windows

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ go:
 sudo: required
 dist: xenial
 
+os:
+- linux
+- windows
+
 services:
 - docker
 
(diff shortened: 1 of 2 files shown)


jsravn:
@mitkus would it be worth adding a very basic Windows test of it? https://docs.travis-ci.com/user/reference/windows/

Not sure how much effort it would be - if it will be a lot of work, then we can create a follow up task for it. Just want to be sure we don't break this in the future given we never use Windows.

jsravn:
If we can't do a windows environment test - can we get sufficient coverage in a unit test?

.travis.yml
```diff
@@ -2,21 +2,32 @@ language: go
...
+  - stage: tests
     name: Unit tests
+    os: linux
+    dist: xenial
     install:
     - travis_retry scripts/setup-tools-for-ubuntu.sh
     script:
     - make test
-
+
```
marcoderama: What is the extraneous space I see before me?!
(and again on line 159)

mitkus: Sorry!

PR #357, opened 2019-03-22
Deploy to console-demo environment on tag

jsravn:
part of https://github.com/lightbend/console-backend/issues/612

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -183,7 +183,7 @@ jobs:
     # travisqueue.go ensures only a single build runs at a time on Openshift
     - go run travisqueue.go
     - travis_retry scripts/setup-openshift.sh ${OC_DEMO_TOKEN}
-    - enterprise-suite/scripts/lbc.py install --namespace=console-demo --local-chart=enterprise-suite --set usePersistentVolumes=true,defaultStorageClass=gp2
+    - TILLER_NAMESPACE=console-demo enterprise-suite/scripts/lbc.py install --namespace=console-demo --local-chart=enterprise-suite --set usePersistentVolumes=true,defaultStorageClass=gp2
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -176,7 +176,7 @@ jobs:
     - travis_retry scripts/trigger-build-helm-charts.sh
 
   - name: Deploy to Demo Environment
-    if: type in (push, api) AND tag is present
+    #if: type in (push, api) AND tag is present
     install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -135,7 +135,7 @@ jobs:
   - stage: deploy
     name: Deploy to GCS
     # Only deploy for pushes to master (and if manually invoked)
```
marcoderama: Might want to tweak this comment now (wrt need for tag)...  (Although that's pretty much self-evident from the code...)

.travis.yml
```diff
@@ -147,9 +147,21 @@ jobs:
...
 
+  - name: Deploy to Demo Environment
+    #if: type in (push, api) AND tag is present
+    install:
+    - travis_retry scripts/setup-tools-for-ubuntu.sh
+    script:
+    # travisqueue.go ensures only a single build runs at a time on Openshift
+    - go run travisqueue.go
+    - travis_retry scripts/setup-openshift.sh ${OC_DEMO_TOKEN}
+    - TILLER_NAMESPACE=console-demo enterprise-suite/scripts/lbc.py install --namespace=console-demo --local-chart=enterprise-suite --set usePersistentVolumes=true,defaultStorageClass=gp2
```
marcoderama: `--set usePersistentVolumes=true,defaultStorageClass=gp2` is redundant now.

(Unless this is an update install, which I don't think is the case...)

.travis.yml
```diff
@@ -147,9 +147,21 @@ jobs:
...
+
 branches:
   only:
   - master
+  # build tags starting with `v`, e.g. v0.0.16
+  - /^v.*$/
```
marcoderama: Doubt it really matters, but that would accept "vagabond" etc. too...

jsravn: Yeah I don't think we need to worry about that.

However I just realize we don't use semver for our console charts tag (I think because we originally thought it would have multiple charts). As a result this won't work. So need to think about that.

marcoderama: How is our tag not semver compliant?  I see tags like:
```
enterprise-suite-1.0.0
enterprise-suite-1.0.0-rc.2
enterprise-suite-1.0.0-rc.3
enterprise-suite-1.0.0-rc.4
enterprise-suite-1.0.0-rc.5
enterprise-suite-1.0.0-rc.6
enterprise-suite-1.0.0-rc.7
enterprise-suite-1.0.1
```
and those all look good to me.  Is there a new tag now?

PR #359, opened 2019-03-25
Remove --devel flag

jsravn:
So users don't pick up the `rc` version if they are using latest, now
that we have 1.0 out.

Also improve lbc_test.py to provide traceback to the test line that
failed.

### DIFF
--- a/enterprise-suite/scripts/lbc_test.py
+++ b/enterprise-suite/scripts/lbc_test.py
@@ -6,6 +6,8 @@ import collections
 import tempfile
 import shutil
 import unittest
+import traceback
+import sys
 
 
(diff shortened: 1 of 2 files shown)


enterprise-suite/scripts/lbc_test.py
```diff
@@ -27,15 +29,16 @@ def test_fail(msg):
 
 # Adds a command to the list
 def expect_cmd(re_pattern, returncode=0, stdout=''):
-    expected_cmds.append((re_pattern, returncode, stdout))
+    # find line number in test for later reporting on a failure
+    stack = traceback.format_stack()[-2]
```
mitkus: Neat

PR #366, opened 2019-03-28
Small change to lbc.py help text

marcoderama:
I don't think `emptyDir` is a valid option.  Just changed the help text to use an example with an option a user might actually use.


lightbend-cla-validator:
Hi @marcoderama,

jsravn:
Nice catch.

PR #370, opened 2019-04-01
Move default monitors to console-api (Part 2)

mitkus:
For https://github.com/lightbend/console-backend/issues/644

This PR does a couple of things:
- Introduces `check-default-monitors-sync` and `sync-default-monitors` makefile targets for `enterprise-suite`, that respectively verify or sync `console-api/default-monitors.json` and `console-api/static-rules.yml` files
- New helm chart value `defaultMonitorsConfigMap` for overriding default monitors
- Updates `console-api` version in the chart to v1.0.14 that bundles its own default monitors

After this PR is merged, all changes to `default-monitors.json` and `static-rules.yml` should happen in https://github.com/lightbend/console-api/ repo!


.travis.yml
```diff
@@ -25,6 +25,7 @@ jobs:
     - travis_retry scripts/setup-tools-for-ubuntu.sh
     script:
     - make test
+    - travis_retry make -C enterprise-suite check-console-api
```
jsravn: Does it make sense to do travis_retry? Doesn't it mean it will fail multiple times in a row?

Also - can we rename the target to something a bit more intuitive. Like "check-console-api-monitors-in-sync" or something.

enterprise-suite/scripts/pull-console-api.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/sh
```
jsravn: `/usr/bin/env bash`

enterprise-suite/scripts/pull-console-api.sh
```diff
@@ -0,0 +1,54 @@
...
+
```
jsravn: Should add `set -eu` ideally.

enterprise-suite/values.yaml
```diff
@@ -90,6 +90,12 @@ exposeServices: false
 # If exposeServices is set to "NodePort" or "LoadBalancer", Console will be exposed on this port.
 esConsoleExposePort: 30080
 
+# Path to custom default monitors file, must be inside the Console Helm Chart directory.
+defaultMonitorsFile:
+
+# Path to custom static rules file, must be inside the Console Helm Chart directory.
+staticRulesFile:
+
```
jsravn: I think this is too limiting - we shouldn't require users to have to check out the repo and do a local chart install to edit the monitors.

Instead these should refer to ConfigMap names - which will let users override with their own ConfigMap.

I'm also thinking we should install the configmaps as well for the default files - and set them. This gives an example to users how it works (rather than it being implicit via the default argument on the image). So the default value for these should be configmaps we create using the copied over files. This also lets users easily copy those files out of the configmaps for tuning.

enterprise-suite/templates/prometheus-deployment.yaml
```diff
@@ -90,8 +90,12 @@ spec:
           args:
             - --configPath=/etc/config/
             - --storagePath=/monitor-data/
-            - --staticConfig=/etc/console-api/static-rules.yml
-            - --defaultMonitorsFile=/etc/console-api/default-monitors.json
+            {{ if .Values.defaultMonitorsFile }}
+            - --defaultMonitorsFile=/etc/{{ .Values.defaultMonitorsFile }}
+            {{ end }}
+            {{ if .Values.staticRulesFile }}
+            - --staticRules=/etc/{{ .Values.staticRulesFile }}
```
marcoderama: I presume there's a corresponding change to console-api to use `--staticRules` instead of `--staticConfig`.

.travis.yml
mitkus: The checking script downloads files from github, so it could fail due to network issues. I'll rename makefile targets to something better.

enterprise-suite/scripts/pull-console-api.sh
mitkus: I'll add `set -u`. The way I'm checking for previous command exit status seems incompatible with `set -e`, the script then exits immediately without printing the reason it failed.

enterprise-suite/templates/prometheus-deployment.yaml
mitkus: Oh, that's just a mistype. Changing to `--staticConfig`.

enterprise-suite/scripts/pull-console-api.sh
mitkus: Fixed

enterprise-suite/values.yaml
mitkus: Updated to use ConfigMaps and to default to configs copied over from `console-api`

.travis.yml
jsravn: The download should happen in a separate retry step in that case.

enterprise-suite/scripts/pull-console-api.sh
jsravn: Right - you need to use `set +e` around the section where you want to manually handle exit code checking.

enterprise-suite/templates/prometheus-deployment.yaml
```diff
@@ -180,9 +185,15 @@ spec:
       volumes:
         - name: config-volume
           emptyDir: {}
+        # By default, console-api-defaults will be mounted twice. Template for prometheus config (prometheus.yml) will always
+        # be picked up from console-api-defaults-volume and default monitors (default-monitors.json, static-rules.yml) from
+        # console-api-volume. This allows to override default monitors volume while keeping default prometheus config template.
+        - name: console-api-defaults-volume
+          configMap:
+            name: console-api-defaults
```
jsravn: I think this is a bit too confusing. Instead can we just create a ConfigMap per console-api file, and let users override each? e.g. `console-api-static-rules`, `console-api-prometheus-config`, `console-api-default-monitors`. Then have a parameter for each that let's a user override them, `console-api { static-rules: console-api-static-rules }`, etc.

Alternatively we can try to be clever and merge a user ConfigMap with our "templates"/"bundled" config map. But that feels a bit too magic.

enterprise-suite/scripts/pull-console-api.sh
mitkus: But that is basically the whole file in this case, so I'll leave it without `set -e`

enterprise-suite/templates/prometheus-deployment.yaml
mitkus: Ok, but should we even allow overriding prometheus config? I think not. I'll expose separate chart values to override the other two.

jsravn: Probably not - but we might need to start making some parts of it configurable, like scrape time.

enterprise-suite/values.yaml
```diff
@@ -90,6 +90,12 @@ exposeServices: false
 # If exposeServices is set to "NodePort" or "LoadBalancer", Console will be exposed on this port.
 esConsoleExposePort: 30080
 
+# Default monitors ConfigMap. Must contain file `default-monitors.json`.
+defaultMonitorsConfigMap: console-api-default-monitors
+
+# Static rules ConfigMap. Must contain file `static-rules.yml`.
+staticRulesConfigMap: console-api-static-rules
```
jsravn: Would it make sense to organize it a bit more hierarchically? I think these are the first console-api specific parameters.

```
consoleAPI:
   defaultMonitorsConfigMap:
   staticRulesConfigMap:
```

jsravn: Not feeling too strongly on this one - but it does feel like our parameters are all over the place at the moment.

mitkus: It would be nice, though all other values that we have are top-level and not hierarchical...

mitkus: Anyway, I grouped them. We should start to group other values into a hierarchy too whenever we make helm chart changes.

PR #373, opened 2019-04-15
Compute values for chart to be installed

jsravn:
For https://github.com/lightbend/console-backend/issues/606

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -12,7 +12,6 @@ if sys.version_info >= (3, 0):
     sys.exit(proc.returncode)
 
 import os
-import glob
 import shlex
 import shutil
 import threading
@@ -23,6 +22,7 @@ import zipfile
 import datetime
 import base64
 import urllib2 as url
+import json
 import time
 from distutils.version import LooseVersion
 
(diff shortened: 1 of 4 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -76,10 +79,6 @@ def fail(msg):
     sys.exit(msg)
 
 
-def make_tempdir():
```
mitkus: Some tests override this function, they will have to be changed if you remove it

jsravn: Yes discovered this when the tests failed...

jsravn:
Ready for re-review.

jsravn:
This is an implementation detail of the PVC work which still needs some good tests - tbd in https://github.com/lightbend/console-backend/issues/608.

jsravn:
Do you mean stubbing out a method? I just find doing that kind of testing is making the maintainability of the install script already quite difficult. I think we need more black box tests for this kind of thing. I will think about it some more.

jsravn:
Unfortunately the `template` command fails if you pass in all the install flags. So I need to prune the flags down to just those that template can use (--set*, --values).

enterprise-suite/scripts/lbc.py
jsravn: Replaced with `make_fetchdir` farther down in the file.

PR #375, opened 2019-04-23
Add gotests for upgrade failure on PVC loss

jsravn:
Also simplify PVC check:
- Remove retained PV checking. We only check the helm flags now, this is
substantially faster and simpler (the PV checking could take 10s+).
- Improve error messages.

Also test conflicting namespace logic.

There is a lot more we could test here, but I think this is a good start
for one PR.

part of https://github.com/lightbend/console-backend/issues/608

### DIFF
--- a/enterprise-suite/gotests/testenv/testenv.go
+++ b/enterprise-suite/gotests/testenv/testenv.go
@@ -5,10 +5,11 @@ import (
 	"os"
 	"time"
 
+	"github.com/lightbend/console-charts/enterprise-suite/gotests/args"
+
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 
-	"github.com/lightbend/console-charts/enterprise-suite/gotests/util/kube"
 	"github.com/lightbend/console-charts/enterprise-suite/gotests/util/lbc"
 	"github.com/lightbend/console-charts/enterprise-suite/gotests/util/minikube"
 	"github.com/lightbend/console-charts/enterprise-suite/gotests/util/oc"
@@ -53,7 +54,7 @@ func InitEnv() {
 	}
 
 	// Setup k8s client
-	config, err := clientcmd.BuildConfigFromFlags("", Kubeconfig)
+	config, err := clientcmd.BuildConfigFromFlags("", args.Kubeconfig)
 	if err != nil {
 		Expect(err).To(Succeed(), "new k8sclient config")
 	}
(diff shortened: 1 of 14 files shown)


jsravn:
Regarding removing retained PV checking, I didn't think it was buying us a lot. The logic is quite complicated (I was struggling to understand all the conditional flow), it wasn't quite correct (it would return all PVs in the cluster - dead ones and ones in other namespaces), and the only effect it seemed to have was in not requiring `--delete-pvcs` when PV was set to `Retained`. The other problem is that this checking was causing the installer to take upwards of 10s additional time in my own tests in a real cluster.

Simplification here is to always require `--delete-pvcs` if the helm flag `usePersistentVolumes` changes from true to false. This is a quick check and should cover all potential data loss scenarios.

enterprise-suite/gotests/testenv/testenv.go
```diff
@@ -62,24 +63,14 @@ func InitEnv() {
...
-			if kube.StorageClassExists(K8sClient, storageClass) {
-				foundStorageClass = true
-				helmArgs = append(helmArgs,
-					fmt.Sprintf("--set usePersistentVolumes=true,defaultStorageClass=%v", storageClass))
-				break
-			}
-		}
-		if !foundStorageClass {
-			helmArgs = append(helmArgs, "--set usePersistentVolumes=false")
-		}
```
jsravn: I don't think this is necessary now that we have support the default storage class.

enterprise-suite/gotests/util/lbc/lbc.go
```diff
@@ -4,27 +4,28 @@ import (
...
+	"github.com/lightbend/console-charts/enterprise-suite/gotests/args"
+
 	"github.com/lightbend/console-charts/enterprise-suite/gotests/util"
 )
 
 const localChartPath = "../../../."
 const lbcPath = "../../../scripts/lbc.py"
 
-func Install(namespace, tillerNamespace string, lbcArgs, helmArgs []string) error {
+func Install(lbcArgs, helmArgs []string) error {
```
jsravn: I think I made this change originally... to move `args` into `testenv`. But now it's back to the way it was :-). It seems redundant to have to specify the namespaces everywhere, as they are fixed based on the installation environment.

enterprise-suite/scripts/lbc.py
```diff
@@ -352,146 +352,45 @@ def check_resource_list(cmd, expected, fail_msg):
     return False
 
 
-# Returns two lists.  For each of the PVC types we care about, the associated PV
-# is in the first or second list if the reclaim policy is RETAIN or not respectively.
-# Each list element is a tuple of PV name, claim, and status.
-def pvs_retained_and_not(namespace):
```
mitkus: It's a relief to see this deleted.

enterprise-suite/gotests/testenv/testenv.go
mitkus: It will no longer work on minishift without default storage class, but I think that's acceptable. Should we even have an option of running console with `emptyDir`? It could slightly simplify helm charts not to have it at all.

jsravn: Do we need to support minishift with gotests? I didn't think so.

jsravn: I think we should still support emptyDir "in case". It's not a huge burden.

PR #376, opened 2019-04-25
Add labels to all resources

jsravn:
Also rename things to be more appropriate:
- Remove legacy es prefix where possible
- Group frontend stuff in "console-frontend"
- Group backend stuff in "console-backend" (rather than calling it prometheus - makes it clear it is our console backend and not a prometheus installation)

For https://github.com/lightbend/console-backend/issues/491


enterprise-suite/templates/alertmanager-configmap.yaml
```diff
@@ -3,6 +3,11 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   labels:
+    app.kubernetes.io/name: {{ template "name" . }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: alertmanager
     app: prometheus
```
mitkus: Other resources no longer have `app:` and `component:` labels. Is there a reason to have them for alertmanager configmap and deployment?

jsravn: Not for the `ConfigMap`, but for the `Deployment` it's necessary to preserve backwards compatibility for upgrades.

PR #379, opened 2019-05-03
Update to console-api 1.1.1

jsravn:
Adds a default 1m warmup period to all monitors.

Also:
- Resync default monitors
- Improve sync script to use `$HOME/.config/hub` if it exists.
- Add e2e test for warmup, and also some refactoring and improvements to some of the tests (e.g. wrapping alertmanager tests in busy wait so they don't flake).

For https://github.com/lightbend/console-backend/issues/653


jsravn:
Blocked on https://github.com/lightbend/console-frontend/issues/684 for now.

jsravn:
Hmm I suppose we could merge this, and just set warmup to 0 until the UI side is sorted.

enterprise-suite/gotests/tests/alertmanager/alertmanager_test.go
```diff
@@ -146,52 +174,75 @@ var _ = Describe("all:alertmanager", func() {
...
-				if val, ok := alert.Labels["alertname"]; ok {
-					alertNames = append(alertNames, val)
+			err := util.WaitUntilSuccess(util.LongWait, func() error {
+				alerts, err := alertm.Alerts()
+				Expect(err).ToNot(HaveOccurred())
+				var alertNames []string
+				var found bool
+				for _, alert := range alerts {
+					if val, ok := alert.Labels["alertname"]; ok {
+						alertNames = append(alertNames, val)
```
mitkus: Looks like there's no more need to add firing alerts to the list if the test now uses `found` variable.

enterprise-suite/gotests/tests/alertmanager/alertmanager_test.go
```diff
@@ -146,52 +174,75 @@ var _ = Describe("all:alertmanager", func() {
...
 				}
-			}
-			Expect(alertNames).To(ContainElement(name))
 
-			deleteAlert(name)
+				return nil
+			})
+			Expect(err).ToNot(HaveOccurred())
+
+			//deleteAlert(name)
```
mitkus: Why not delete the alert here?

enterprise-suite/gotests/tests/installer/installer_test.go
```diff
@@ -50,25 +50,38 @@ var _ = Describe("all:lbc.py", func() {
 
 	Context("upgrades", func() {
 		Context("disable persistent volumes", func() {
+			var installer *lbc.Installer
+
 			BeforeEach(func() {
-				Expect(lbc.Install([]string{}, []string{"--set usePersistentVolumes=true"})).To(Succeed(), "install with PVs")
+				preInstaller := lbc.DefaultInstaller()
+				preInstaller.UsePersistentVolumes = "true"
```
mitkus: I like how much more readable this is now

enterprise-suite/gotests/util/lbc/lbc.go
```diff
@@ -12,13 +14,44 @@ import (
 const localChartPath = "../../../."
 const lbcPath = "../../../scripts/lbc.py"
 
-func Install(lbcArgs, helmArgs []string) error {
+type Installer struct {
+	UsePersistentVolumes string
```
mitkus: It would be more straightforward if this was a `bool`

jsravn: Not quite so simple - I need the "not present" value here, so we don't set it at all (because a `--set` overrides a `-f`). I would have to use a pointer to bool, and set it to nil to not set. I thought a string was simpler.

enterprise-suite/gotests/tests/alertmanager/alertmanager_test.go
jsravn: Isn't it worth it for the logging when the test fails?

enterprise-suite/gotests/tests/alertmanager/alertmanager_test.go
jsravn: Fixed... leftover from debugging.

enterprise-suite/gotests/util/lbc/lbc.go
jsravn: Note the tests where this is used are testing out the values file (PVC detection logic).

enterprise-suite/gotests/tests/alertmanager/alertmanager_test.go
mitkus: Oh, didn't notice that at first

PR #381, opened 2019-05-15
Update console-api to v1.1.3

mitkus:


jsravn:
monitor history spec seems to be blowing up

PR #390, opened 2019-05-29
Fix lbc.py --export-yaml.

marcoderama:
Proposed fix for https://github.com/lightbend/console-backend/issues/691


marcoderama:
This should get some thorough testing.  It fixes things for the `--export-yaml` case but affects the `install` logic in general...

jsravn:
Thanks. Yes, a test would be great if we’re missing it.

PR #391, opened 2019-05-29
Ignore errors from templating dump-values.yaml

marcoderama:
Proposed fix for https://github.com/lightbend/console-backend/issues/693


jsravn:
Thanks @marcoderama!

PR #392, opened 2019-05-31
Fix debug-dump zipfile to contain logs

jsravn:
It was incorrectly storing `kubectl get pods` in the logfiles. This is
fixed now. A workaround prior to this change is to use `--print` which
works correctly.

Also improve debug-dump logs. Print to stderr as its collecting
information, so it's clear it's not just hanging.

Also limit logs to last 250 lines to prevent giant debug files.

Also fix debug-dump failing if logs fail to retrieve, which can happen
when a pod is crash looping.

Fix https://github.com/lightbend/console-backend/issues/694


jsravn:
Force merging, to address customer issue as @mitkus is out. @mitkus please review post-hoc when you can :).

PR #394, opened 2019-06-12
Make install script backwards compatible with old deployment names

mitkus:
For https://github.com/lightbend/console-backend/issues/699

This makes install script backwards compatible with old deployment names, for example, those used in 1.1 and previous versions.

Instead of parsing version string to decide what to check, I opted to use a more fool-proof approach - check old names if new ones fail. This prints out more stuff when `verify` is done on old versions, but should work the same as the current installer in future versions.


mitkus:
We have verify test in gotests. It didn't catch this issue because it tests on master branch local chart, not older versions. Not sure if it's worth adding older versions test too, as that increases test running time significantly...

PR #396, opened 2019-06-14
Disable frontend monitor-history test due to disable monitor edit by default

katsutoxin:
connect to https://github.com/lightbend/console-frontend/issues/698


jsravn:
@katsutoxin what's the mechanism to enable/disable the edit monitor dialogue?

katsutoxin:
> @katsutoxin what's the mechanism to enable/disable the edit monitor dialogue?

@jsravn Good point.  I just add this part here https://github.com/lightbend/console-charts/pull/396/files#diff-5f926e548bdb8f0adc9d446a676d740d

katsutoxin:
Hi @jsravn ,  could you take a look why the openshift e2e go test is failed?  Thanks!

enterprise-suite/values.yaml
```diff
@@ -3,6 +3,7 @@
 #
 consoleUIConfig:
   logo: ""
+  isMonitorEditEnabled: false
```
pcraddick: I imagine you named this the same as the field in our env, to minimize confusion – which is a worthy goal. This is a bit of a mouthful though for a customer. Maybe it doesn't matter, since we don't think (m)any customers will be leveraging this. If they do, we probably want a more concise field.

katsutoxin: yeah.  I use this name is just to be the same name as our env field.  If you want, I can change both UI config field and env field to shorter name like 'enableMonitorEdit' .

pcraddick: If customers start using this, maybe we'll change it; or might end up just being a non-issue

jsravn:
Build is failing due to central park 2 having problems.

jsravn:
Asked @lightbend/play-team to fix it - looks like their build is filling up CP2.

jsravn:
CP2 should be working again.

PR #398, opened 2019-06-19
Use templated value for `apiVersion` in all resources

mitkus:
For https://github.com/lightbend/console-backend/issues/695

This fixes up the few remaining deployments to use the appropriate `apiVersion` templated values.

This PR also has minor stability improvements for some tests.


mitkus:
Ok, templating service and configmap api versions maybe a bit too much. I reverted that commit so now this PR touches only deployments.

PR #410, opened 2019-07-29
Deploy chaos apps to demo environment

jsravn:
So we can show our monitors in action.


jsravn:
@mitkus what do you mean? You mean we won't deploy often enough? I think that's okay - we don't want to redeploy on every commit, as that will cause us to lose history in the monitors. I think it's okay as-is since we don't expect the chaos apps to change very frequently in the future.

I thought about deploying it from the console-chaos-apps repo, but I think this is uniquely for our console-demo so that doesn't make sense either.

mitkus:
I just mean that both this repo and `console-chaos-apps` has k8s resource yamls for these demos and it might be hard to keep them in sync. Not a big deal really, we just have to remember to update resources here too if we want to see the changes in demo deployments.

PR #413, opened 2019-08-02
Move default.conf into es-console-spa

jsravn:
As part of the work to reduce the size of the console charts repo, and
to keep the UI code more self-contained in a single repo. The helm chart
keeps version.js since that is the helm version.

For https://github.com/lightbend/console-backend/issues/735
Depends on https://github.com/lightbend/es-console-spa/pull/535


jsravn:
This needs the new console spa image to pass with the changes I made.

PR #426, opened 2019-08-16
Add elasticsearch service to console-charts

mitkus:
For https://github.com/lightbend/console-backend/issues/743

This needs a change in `es-console-spa` that exposes `/services/elasticsearch` http path: https://github.com/lightbend/es-console-spa/pull/542


seglo:
@jsravn Is it possible to get an internal release of Console so that we can test integration with ES in Pipelines?

jsravn:
@seglo this is already available in 1.2.2 and the nightly chart. Need `enableElasticsearch` set to true.

seglo:
@jsravn Great, thanks.  We'll update our build with that version.

PR #435, opened 2019-08-30
Put console-operator docker image in a public repository

mitkus:
This reduces one step in installing Lightbend Console using operator. Other console images are still in the commercial registry, so setting up credentials is needed but it can be done via fields in the CR now.


enterprise-suite/console-operator/deploy/crds/console_v1alpha1_console_cr.yaml
```diff
@@ -16,13 +16,15 @@ spec:
   # Update these versions for a release.
   #
   # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-console
-  esConsoleVersion: v1.2.5
+  esConsoleVersion: v1.2.6
   # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/console-api
-  esMonitorVersion: v1.2.2
+  esMonitorVersion: v1.2.3
```
jsravn: Isn't there something that checks these are up to date? How come we're changing them in this PR?

mitkus: Yes, `verify-operator` build step checks that operator resources are in-sync with helm charts.

PR #436, opened 2019-08-30
Disable Elasticsearch by default, fix Elasticsearch volume

jsravn:
Also re-enable disabled installer test.

For https://github.com/lightbend/console-backend/issues/750


enterprise-suite/gotests/tests/elasticsearch/elasticsearch_test.go
```diff
@@ -0,0 +1,35 @@
...
+})
+
+var _ = AfterSuite(func() {
+	testenv.CloseEnv()
+})
+
+var _ = Describe("all:elasticsearch", func() {
+	It("can access Elasticsearch", func() {
+		addr := fmt.Sprintf("%v/%v", testenv.ConsoleAddr, "service/elasticsearch/")
+		Expect(urls.Get200(addr)).To(Succeed(), addr)
```
mitkus: This is already covered by tests/nginx_test.go:47

jsravn: I removed that, because the default install doesn't install ES anymore. So decided to put it in a dedicated test here instead when ES is enabled.

PR #439, opened 2019-09-09
Updated rule to detect flink workload

seglo:
New tested rule to ID Flink workloads.

### DIFF
--- a/enterprise-suite/console-api/prometheus.yml
+++ b/enterprise-suite/console-api/prometheus.yml
@@ -84,9 +84,10 @@ scrape_configs:
         regex: (.*)-[0-9]{13}-(exec|driver)-[0-9]+;
         replacement: ${1}-${2}
         target_label: es_workload
-    # flink workload — of the form 'taxi-ride-fare-seglo-processor-dde77c7d-jm' or 'taxi-ride-fare-seglo-processor-dde77c7d-tm'
-      - source_labels: [kubernetes_pod_name, es_workload]
-        regex: (.*)-[a-z,0-9]{8}-(jm|tm);
+      # flink workload — of the form 'taxi-ride-fare-seglo-processor-dde77c7d-jm' or 'taxi-ride-fare-seglo-processor-dde77c7d-tm'
+      - source_labels: [kubernetes_pod_name]
+        action: replace
+        regex: '(.*)-[a-z,0-9]{8}-(jm|tm)-[456789bcdf]{4,10}-[bcdfghjklmnpqrstvwxz2456789]{5}'
         replacement: ${1}-${2}
         target_label: es_workload
       # extract workload from workload-num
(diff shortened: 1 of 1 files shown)


enterprise-suite/console-api/prometheus.yml
```diff
@@ -84,9 +84,10 @@ scrape_configs:
         regex: (.*)-[0-9]{13}-(exec|driver)-[0-9]+;
         replacement: ${1}-${2}
         target_label: es_workload
-    # flink workload — of the form 'taxi-ride-fare-seglo-processor-dde77c7d-jm' or 'taxi-ride-fare-seglo-processor-dde77c7d-tm'
-      - source_labels: [kubernetes_pod_name, es_workload]
-        regex: (.*)-[a-z,0-9]{8}-(jm|tm);
+      # flink workload — of the form 'taxi-ride-fare-seglo-processor-dde77c7d-jm' or 'taxi-ride-fare-seglo-processor-dde77c7d-tm'
+      - source_labels: [kubernetes_pod_name]
```
jsravn: Can you add back the `, es_workload`? It probably doesn't matter much here, but it maintains consistency of the ordering precedence these matching are using (e.g. if it matches an earlier rule, it won't then match this rule).

enterprise-suite/console-api/prometheus.yml
```diff
@@ -84,9 +84,10 @@ scrape_configs:
...
+        action: replace
+        regex: '(.*)-[a-z,0-9]{8}-(jm|tm)-[456789bcdf]{4,10}-[bcdfghjklmnpqrstvwxz2456789]{5}'
```
jsravn: Are you sure this is right? It doesn't seem to match any of the examples prior.

jsravn:
Looking into the build, unfortunately it's one of the poorly written tests failing (where it uses `count(something) == 1`, which is a holdover from the original bash based scripts). These are hard to see what's going wrong, so I'll dig into it.

enterprise-suite/console-api/prometheus.yml
seglo: It looked like rules match in a cascade.  One rule can match, then a subsequent one, etc.  It was difficult to come up with a rule that output the right value to `es_workload` for flink because it can match on the general pod rule as well.

We could add `es_workload` back, but then the regex would just become the following, to skip matching on `es_workload` values that were produced from before.

```
;(.*)-[a-z,0-9]{8}-(jm|tm)-[456789bcdf]{4,10}-[bcdfghjklmnpqrstvwxz2456789]{5}
```

enterprise-suite/console-api/prometheus.yml
seglo: I've deleted the example.

seglo:
It seems like these rules still don't make ES workloads show up correctly.  I've confirmed that the pod metrics have the right values for `es_workload`, but it seems that the recording rules that Console relies on to organize the topology are still incorrect.  It would be ideal if someone from the Console backend team could look at this in more detail. /cc @pcraddick

PR #440, opened 2019-09-09
Improve helm based operator

jsravn:
There are some bugs to fix:

- Makefile was calling the wrong target when validating the operator files hadn't changed
- `lightbend` namespace is hardcoded into the cluster role binding - thus it will only work in the `lightbend` namespace. Need to make this configurable.
- The build script copies the manually created files out of the way - this is sort of weird and can mess up code editors (file link changes). Going to change this so the operator directory is purely generated out of source files.
- kustomization.yaml should configure the namespace and watched namespaces.
- releasing the operator has a race condition with releasing the helm chart as it depends on the external helm chart - it should just use the local helm chart files so it is a self contained release process.

For https://github.com/lightbend/console-backend/issues/744


operator/build-operator.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+rm -rf build && mkdir build
+
+cd build
+operator-sdk new console-operator --type=helm --kind=Console --api-version=${api_version} \
+    ${new_operator_args}
+
+cd console-operator
+operator-sdk build "${full_docker_name}"
```
marcoderama: Do you worry that someone might ever accidentally create a `build` directory for other reasons and get a nasty surprise?  Just wondering if it might be an idea to use a more unusual name here.  Or not.

marcoderama:
And I just noticed now that this is a WIP...

jsravn:
It should be complete now @marcoderama although I see the build is failing. Will have a look tomorrow.

enterprise-suite/console-operator/deploy/crds/console.lightbend.com_consoles_crd.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
marcoderama: I'm confused.  How does this file relate to the renamed 
`enterprise-suite/console-operator/deploy/crds/console_v1alpha1_console_crd.yaml → operator/manifests/crds/console_v1alpha1_console_crd.yaml`
below?  They look to be essentially the same thing.  (That plural `_consoles` on the end of this filename threw me btw.)

(Also, I note that `operator/manifests/crds/console_v1alpha1_console_crd.yaml` specifies both a `version` and `versions`.  I think you got rid of the `version` line (from somewhere) in stuff I reviewed earlier.  Not sure if you meant it to be gone now or not.)

operator/README.md
```diff
@@ -0,0 +1,16 @@
+# Console Operator
+
+This directory provides files to install the Console Operator. To install, use the kustomization.yaml file:
```
marcoderama: Do you want to clarify that this just installs the operator?  Do you want to describe (or just provide a link to) how to use this to actually install _console_?

operator/kustomization.yaml
```diff
@@ -0,0 +1,26 @@
...
+resources:
+- manifests/crds/console_v1alpha1_console_crd.yaml
+- manifests/service_account.yaml
+- manifests/role.yaml
+- manifests/role_binding.yaml
+- manifests/operator.yaml
+- src/cluster_role.yaml
+- src/cluster_role_binding.yaml
+
+patchesJson6902:
```
marcoderama: Is the reason for doing this worthy of a comment?

PR #443, opened 2019-09-18
Fixes for OLM support

jsravn:
- Change image credentials to always be created, so we can't get an invalid `Secret` via OLM. This however means Helm always manages the credentials now. This should be okay because it's the only mechanism we document.
- Check in the OLM CSV and test files so it's easier to test without having to recreate from scratch. Eventually this should be automated.


marcoderama:
Hmmm.  I approved that last one-liner change.  Reviewing the rest now...

PR #449, opened 2019-09-27
Update CSV to match community-operators

jsravn:


operator/olm/lightbend-console-operator.v0.0.1/lightbend-console-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -103,7 +103,6 @@ spec:
     credentials to the Console custom resource.
   maturity: alpha
   version: 0.0.1
-  replaces: lightbend-console-operator.v0.0.1
```
marcoderama: Yeah, I've only ever had problems when I included `replaces`.  Not sure what's up with that.

PR #451, opened 2019-10-01
enable both TCP and UDP port in logstash

katsutoxin:
After discussed with @jsravn and @pvlugter  , we feel that TCP logstash connection is better over UDP.  Support both TCP and UDP for better compatibility. 

I have validated that both TCP and UDP work end to end with this setting.

The akka cluster changes for TCP connection is here https://github.com/lightbend/console-akka-cluster-explore/pull/15

### DIFF
--- a/enterprise-suite/console-api/logstash.conf
+++ b/enterprise-suite/console-api/logstash.conf
@@ -3,6 +3,10 @@ input {
     port => 5000
     codec => json
   }
+  tcp {
+    port => 5000
+    codec => json
+  }
 }
 
 output {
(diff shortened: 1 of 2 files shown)


katsutoxin:
@jsravn   I don't understand why ci tests are failed.  Could you take a look?

katsutoxin:
@steinybot  I updated this PR to support both TCP and UDP connection. Could you kindly review it again. Thanks :)

jsravn:
It's failing due to kubecfg now requiring at least golang 1.12. Fixing it in my branch, once it's merged you can rebase to fix.

PR #460, opened 2019-11-06
Support namespaces opt-in to scrape

katsutoxin:
To test it in local

step 1. create es-demo app in namespace demo1/demo2/demo3
```
kubectl create ns demo1
kubectl create ns demo2
kubectl create ns demo3
kubectl apply -f ./enterprise-suite/tests/e2e/config/es-demo.yaml -n demo1
kubectl apply -f ./enterprise-suite/tests/e2e/config/es-demo.yaml -n demo2
kubectl apply -f ./enterprise-suite/tests/e2e/config/es-demo.yaml -n demo3
```

step 2. 

modify file `enterprise-suite/values.yaml` , 
change line 
```
kubeStateMetricsScrapeNamespaces: ""
```
to
```
kubeStateMetricsScrapeNamespaces: "demo1,demo3"
```
step 3. 
Install lightbend console with command 
```
cd enterprise-suite
./scripts/lbc.py install --local-chart=. --namespace lightbend  --set exposeServices=NodePort
```

step 4. check ui and see if it scrape demo1 and demo3 namespaces

-----------------

The usage scenario we imagine for customers would be to define a local yaml file that provides the override for ```kubeStateMetricsScrapeNamespaces``` only; e.g. values.yaml like so:
```kubeStateMetricsScrapeNamespaces: "demo1,demo3"```

You then reference this file, as an override/overlay, like so:

```./scripts/lbc.py install --local-chart=. --namespace lightbend  --set exposeServices=NodePort -- --values=values.yaml```

This example is predicated on having the console-charts pulled locally; so we need to verify that there is no issue when a customer pulls down our install script only and follows the same pattern

### DIFF
--- a/enterprise-suite/templates/kube-state-metrics-deployment.yaml
+++ b/enterprise-suite/templates/kube-state-metrics-deployment.yaml
@@ -45,6 +45,7 @@ spec:
           image: {{ .Values.kubeStateMetricsImage }}:{{ .Values.kubeStateMetricsVersion }}
           args:
             - --port=8080
+            - --namespace={{ .Values.kubeStateMetricsScrapeNamespaces }}
             - --telemetry-port=8081
           resources:
             requests:
@@ -53,4 +54,3 @@ spec:
           ports:
             - name: metrics
               containerPort: 8080
-
--- a/enterprise-suite/gotests/tests/prometheus/prometheus_test.go
+++ b/enterprise-suite/gotests/tests/prometheus/prometheus_test.go
@@ -155,6 +155,7 @@ var _ = Describe("all:prometheus", func() {
 
 	Context("kube-state-metrics", func() {
 		It("should only scrape a single instance", func() {
+			time.Sleep(11 * time.Second)
 			query := fmt.Sprintf(`kube_pod_status_ready{namespace="%s", es_workload="console-backend", condition="true"} == 1`, args.ConsoleNamespace)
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: enterprise-suite/templates/kube-state-metrics-deployment.yaml [hunk 1]
```diff
@@ -45,6 +45,7 @@ spec:
           image: {{ .Values.kubeStateMetricsImage }}:{{ .Values.kubeStateMetricsVersion }}
           args:
             - --port=8080
+            - --namespace={{ .Values.kubeStateMetricsScrapeNamespaces }}
```
pcraddick: Don't we want a conditional check here, so that if nothing is specified we don't set this option at all? Or is the blank value equivalent to wildcard?

katsutoxin: We don't need the condition check.   Based on the official docs, if we don't set it, it will use the default value "".  We set this value to default value in values.yaml already

Here is the official doc
https://github.com/kubernetes/kube-state-metrics/blob/master/docs/cli-arguments.md

longmuir:
Nice work @katsutoxin! 

@BillyAutrey As an FYI, this should allow us to control which namespaces Console scrapes. Any feedback?

Also, do you think it will address this request https://github.com/lightbend/console-charts/issues/459? If it does, great, otherwise, this is still excellent.

PR #463, opened 2019-11-21
Add helm 3 support

steinybot:
All of the installer subcommands should now work with both helm 2 and 3.

If you do an install of console 1.2.4 with helm 2 and then try and to an install of 1.2.5 with helm 3 it should detect the old installation and fail as manual migrations steps are required. Once doing the migration the install should then succeed.

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -639,8 +639,8 @@ def install(creds_file):
                     execute('helm install {} {} {}'.format(args.helm_name, chart_name, full_args))
     finally:
         if tempdir:
-            if args.keep_temp:
-                printout('info: Keeping temporary directory: ', tempdir)
+            if args.keep_chart:
+                printout('info: Keeping downloaded chart: ', tempdir)
             else:
                 shutil.rmtree(tempdir)
 
@@ -866,7 +866,7 @@ def setup_args(argv):
                          action='store_true')
     install.add_argument('--set', help='set a helm chart value, can be repeated for multiple values', type=str,
                          action='append')
-    install.add_argument('--keep-temp', help='does not delete the temporary directory', action='store_true')
+    install.add_argument('--keep-chart', help='does not delete the downloaded chart', action='store_true')
 
 
     # Common arguments for install and uninstall
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -30,6 +30,8 @@ from distutils.version import LooseVersion
 # Minimum required dependency versions
 REQ_VER_KUBECTL = '1.10'
 REQ_VER_HELM = '2.10'
+# Helm tends to break backwards compatibility this is the maximum version.
+MAX_VER_HELM = '3.x'
 REQ_VER_MINIKUBE = '0.29'
 REQ_VER_MINISHIFT = '1.20'
 REQ_VER_OC = '3.9'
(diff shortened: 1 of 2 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -132,31 +135,48 @@ def execute(cmd, can_fail=False, print_to_stdout=False):
...
         match = version_re.search(stdout)
         if match != None:
             current = LooseVersion(match.group())
             required = LooseVersion(required_version)
             if current >= required:
-                return
+                if max_version is None or current < LooseVersion(max_version):
+                    return match.groupdict()
+                else:
+                    fail("Installed version of {} is too new. Found: {}, required: {}"
```
marcoderama: Perhaps "require at most" instead of "required"?

enterprise-suite/scripts/lbc.py
```diff
@@ -132,31 +135,48 @@ def execute(cmd, can_fail=False, print_to_stdout=False):
...
+                        .format(name, current, max_version))
             else:
-                fail("Installed version of '" + name + "' is too old. Found: {}, required: {}"
-                     .format(current, required))
+                fail("Installed version of {} is too old. Found: {}, required: {}"
```
marcoderama: Perhaps "require at least" instead of "required"?

enterprise-suite/scripts/lbc.py
```diff
@@ -276,10 +301,32 @@ def check_new_install_script():
...
+               '--ignore-not-found=true'.format(args.tiller_namespace, args.helm_name))
+        printout(cmd)
+        returncode, stdout, _ = run(cmd)
+        if returncode != 0:
+            fail('\nwarning: Failed to check for legacy helm 2 installations.')
+        elif stdout != '':
+            message = (
+                '\nerror: Detected a previous Console installation which was installed with helm 2 '
+                'which has not been migrated to helm 3.\n'
+                'You must migrate these manually by following the instructions on '
```
marcoderama: I find this wording a bit confusing.  Perhaps something along the lines of? :

Detected a previous Console installation which was installed with helm 2.  To use helm `helm_version` you must migrate the installation manually following the instructions on [...].

enterprise-suite/scripts/lbc.py
steinybot: What about `Installed version of helm is too new. Found 4.0.0 but the newest supported version is 3.x.`?

enterprise-suite/scripts/lbc.py
steinybot: And how about `Installed version of helm is too old. Found 2.9.0 but the oldest supported version is 2.10.`?

enterprise-suite/scripts/lbc.py
marcoderama: That's nicer.  Maybe "latest supported version"?

enterprise-suite/scripts/lbc.py
marcoderama: That's nicer.

enterprise-suite/scripts/lbc_test.py
```diff
@@ -130,41 +134,103 @@ def tearDown(self):
...
+            setup_checks()
+        expect_cmd('helm repo add es-repo ' + repo)
+        expect_cmd('helm repo update')
+        expect_cmd(r'helm fetch .* es-repo/enterprise-suite')
+        expect_cmd(r'helm template .*')
+
+    @abstractmethod
+    def uninstall(self, args=[]):
+        pass
+
```
pcraddick: 👍 For tests

enterprise-suite/scripts/lbc.py
```diff
@@ -276,10 +302,31 @@ def check_new_install_script():
...
+               '--ignore-not-found=true'.format(args.tiller_namespace, args.helm_name))
+        printout(cmd)
+        returncode, stdout, _ = run(cmd)
+        if returncode != 0:
+            fail('\nwarning: Failed to check for legacy helm 2 installations.')
+        elif stdout != '':
+            message = (
+                '\nerror: Detected a previous Console installation which was installed with helm 2.\n'
+                'To use helm {} you must migrate the installation manually by '.format(helm_version) +
+                'following the instructions on https://helm.sh/blog/migrate-from-helm-v2-to-helm-v3/'
```
pcraddick: 👍 This is cool (and ran down this scenario locally to see it in action)

PR #470, opened 2020-03-30
support k8s 1.16

a1kemist:
This is an initial attempt to update the helm chart for k8s `1.16.x`.

- Updates the api versions to `apps/v1` due to the removal of the beta apis. 
- Updates cadvisor metrics that have been renamed (see the [release notes](https://v1-16.docs.kubernetes.io/docs/setup/release/notes/#removed-metrics)):
  - `pod_name` -> `pod`
  - `container_name` -> `container`
- Updates `kube-state-metrics` to v1.9.0 (the minimum version to support k8s `1.16.x`)
- Updates the rbac role:
  - Adds the proper permissions to collect cadvisor metrics (see https://github.com/linkerd/linkerd2/pull/1142).
  - Adds additional apiGroups for resources that `kube-state-metrics` now accesses.

### DIFF
--- a/enterprise-suite/console-api/prometheus.yml
+++ b/enterprise-suite/console-api/prometheus.yml
@@ -53,18 +53,18 @@ scrape_configs:
 
     # These are metric_relabel_configs since the source labels come from the scraped `/metrics`.
     metric_relabel_configs:
-      - source_labels: [pod_name]
+      - source_labels: [pod]
         target_label: kubernetes_pod_name
       # pause containers have all the network stats for a pod
-      - source_labels: [container_name, __name__]
+      - source_labels: [container, __name__]
         regex: POD;container_(network).*
-        target_label: container_name
+        target_label: container
       # drop all other pause container stats
-      - source_labels: [container_name]
+      - source_labels: [container]
         regex: POD
         action: drop
       # drop system containers with no name
-      - source_labels: [container_name]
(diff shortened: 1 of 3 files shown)


a1kemist:
@steinybot we have been using a forked version of this in development and production for some time, is there any chance this PR could get some feedback/review?

marcoderama:
Have a look at PR https://github.com/lightbend/console-charts/pull/476 @a1kemist.  This has much the same changes as your PR but
* Works with k8s < 1.16 as well...
* ...as long as the user overrides things as appropriate at install time.  (cf. https://github.com/lightbend/console-charts/pull/476#issuecomment-669521064)

So a couple of your changes, in particular to `enterprise-suite/values.yaml `, are the responsibility of the user.

I think those changes address the same issues as yours.  If you agree, then we can close this one I'd say.  Thanks for the guide posts!

marcoderama:
This should be fixed with the Console 1.2.11 release.  Note that for k8s 1.16, you'll need to override the following configs with a `values.yaml` file with the contents:

```
kubeStateMetricsImage: quay.io/coreos/kube-state-metrics
kubeStateMetricsVersion: v1.9.7
deploymentApiVersion: apps/v1
daemonSetApiVersion: apps/v1
```
Then when installing with `lbc.py`, make sure to tack `-- --values=values.yaml` on to the end of the command.

PR #476, opened 2020-07-31
Changes for Cluster Metrics graphs (k8s <= 1.16)

marcoderama:
These changes, in conjunction with appropriate `values.yaml` mods, get the Cluster Metrics page to work in Grafana.

These changes are appropriate for the previous k8s versions we used to work with and now also up to and including 1.16.  _For k8s 1.15 and beyond, the user must create a `values.yaml` file to apply at install time.  cf. https://github.com/lightbend/console-charts/pull/476#issuecomment-669521064._

---------

(This is the original comment.  Ignore now...)

~~This is a Draft PR at this time as I don't think these changes will be backwards compatible.  (ie won't work on k8s < 1.14)~~

If trying to install, use a `values.yaml` file with, for now:
```\kubeStateMetricsImage: quay.io/coreos/kube-state-metrics
kubeStateMetricsVersion: v1.9.7

deploymentApiVersion: apps/v1
daemonSetApiVersion: apps/v1
```

### DIFF
--- a/enterprise-suite/console-api/prometheus.yml
+++ b/enterprise-suite/console-api/prometheus.yml
@@ -53,18 +53,18 @@ scrape_configs:
 
     # These are metric_relabel_configs since the source labels come from the scraped `/metrics`.
     metric_relabel_configs:
-      - source_labels: [pod_name]
+      - source_labels: [pod]
         target_label: kubernetes_pod_name
       # pause containers have all the network stats for a pod
-      - source_labels: [container_name, __name__]
+      - source_labels: [container, __name__]
         regex: POD;container_(network).*
-        target_label: container_name
+        target_label: container
       # drop all other pause container stats
-      - source_labels: [container_name]
+      - source_labels: [container]
         regex: POD
         action: drop
       # drop system containers with no name
-      - source_labels: [container_name]
+      - source_labels: [container]
         regex: ^$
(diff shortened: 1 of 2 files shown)


marcoderama:
Making this a real PR now.  These changes work for k8s 1.13 - 1.16 _as long as the proper `kube-state-metrics` version is applied_ with a suitable `values.yaml` file at install time.

   * 1.13 - no changes required but one may want to upgrade to
```
kubeStateMetricsVersion: v1.8.0
```
   * 1.14 - no changes required but one may want to upgrade to
```
kubeStateMetricsVersion: v1.8.0
```
   * 1.15
```
kubeStateMetricsVersion: v1.8.0
```
   * 1.16
```
kubeStateMetricsImage: quay.io/coreos/kube-state-metrics
kubeStateMetricsVersion: v1.9.7
deploymentApiVersion: apps/v1
daemonSetApiVersion: apps/v1
```

marcoderama:
Note that the additional permissions in `enterprise-suite/templates/kube-state-metrics-role.yaml` are irrelevant but not harmful in the older versions of k8s.  Some were needed to work with k8s 1.15, and more were required for k8s 1.16.

marcoderama:
We'll have to update the install docs accordingly when there's a new release.

PR #479, opened 2020-08-27
Move to Prometheus 2.19.3

marcoderama:
Prometheus versions newer than 2.19.3 manage the Prometheus WAL file differently, so one wouldn't (easily) be able to go back to v2.9.2 if they wanted.  cf. https://github.com/prometheus/prometheus/releases/tag/v2.20.0

We'll have to test thoroughly before we go beyond this.


marcoderama:
(Partial) fix for https://github.com/lightbend/console-home/issues/977

katsutoxin (approved):
Awesome !

PR #482, opened 2020-09-24
Update minikube version used in tests

marcoderama:
Had this lying around.  We really should update this...

### DIFF
--- a/enterprise-suite/tests/e2e/scripts/wait_es_services.sh
+++ b/enterprise-suite/tests/e2e/scripts/wait_es_services.sh
@@ -26,5 +26,8 @@ function polling_service {
   echo "$1 is up"
 }
 
+set -x
+echo "SERVICES: $SERVICES"
 polling_service $ES_CONSOLE_URL
+set +x
 echo "poll services takes $seconds sec"
--- a/enterprise-suite/tests/e2e/scripts/wait_es_services.sh
+++ b/enterprise-suite/tests/e2e/scripts/wait_es_services.sh
@@ -5,7 +5,7 @@ script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 # wait es_services are ready. polling es-prom-server/es-console/es-grafana
 SERVICES=$(minikube service list)
 CURL='curl --connect-timeout 1 --max-time 1 --output /dev/null --silent'
-ES_CONSOLE_URL=`echo "$SERVICES" | grep expose-es-console | awk  '{print $6}'`
+ES_CONSOLE_URL=`echo "$SERVICES" | grep expose-es-console | awk '{print $8}'`
 
 
 # polling es-prom-server
@@ -26,8 +26,6 @@ function polling_service {
   echo "$1 is up"
 }
 
-set -x
 echo "SERVICES: $SERVICES"
 polling_service $ES_CONSOLE_URL
-set +x
 echo "poll services takes $seconds sec"
--- a/scripts/setup-minikube-for-linux.sh
+++ b/scripts/setup-minikube-for-linux.sh
@@ -3,7 +3,7 @@
 set -exu
 
 KUBERNETES_VERSION="v1.15.2"
-MINIKUBE_VERSION="v1.2.0"
+MINIKUBE_VERSION="v1.12.1"
 
 # Preflight checks
 if ! command -v helm > /dev/null; then


jedahu (approved):

katsutoxin:
I just checked circleCI log https://app.circleci.com/pipelines/github/lightbend/console-charts/626/workflows/92d4e692-54ea-42dd-96b3-43671b7c3a29

the script failed in `./scripts/wait_es_services.sh`, which is to wait for console-ui is up.

The potential problem is in this line https://github.com/lightbend/console-charts/blob/master/enterprise-suite/tests/e2e/scripts/wait_es_services.sh#L8
```
ES_CONSOLE_URL=`echo "$SERVICES" | grep expose-es-console | awk  '{print $6}'`
```

in case the minikube upgrade, and the format change, it could have problem.

To debug, 
we can 
- add `set -x` in the script to know what is going on
- dump the output of `minikube service list`

marcospereira:
> in case the minikube upgrade, and the format change, it could have problem.

I'm not sure this is a change in minikube output format, but instead, services don't have a url by default unless you run `minikube service service-name`. At least this is what happens on my machine:

```bash
$ minikube version
minikube version: v1.13.0
commit: 0c5e9de4ca6f9c55147ae7f90af97eff5befef5f

$ minikube status
minikube
type: Control Plane
host: Running
kubelet: Running
apiserver: Running
kubeconfig: Configured

$ minikube service list
|----------------------|-------------------------------|--------------|-----|
|      NAMESPACE       |             NAME              | TARGET PORT  | URL |
|----------------------|-------------------------------|--------------|-----|
| default              | kubernetes                    | No node port |
| kube-system          | kube-dns                      | No node port |
| kube-system          | metrics-server                | No node port |
| kubernetes-dashboard | dashboard-metrics-scraper     | No node port |
| kubernetes-dashboard | kubernetes-dashboard          | No node port |
| lightbend            | console-api                   | No node port |
| lightbend            | console-server                | No node port |
| lightbend            | expose-es-console             |           80 |     |
| lightbend            | grafana-server                | No node port |
| lightbend            | lagom-k8s-sample-svc          | No node port |
| lightbend            | play-k8s-sample-svc           | No node port |
| lightbend            | prometheus-kube-state-metrics | No node port |
| lightbend            | prometheus-server             | No node port |
| lightbend            | reactive-bbq                  | No node port |
|----------------------|-------------------------------|--------------|-----|
```

But, after running `minikube service expose-es-console --url -n lightbend` I get:

```bash
$ minikube service expose-es-console --url -n lightbend
🏃  Starting tunnel for service expose-es-console.
|-----------|-------------------|-------------|------------------------|
| NAMESPACE |       NAME        | TARGET PORT |          URL           |
|-----------|-------------------|-------------|------------------------|
| lightbend | expose-es-console |             | http://127.0.0.1:61694 |
|-----------|-------------------|-------------|------------------------|
http://127.0.0.1:61694
```

I also got a warning saying that "because you are using a Docker driver on darwin, the terminal needs to be open to run it." Maybe this behaviour only happens on macOS?

Anyway, I suspect this is what we need to change the script to be like:

```bash
# create a tunnel to `expose-es-console` service. --url avoids that browsers are open.
SERVICES=$(minikube service expose-es-console --url)
CURL='curl --connect-timeout 1 --max-time 1 --output /dev/null --silent'
ES_CONSOLE_URL=`echo "$SERVICES" | tail -1`
```

It will either create the tunnel, or at least it relys on a minikube command to get the service url.

> dump the output of `minikube service list`

Adding debug information to the script sounds good to me no matter the change we decide to do here.

marcoderama:
I'm not sure that's the right approach @marcospereira.  I think @katsutoxin had the right idea.  Take a look at my most recent change.  Now it's getting farther and failing in a different place.  (The tests must do the expose business.)

Looks like the name of the cluster is not what it's expecting at least.  @katsutoxin, any ideas on the latest errors?  Seems to me we'd have to run the tests locally and see what's going on.

marcoderama:
I'm not sure how much effort this change is worth.  It's kinda interesting to see what breaks though...

katsutoxin:
Hi @marcoderama and @marcospereira ,

Great job !  I am glad to see the e2e test starting running.   There are two fails in e2e test 
The log is https://app.circleci.com/pipelines/github/lightbend/console-charts/635/workflows/932998cf-8e97-49bd-8c39-9125f73c3a82/jobs/2051/parallel-runs/0/steps/0-106

##  The first error is 
```
Running: cluster-page.spec.ts...  
 <testcase name="Cluster Page Test cluster detail panel check"
   <failure><![CDATA[CypressError: Timed out retrying: Expected to find content: 'Name' within the element: <lb-panel.panel.no-margin> and with the selector: '.label-key' but never did.
```
The test code is in https://github.com/lightbend/console-charts/blob/master/enterprise-suite/tests/e2e/cypress/integration/cluster-page.spec.ts#L12

If we trace down, the problem is in this line https://github.com/lightbend/console-charts/blob/5e4a48f1cddac6f4e406f10b5af74e82bcfaafad/enterprise-suite/tests/e2e/cypress/support/cluster-details.ts#L4

It means that it does not show text "Name" in detail panel (the left sidebar).

We can checkout the running video here 
https://app.circleci.com/pipelines/github/lightbend/console-charts/635/workflows/932998cf-8e97-49bd-8c39-9125f73c3a82/jobs/2051/artifacts

Please click on the "cluster-page.spec.ts.mp4" one. We can see there is no "Name" field. As I remember, to show the cluster name was a hack and only works in minikube. Looks like new minikube does not support it.  We can even clean up the frontend and e2e_test code for showing it.



## The second error is 
```
Running: workload-page.spec.ts... 

 <testcase name="Workload Page Test validate detail panels" 
    <failure><![CDATA[CypressError: Timed out retrying: expected '<span.label-value.ellipsize>' to have text 'minikube', but the text was 'default-023eb11f-33fc-4af3-bfd2-4055e42dc801'

```
It means that it is expected to show "minikube" in node name field. However, it shows 'default-023eb11f-33fc-4af3-bfd2-4055e42dc801'  ,  Looks like it is introduced by minikube upgrade


## Other debug tips

We recorded the video, so we can check it.

The code is in https://github.com/lightbend/console-charts/blob/5e4a48f1cddac6f4e406f10b5af74e82bcfaafad/.circleci/config.yml#L67-L68

marcospereira:
@marcoderama we are now using the latest version of minikube. So I guess we can close this one. Re-open if there is something missing, though.

PR #481, opened 2020-09-22
helm3 compat: Add --force-update to helm repo add

pyoio:
`helm repo add ...` under v2 would be fine if you added the same repo multiple times (it would by default overwrite the repo name with the provided URL).

As of v3, if you attempt to add a repo whose name already exists, you get an error. Seen on [this support ticket](https://support.lightbend.com/support/lightbend/ShowHomePage.do#Cases/dv/427921000009056037)

This PR adds the `--force-update` so that the customer can rerun the lbc.py script without error.

PR #485, opened 2020-10-14
Support Helm 3.3.2+

marcospereira:
Helm 3.3.2+ fails if there is an attempt to add the same repo multiple times.

To prevent that, it is necessary to use `--force-update` flag.

This supersedes #481.

### DIFF
--- a/enterprise-suite/scripts/lbc.py
+++ b/enterprise-suite/scripts/lbc.py
@@ -549,11 +549,8 @@ def install(creds_file):
             chart_name = "{}/{}".format(args.repo_name, args.chart)
 
             # helm 3.3.2+ requires helm repo add --force-update to avoid errors re-running this script
-            supports_force_flag = version_is_at_least('helm version --client --short', '3.3.2')
-            if (supports_force_flag):
-                execute('helm repo add --force-update {} {}'.format(args.repo_name, args.repo))
-            else:
-                execute('helm repo add {} {}'.format(args.repo_name, args.repo))
+            force_flag = '--force-update ' if version_is_at_least('helm version --client --short', '3.3.2') else ''
+            execute('helm repo add {}{} {}'.format(force_flag, args.repo_name, args.repo))
 
             execute('helm repo update')
             tempdir = make_fetchdir()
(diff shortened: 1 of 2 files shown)


enterprise-suite/scripts/lbc.py
```diff
@@ -539,7 +547,14 @@ def install(creds_file):
...
         else:
             chart_name = "{}/{}".format(args.repo_name, args.chart)
-            execute('helm repo add {} {}'.format(args.repo_name, args.repo))
+
+            # helm 3.3.2+ requires helm repo add --force-update to avoid errors re-running this script
+            supports_force_flag = version_is_at_least('helm version --client --short', '3.3.2')
+            if (supports_force_flag):
+                execute('helm repo add --force-update {} {}'.format(args.repo_name, args.repo))
+            else:
+                execute('helm repo add {} {}'.format(args.repo_name, args.repo))
```
marcoderama, line 556: I might go with something like:
```
            force-flag = ""
            if (supports_force_flag):
                force-flag = "--force-update"
            execute('helm repo add {} {} {}'.format(force-flag, args.repo_name, args.repo))
```
but I'm not a python dude so maybe that's anathema.  I just like to see the command defined once in case it ever needs to get updated.

marcoderama (approved):
Had one ignorable comment.

enterprise-suite/scripts/lbc.py
marcospereira: Yeah, I've considered that too while writing the code, but decide that it would be clear to have it the way it is now. No strong opinion about it, though, and I would be happy with whichever option.

pyoio: No strong opinions here either tbh, I agree that one command would be preferable but I think 2 is OK, if we need more variants then I'd refactor it out into generating the args dynamically.

pyoio:
LGTM!

enterprise-suite/scripts/lbc.py
marcospereira: See 748fb2d. It looks clearer, indeed. :-)

marcoderama: And I like your clever handling of the added space character!

PR #488, opened 2020-10-19
Remove openshift demo deployment

marcospereira:
We don't even have an openshift cluster anymore.


marcoderama (approved):

marcospereira:
I'm merging this myself since the build is broken when releasing, and the Central Park env does not even exist anymore.

PR #489, opened 2021-01-06
set-tools-for-debian.sh to use new stable helm repo

marcoderama:
The `release_nightly` CI build step is failing.  It fails when running `helm init` with
```
Error: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Failed to fetch https://kubernetes-charts.storage.googleapis.com/index.yaml : 403 Forbidden
```

This change should fix that, as suggested here:  https://stackoverflow.com/a/65404574/12957030

### DIFF
--- a/scripts/setup-tools-for-debian.sh
+++ b/scripts/setup-tools-for-debian.sh
@@ -30,7 +30,7 @@ curl -LO https://storage.googleapis.com/kubernetes-helm/${helm_file}
 echo "0fa2ed4983b1e4a3f90f776d08b88b0c73fd83f305b5b634175cb15e61342ffe ${helm_file}" | sha256sum --check
 tar xzvf ${helm_file}
 sudo cp linux-amd64/helm /usr/local/bin/
-helm init -c
+helm init -c --stable-repo-url https://charts.helm.sh/stable
 
 # socat is needed for helm init --wait to work
 sudo apt-get install -y socat


marcospereira:
It looks like Helm 2.17.0 changed the default already. See https://github.com/helm/helm/releases/tag/v2.17.0 and https://github.com/helm/helm/pull/8901. I wonder if we should instead test against that version? Is helm 2.10.x supported?

marcospereira:
Kind of answering my own question:

https://helm.sh/docs/topics/version_skew/

> As of Helm 3, Helm is assumed to be compatible with n-3 versions of Kubernetes it was compiled against. Due to Kubernetes' changes between minor versions, Helm 2's support policy is slightly stricter, assuming to be compatible with n-1 versions of Kubernetes.
> 
> ...
> 
> It is not recommended to use Helm with a version of Kubernetes that is newer than the version it was compiled against, as Helm does not make any forward compatiblility guarantees.

And Helm 2.10.x supports the Kubernetes version from 1.9.x to 1.10.x (which are not supported anymore).

---

Anyway, Helm 2.17.0 is probably too edgy for us right now, so the solution you proposed makes sense to me.

marcospereira (approved):

marcoderama:
The nightly build ran fine after this change.

PR #490, opened 2021-01-08
Update multiple images to versions using alpine 3.12

marcospereira:
- console-gui to version 1.4.10-RC1
- console-grafana to version 0.5.0-RC1

### DIFF
--- a/enterprise-suite/values.yaml
+++ b/enterprise-suite/values.yaml
@@ -9,11 +9,17 @@ consoleUIConfig:
 # Update these versions for a release.
 #
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-console
-esConsoleVersion: v1.4.9
+# Repo: console-gui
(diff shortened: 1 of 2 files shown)


enterprise-suite/values.yaml
```diff
@@ -9,11 +9,17 @@ consoleUIConfig:
 # Update these versions for a release.
 #
 # lightbend-docker-commercial-registry.bintray.io/enterprise-suite/es-console
-esConsoleVersion: v1.4.9
+# Github repo: https://github.com/lightbend/console-gui/
```
marcoderama, line 12: This file is open to everyone so I'm not sure we want to add references to private repos here.

Ditto for other comments here.

marcospereira: Remove the URLs to the repositories. I've kept the names, though.

marcospereira:
Merging this myself too so that we can have a RC here.

PR #492, opened 2021-04-20
Use compatible Kubernetes versions for client and server

marcospereira:
- Use latest version of minikube
- Use a consistent version of k8s between server and client

### DIFF
--- a/scripts/setup-minikube-for-linux.sh
+++ b/scripts/setup-minikube-for-linux.sh
@@ -3,7 +3,7 @@
 set -exu
 
 KUBERNETES_VERSION="v1.15.2"
-MINIKUBE_VERSION="v1.2.0"
+MINIKUBE_VERSION="latest"
 
 # Preflight checks
 if ! command -v helm > /dev/null; then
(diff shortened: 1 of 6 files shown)


scripts/setup-minikube-for-linux.sh
```diff
@@ -2,8 +2,9 @@
 
 set -exu
 
-KUBERNETES_VERSION="v1.15.2"
-MINIKUBE_VERSION="v1.2.0"
+# Keep it in sync with the version declared in scripts/setup-tools-for-debian.sh
+KUBERNETES_VERSION="v1.18.18"
+MINIKUBE_VERSION="latest"
```
marcoderama, line 7: You sure you want this to be a moving target?

scripts/setup-tools-for-debian.sh
```diff
@@ -36,7 +39,9 @@ helm init -c --stable-repo-url https://charts.helm.sh/stable
 sudo apt-get install -y socat
 
 # kubectl
-curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl \
+# Keep it in sync with the version declared in scripts/setup-minikube-for-linux.sh
+KUBERNETES_VERSION="v1.18.18"
```
marcoderama, line 43: Could you pull this directly from `scripts/setup-minikube-for-linux.sh` itself, just to avoid there being a mismatch?

scripts/setup-minikube-for-linux.sh
marcospereira: Yes, that is that we are doing for lightbend/console-gui already, which uses this reposity in its build. So at least it is consistent.

scripts/setup-tools-for-debian.sh
marcospereira: For other reasons (mainly the helm version used), it is better not to use the same version. I will update helm in another PR and try to keep the versions in a single place as you suggested. :-)

marcoderama (approved):
I don't understand the minikube/debian kubernetes difference, but it was different before, and I figure you know what you're doing.

marcospereira:
> I figure you know what you're doing.

You are putting too much faith in me. 😅

marcospereira:
Green build, so I'm merging this myself. :-)

PR #497, opened 2021-04-22
Add docker-commercial-credentials to whitesource workflow

marcospereira:


enterprise-suite/tests/e2e/scripts/wait_es_services.sh
```diff
@@ -11,6 +11,9 @@ CURL='curl --connect-timeout 1 --max-time 1 --output /dev/null --silent'
 ES_CONSOLE_URL=$(echo "$SERVICES" | grep expose-es-console | awk -F "|" '{ print $(NF-1)}')
 echo "ES-Console URL: ${ES_CONSOLE_URL}"
 
+# TODO remove before merging the pr
+minikube service --url expose-es-console
```
marcoderama: Don't forget! :-)

scripts/trigger-build-helm-charts.sh
```diff
@@ -4,7 +4,7 @@ set -eu
 
 vcs=github
 org=lightbend
-project=console-gui
+project=build-helm-charts
```
marcoderama, line 7: I wouldn't mind if you hard-wired that in the curl url again.  Or called the variable something else.  But leaving this as is works too.

marcoderama (approved):
Approved as long as you don't forget to remove that expose business.

scripts/trigger-build-helm-charts.sh
marcospereira: Yeah, both work for me too. I left it like this because it looks easier to me than mixes with other CircleCI+curl parameters.

enterprise-suite/tests/e2e/scripts/wait_es_services.sh
marcospereira: See it again. :-)

There is a way to get the URL without parsing the `service list` output. :-)

marcospereira:
@marcoderama please, if you have the time, do a quick review again. If you are happy, feel free to merge.

marcoderama (approved):

PR #498, opened 2021-04-26
Add commercial-registry.yaml to make it easier to install older versions of Console

marcospereira:


marcoderama (approved):
Looks reasonable.  How does one use it?

marcospereira:
> How does one use it?

It would be something like:

```bash
./lbc.py install \
    --namespace="lightbend" \
    --version=1.2.14 \
    --set exposeServices=NodePort -- \
    --values=values.yaml \
    -f https://raw.githubusercontent.com/lightbend/console-charts/master/enterprise-suite/commercial-registry.yaml
```

I just validated it locally using Minikube. Also, we did the same for Cloudflow: lightbend/cloudflow-helm-charts#18

